PR #28, opened 2020-10-28
Simplify comparison star notebook and functions

mwcraig:


mwcraig:
@madelyn914 -- you will need to pull in the latest version from master. Grab the notebook `comp-star-template.ipynb` and it should 🤞work.

PR #42, opened 2021-06-09
Add optional column name

madelyn914:


mwcraig:
Can you please add a couple of lines of documentation for the new argument? It should go at roughly line 35.

Looks like something changed again in photutils, which is what is calling all of these failures.

mwcraig (approved):

PR #90, opened 2023-05-26
Typos and other cleanup

JuanCab:
So I have worked to do the following:
- Fixed up source code in various places to export (via `__all__` list) most of their functions and classes.  This also required editing some of the `__init__.py` source files to import the source files.  Most of the changes were in `visualization` directory of `stellarphot`.
- Fixed the documentation configuration to know it is making `stellarphot` and not `glowing-waffles` documentation and also to search in additional source code locations (notably `analysis.py` and `source_detection.py`.
- Removed unnecessary `coordinates.py` file which defined one function used in one spot in `photometry.py` once.  Added the call to retrieve WCS data directly to `photometry.py`.
- All references to `np.float`, `np.int`, and `np.bool` replaced with `float`, `int`, and `bool` respectively since those aliases were deprecated in NumPy 1.20 and removed in NumPy 1.24.


mwcraig (changes requested):
Looking good, just some minor changes. We should maybe change to automating code style things/use [`black`](https://black.readthedocs.io/)

mwcraig:
Thanks, @JuanCab

PR #94, opened 2023-06-06
Purging and tweaking notebooks

JuanCab:
I went through the notebooks folder and purged redundant or outdated jupyter notebooks and updated one notebook which still tried to load `glowing-waffles` instead of `stellarphot` libraries.


CHANGES.rst
```diff
@@ -1,18 +1,48 @@
-1.1.2 (2022-10-18)
+1.4.0 (2023-06-06)
 ------------------
 
 New Features
 ^^^^^^^^^^^^
 
+Other Changes and Additions
```
mwcraig, line 7: All of the changes can go under `1.3.9`, I think, and the date can eclipsing binary set to `unreleased` for now. The idea is to build a list of changes that will be included in the next release.

JuanCab: I'll apply a fix.  

"eclipsing binary" ... do you have some text expander macro misfiring there?

JuanCab:
I am closing this pull request because the `CHANGES.rst` changes can be delayed until my next PR.

CHANGES.rst
mwcraig: yeah, TextExpander has "eb" = "eclipsing binary".

Might take that one out since I don't use it much and I transpose letters much more often than I used to...

PR #108, opened 2023-06-27
Fixes to ComparisonViewer for release version of ImageWidget

JuanCab:
Apparently the release version of astrowidgets does NOT support overwrite as an option to saving an ImageWidget, I implemented the same functionality via the os package.  I also implemented an implementation-wide parameter to allow overwriting on all output files.

Also fixed an issue where `if v_angle:` was triggering an exception as `ambiguous` (I suspect this was a new behavior added in a recent version of astropy), instead of having v_angle evaluate as true is not empty.  I replaced it with a check of `if len(v_angle):` which evaluates the desired way.

I also went into the code for retrieving GAIA stars and made the server a constant instead of a parameter to be passed in (it is still a public attribute and can be seen).

### DIFF
--- a/stellarphot/visualization/comparison_functions.py
+++ b/stellarphot/visualization/comparison_functions.py
@@ -575,8 +575,7 @@ class ComparisonViewer:
         try:
             self.variables.write(filename, overwrite=self.overwrite_outputs)
         except OSError:
-            print(f"ERROR: Can't save variables to file, likely because existing file ({filename}) can not be overwritten. Set overwrite_outputs=True to address this.")
-
+            raise OSError(f"Existing file ({filename}) can not be overwritten. Set overwrite_outputs=True to address this.")
 
     def _show_label_button_handler(self, change):
         value = change['new']
@@ -596,7 +595,7 @@ class ComparisonViewer:
         try:
             self.generate_table().write(filename, overwrite=self.overwrite_outputs)
         except OSError:
-            print(f"ERROR: Can't save apertures to file, likely because existing file ({filename}) can not be overwritten. Set overwrite_outputs=True to address this.")
+            raise OSError(f"Existing file ({filename}) can not be overwritten. Set overwrite_outputs=True to address this.")
 
     def _make_control_bar(self):
         self._show_labels_button = ipw.ToggleButton(description='Click to show labels')
@@ -728,7 +727,7 @@ class ComparisonViewer:
             try:
                 self.iw.save(self._field_name.value)
             except OSError:
-                print(f"ERROR: Can't save full field image, likely because existing file ({self._field_name.value}) can not be overwritten. Set overwrite_outputs=True to address this.")
+                raise OSError(f"Existing file ({self._field_name.value}) can not be overwritten. Set overwrite_outputs=True to address this.")
 
         if self._zoom_name.value:
             self.tess_field_zoom_view()
@@ -738,7 +737,7 @@ class ComparisonViewer:
             try:
                 self.iw.save(self._zoom_name.value)
             except OSError:
-                print(f"ERROR: Can't save zoomed image, likely because existing file ({self._zoom_name.value}) can not be overwritten. Set overwrite_outputs=True to address this.")
+                raise OSError(f"Existing file ({self._zoom_name.value}) can not be overwritten. Set overwrite_outputs=True to address this.")
 
     def generate_table(self):
         """
--- a/stellarphot/visualization/comparison_functions.py
+++ b/stellarphot/visualization/comparison_functions.py
@@ -596,7 +596,7 @@ class ComparisonViewer:
         try:
             self.generate_table().write(filename, overwrite=self.overwrite_outputs)
         except OSError:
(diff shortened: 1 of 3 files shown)


stellarphot/visualization/comparison_functions.py
```diff
@@ -188,12 +189,12 @@ def mag_scale(cmag, apass, v_angle, RD_angle,
     """
     high_mag = apass['r_mag'] < cmag + dimmer_dmag
     low_mag = apass['r_mag'] > cmag - brighter_dmag
-    if v_angle:
+    if len(v_angle)>0:
```
mwcraig, line 192:
```suggestion
    if len(v_angle) > 0:
```

stellarphot/visualization/comparison_functions.py
```diff
@@ -188,12 +189,12 @@ def mag_scale(cmag, apass, v_angle, RD_angle,
...
         good_v_angle = v_angle > 1.0 * u.arcsec
     else:
         good_v_angle = True
 
-    if RD_angle:
+    if len(RD_angle)>0:
```
mwcraig, line 197:
```suggestion
    if len(RD_angle) > 0:
```

stellarphot/visualization/comparison_functions.py
```diff
@@ -707,11 +717,23 @@ def save_tess_files(self, button=None):
         """
         if self._field_name.value:
             self.tess_field_view()
-            self.iw.save(self._field_name.value, overwrite=True)
+            # Remove output file if it exists
+            if os.path.exists(self._field_name.value) and self.overwrite_outputs:
```
mwcraig, line 721: It would be nice to switch this to `pathlib` instead. I can do that later in a separate PR if you prefer,

stellarphot/visualization/comparison_functions.py
```diff
@@ -707,11 +717,23 @@ def save_tess_files(self, button=None):
...
+                os.remove(self._field_name.value)
```
mwcraig, line 722: `pathlib` can also do the remove

stellarphot/visualization/comparison_functions.py
```diff
@@ -707,11 +717,23 @@ def save_tess_files(self, button=None):
...
+            try:
+                self.iw.save(self._field_name.value)
+            except:
+                print("Error saving full field image.")
 
         if self._zoom_name.value:
             self.tess_field_zoom_view()
-            self.iw.save(self._zoom_name.value, overwrite=True)
+            # Remove output file if it exists
+            if os.path.exists(self._zoom_name.value) and self.overwrite_outputs:
```
mwcraig, line 731: Since this came up twice should it be moved into a function? Hopefully the need for it will go away soon...

mwcraig (changes requested):
A couple really minor PEP8 changes and one other change I'm happy to implement if you want.

stellarphot/visualization/comparison_functions.py
JuanCab: I'll get this addressed shortly.

stellarphot/visualization/comparison_functions.py
JuanCab: I'll get this addressed shortly.

stellarphot/visualization/comparison_functions.py
JuanCab: Lets see if those astrowidgets guys fix this.

JuanCab (commented):
Still need to pathlib things, but close.

JuanCab (commented):
I have made the switch from `os` to `pathlib` and checked the new code by running comparison viewer and triggering the overwrite situation.  It is working. pytest cleared it as well.

stellarphot/visualization/comparison_functions.py
```diff
@@ -707,11 +715,23 @@ def save_tess_files(self, button=None):
         """
         if self._field_name.value:
             self.tess_field_view()
-            self.iw.save(self._field_name.value, overwrite=True)
+            # Remove output file if it exists
+            if Path(self._field_name.value).exists() and self.overwrite_outputs:
+                Path(self._field_name.value).unlink()
+            try:
+                self.iw.save(self._field_name.value)
+            except:
```
mwcraig, line 723: Usually it is better to not use a bare `except`, I think, in case you accidentally capture an error message that is really a bug. I think we could get away with no try/except here, anticipating that if an error is raised it will probably be for something real, and should be reported to the user with the exception details/traceback.

mwcraig (changes requested):
Apparently I'm a little fussy today -- what do you think about removing the try/except?

stellarphot/visualization/comparison_functions.py
JuanCab: Oh crap, I forgot to capture the specific error.  I ended up wrapping all the file creation calls that could fail due to attempting to overwrite a file without permission with a try..except OSError.  That should be specific enough.  I notice the errors still show up on the Jupyterlab log page, so may not really be much easier for users.

JuanCab (commented):
Now raises an OSError returning a custom message if it attempts to overwrite a file without permission.

mwcraig (approved):

PR #125, opened 2023-06-30
Initial version of BaseEnhancedTable and PhotometryData added

### LINKED ISSUE #114: New base class that is Table plus header/attributes
The base class should:

+ Enforce names that are common to derived from classes
    - Unique object ID
    - ra
    - dec
+ Do not add ability to add columns as attributes for now

### LINKED ISSUE #115: Write Photometry class that uses new base clase

JuanCab:
Fixes #114 and fixes #115.

I added an initial version of the `BaseEnhancedTable` and `PhotometryData` classes to the `core.py` file and added extensive tests to the `tests/test_core.py`.

Please note while `BaseEnhancedTable` does force certain attributes to exist, the `PhotometryData` did NOT add any required attributes.  Its easy enough to add, I just wasn't sure what you wanted to require as being there.


CHANGES.rst
```diff
@@ -3,12 +3,15 @@
 
 New Features
 ^^^^^^^^^^^^
++ Continued fleshing out of the documentation.
```
mwcraig, line 6: Don't think we need this. The new stuff has documentation but that should be the norm going forward.

CHANGES.rst
```diff
@@ -3,12 +3,15 @@
...
++ Creation of new data classes for handling aperture, photometry, and catalog data in a more consistent way.
```
mwcraig, line 7:
```suggestion
+ Creation of new data classes for handling aperture, photometry, and catalog data in a more consistent way. [#125]
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
         self.read_noise = read_noise
         self.dark_current = dark_current
+
+
+class BaseEnhancedTable:
+    """
+    A base class to hold an `astropy.table.Table` data table and extend it
+    so that chosen columns can be accessed as attributes.  This base class
+    imposes the requirement that the table have columns corresponding to
+    attributes 'id', 'ra', and 'dec' attributes, as well as confirming
```
mwcraig, line 74:
```suggestion
    attributes 'id', 'ra', and 'dec', as well as confirming
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    the 'ra' and 'dec' are in degrees.
+
+    Parameters
+    ----------
+
+    table_descript: `numpy.ndarray`
```
mwcraig, line 80:
```suggestion
    table_description: `numpy.ndarray`
```

May as well complete the word...

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        This is a 2-D numpy array where each row of the array is the
```
mwcraig, line 81: Splitting hairs here, but does it really need to be a numpy arraY? Need to think about this a little more

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        table column name, dtype, astropy unit (can be None), and
+        the associated attribute name (can also be None, in which case
+        tha column of the table has no associated instance attribute).
+        The description must include 'ra', 'dec', and 'id' attributes
+        among the listed attributes.
+
+    data: `astropy.table.Table`, optional
+        A table containing all the instrumental aperture photometry results
```
mwcraig, line 89: This doesn't have to be a photometry at this stage, right?

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        for a given field on a given night.  If no data is passed,
+        an empty data table with the proper columns but no data is created.
+        If a table is provided, its format is checked against the
+        table_descript.
```
mwcraig, line 93:
```suggestion
        table_description.
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+
+        If data is passed in, it will be checked to make sure all columns
+        listed in table_descript exist, HOWEVER, additional columns in data
```
mwcraig, line 96:
```suggestion
        listed in table_description exist, HOWEVER, additional columns in data
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        not listed in table_descript will NOT be removed.
```
mwcraig, line 97:
```suggestion
        not listed in table_description will NOT be removed.
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    id: `astropy.table.Column`
+        A column of object id values (no specific format assumed)
+
+    ra: `astropy.table.Column`
+        A column of right ascension values with units of degrees
+
+    Other attributes may be created as defined by table_descript.
+    """
+
+    def __init__(self, table_descript=None, data=None):
```
mwcraig, line 117: Won't keep going with changing `table_descript` → `table_description` since the code chages are easier to do in code.

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        # Handle parameters
+        self._table_descript = table_descript
+        self.data = data
+
+        # Confirm a proper table description is passed
+        if (type(self._table_descript) is not np.ndarray):
```
mwcraig, line 123: I think the preferred idiom is below, but do we need to explicitly check?

```suggestion
        if not isinstance(self._table_descript, np.ndarray)):
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+                    raise ValueError(f"column {this_col} is not present in input data.")
+                # Check type
+                if data[this_col].dtype != coltypes[i]:
+                    raise ValueError(f"data[{this_col}] is of wrong type (declared {coltypes[i]} but reported as {data[this_col].dtype}).")
+                if data[this_col].unit != colunits[i]:
+                    raise ValueError(f"data[{this_col}] is of wrong unit (declared {colunits[i]} but reported as {data[this_col].unit}).")
+
+        # Create attibutes corresponding to table column
+        self.cols_to_attr(colnames, attrnames)
+
```
mwcraig, line 166:
```suggestion
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+            # each column listed in table_descript exists and is he write type and unit.
+            # NOTE: This ignores any columns not in the table_descript, it does not remove them.
+            for i,this_col in enumerate(colnames):
+                # Failure is assumed to be due to bad column value
+                try:
+                    data[this_col]
+                except KeyError:
+                    raise ValueError(f"column {this_col} is not present in input data.")
+                # Check type
+                if data[this_col].dtype != coltypes[i]:
```
mwcraig, line 159: I'm going to read up on dtypes; I think this might be too restrictive: for example, what if the column is `float32` instead of `float64`?

Maybe the better approach is to always create the table, like on line 147, but then set the column?

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+                    raise ValueError(f"data[{this_col}] is of wrong type (declared {coltypes[i]} but reported as {data[this_col].dtype}).")
+                if data[this_col].unit != colunits[i]:
```
mwcraig, line 161: I think the preferred comparison is 

```suggestion
                if data[this_col].unit is not colunits[i]:
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+                    raise ValueError(f"data[{this_col}] is of wrong unit (declared {colunits[i]} but reported as {data[this_col].unit}).")
+
+        # Create attibutes corresponding to table column
+        self.cols_to_attr(colnames, attrnames)
+
+
+    def cols_to_attr(self, colnames, attrnames):
+        # Connects table columns to attributes
+        for i,attr in enumerate(attrnames):
```
mwcraig, line 170:
```suggestion
        for i, attr in enumerate(attrnames):
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+            if attr is not None:
+                self.__setattr__(attr, self.data[colnames[i]])
+
+
+class PhotometryData(BaseEnhancedTable):
+    """
+    A base class to hold an `astropy.table.Table` table of reduce photometry
+    data, extending it with a mapping of various table columns to attributes.
+
+    USAGE NOTES: If you input a data file, it MUST contain the following columns
```
mwcraig, line 180: I think this should be lower in the docstring, in the "Notes" section. I also think we need to auto-generate this.

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    annulus_outer       float64     pix
+    annulus_area        float64
+    exposure            float64       s
+    date-obs            str23
+    night               int64
+    aperture_net_flux   float64     adu
+    BJD                 float64
+    mag_inst            float64
+    airmass             float64
+    filter              str2
```
mwcraig, line 210: The string types here are problematic because valid values may be longer than the length limit being imposed here. Some of the AAVSO filter names may be more than two characters for example.

I'll look into options here, but my recollection is that there is not a great workaround except making the length large enough to holad any reasonable value.

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    file                 str38
+    star_id             int64
+    mag_error           float64     1 / adu
```
mwcraig, line 213: I don't think this is the right unit; also, it isn't consistent with the magnitude unit.

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    noise               float64
+    noise-aij           float64
+    snr                 float64     adu
+
+    Parameters
+    ----------
+
+    observatory: `astropy.coordinates.EarthLocation`
```
mwcraig, line 221: Am wondering if this should be its own class? Is there anything that needs to be stored beyond the location? Actually, `astroplan` might have an `Observatory` class already.

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        # Set metavariables
+        self.observatory = observatory
+        self.camera = camera
+        self.filter_map = filter_map
+
+        # Handle data table
+        super().__init__(self.phot_descript, data)
+
+
+    def update_filters(self):
```
mwcraig, line 366: It seems like this should be called in the `__init__`

mwcraig: Also, I think we'll need to allow for the possibility that there is not a corresponding filter, eg `540_bp40`.

stellarphot/tests/test_core.py
```diff
@@ -9,3 +15,105 @@ def test_camera_attributes():
...
+                                ['sky_per_pix_med', '<f8', u.adu, 'spp_med'],
+                                ['sky_per_pix_std', '<f8', u.adu, 'spp_std'],
+                                ['fwhm_x', '<f8', None, None],
+                                ['fwhm_y', '<f8', None, None],
+                                ['width', '<f8', None, 'fwhm']])
+
+    # This should create a blank data set
+    test_base = BaseEnhancedTable(test_descript)
+
+    assert type(test_base.data) == Table
```
mwcraig, line 45: I think this should be 
```suggestion
    assert isinstance(test_base.data, Table)
```

stellarphot/core.py
mwcraig, line 117: I don't think the `table_descript` is really optional.

```suggestion
    def __init__(self, table_descript, data=None):
```

stellarphot/core.py
```diff
@@ -62,3 +64,307 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+
+        # Perform processing once initialized
+        if self.data is None:
+            # Create empty astropy table with right format and data
+            self.data = Table(names=colnames, masked=False, dtype=coltypes, units=colunits)
+        else:
+            # Check the format of the data table matches the table_descript by checking
+            # each column listed in table_descript exists and is he write type and unit.
+            # NOTE: This ignores any columns not in the table_descript, it does not remove them.
+            for i,this_col in enumerate(colnames):
```
mwcraig, line 152:
```suggestion
            for i, this_col in enumerate(colnames):
```

Also, why not do

```python
for coltype, this_col in zip(coltypes, colnames):
    ...
```

stellarphot/tests/test_core.py
```diff
@@ -9,3 +15,105 @@ def test_camera_attributes():
...
+    assert len(test_base.ra) == 0
+    assert len(test_base.dec) == 0
+
+    # Should create a populated dataset properly and display the astropy data
+    test_base2 = BaseEnhancedTable(test_descript, testdata)
+    assert len(test_base2.ra) == 1
+    assert len(test_base2.dec) == 1
+
+    # Should raise exception because no inputs are passed
+    with pytest.raises(Exception):
```
mwcraig, line 55: We should check the exception type too, though if `table_description` is a required argument we probably don't need to test  whether an exception is raised when the argument isn't given.

I do think it makes sense to check each of the cases where the initialization can fail (with a `TypeError` or `ValueError`)

stellarphot/tests/test_core.py
```diff
@@ -9,3 +15,105 @@ def test_camera_attributes():
...
+                                ['width', '<f8', None, 'fwhm']])
+
+    # This should create a blank data set
+    test_base = BaseEnhancedTable(test_descript)
+
+    assert type(test_base.data) == Table
+    assert len(test_base.ra) == 0
+    assert len(test_base.dec) == 0
+
+    # Should create a populated dataset properly and display the astropy data
```
mwcraig, line 49: It would be nice if each of these was a separate test. The initialization can be handled either by making a test class that contains it or by making a pytest fixture. 

The reason for separating the tests is that when a test fails you know exactly what the failure is. As written, if `test_base_enhanced_table` fails you will have to dig a bit to find which line the failure is on.

stellarphot/tests/test_core.py
```diff
@@ -9,3 +15,105 @@ def test_camera_attributes():
...
+    assert phot_data.observatory.lat.value == 46.86678
+    assert phot_data.observatory.lat.unit == u.deg
+    assert phot_data.observatory.lon.value == -96.45328
+    assert phot_data.observatory.lon.unit == u.deg
+    assert round(phot_data.observatory.height.value) == 311
+    assert phot_data.observatory.height.unit == u.m
+    assert phot_data.filter_map['up'] == 'SU'
+    assert type(phot_data.data) == Table
+    assert len(phot_data.ra) == 1
+    assert len(phot_data.dec) == 1
```
mwcraig, line 119:
```suggestion
    assert len(phot_data.dec) == 1
    
```

mwcraig (commented):
I'm sure I'll have more to say -- I'm going to look over the metaprogramming section of fluent python to see if there is a different approach to take instead of subclassing.

stellarphot/core.py
JuanCab: I just needed something that was visually easy to construct and whose columns I could slice out.  I suppose one could use an astropy table, but it seemed overkill.

stellarphot/core.py
JuanCab: That's literally the units in the tables your currently generate.  Like the areas which should be pix*pix, I think some cleanup of units may need to happen.

stellarphot/core.py
JuanCab: All I could think of was what was the location unless you needed telescope properties for some reason.

stellarphot/core.py
JuanCab: The None allowed easier catching of someone passing no inputs in for some reason.

JuanCab:
Well let me know.  I looked at @dataclasses but subclassing data classes looked... painful and it forced an order to parameters I was unhappy with.

stellarphot/tests/test_core.py
JuanCab: When pytest fails, it gives the line number of the failing test, but I will see what I can do.

CHANGES.rst
JuanCab: Noted.

stellarphot/core.py
JuanCab: Well, the way it currently works, those filters will simply not be changed.

JuanCab: > It seems like this should be called in the `__init__`

Wan't sure if you wanted to force the issue, also, it should be possible to create a BLANK table using these classes, and then fill the table, in which case, you would want to call update_filters() later.

stellarphot/core.py
JuanCab: OK, done throughout code.

stellarphot/core.py
JuanCab: Whoops, cut/paste error.  Fixed.

stellarphot/core.py
JuanCab: Fixed throughout.

stellarphot/core.py
JuanCab: Fixed throughout.

stellarphot/core.py
JuanCab: Interestingly, this change causes a failure of "ValueError: data[mag_error] is of wrong unit (declared 1 / adu but reported as 1 / adu)." when running the tests.  I switched it back, but not sure why the error occurs.  I depends on if "is not" means literally not equal including memory location or if that is !=

stellarphot/core.py
JuanCab: I moved it, but still not my favorite approach.  I do think if you can think of attributes we don't need, go ahead and remove them.

stellarphot/tests/test_core.py
JuanCab: Tests have been separated.

stellarphot/core.py
JuanCab: 'is not' is checking for identity being the same, e.g. they are the SAME OBJECT, != checks for equality.  I suspect it fails with 1/adu because that is a constructed unit and so two different instances of it are considered two different objects.  An equality check seems to be the way to go.

stellarphot/core.py
JuanCab: Huh?

stellarphot/core.py
JuanCab: Whatever approach you think is acceptable is fine.

stellarphot/core.py
JuanCab: > Also, I think we'll need to allow for the possibility that there is not a corresponding filter, eg `540_bp40`.

Well, the way the code currently currently works, if the filter is not mentioned in the filter_map, the filter will simply not be changed.  So the filter_map should only be for those filters that *do* map to standard AAVSO names.

stellarphot/tests/test_core.py
JuanCab: Done.

JuanCab (commented):
I went through your comments and tried to address most of them.

JuanCab (commented):
I went through your comments and tried to address most of them.

stellarphot/core.py
mwcraig: Agreed about why `is not` doesn't work...

stellarphot/core.py
mwcraig: One too many blank lines. I'll get linting set up in github tomorrow...

stellarphot/core.py
mwcraig: 😬

stellarphot/core.py
mwcraig: > Well, the way the code currently currently works, if the filter is not mentioned in the filter_map, the filter will simply not be changed. So the filter_map should only be for those filters that do map to standard AAVSO names.

Good point...

mwcraig: > Wan't sure if you wanted to force the issue, also, it should be possible to create a BLANK table using these classes, and then fill the table, in which case, you would want to call update_filters() later.

I think we do want to force the issue but you are right we also need to update if data is added later. Not 100% sure we can enforce that, though

stellarphot/core.py
mwcraig: oops, I meant move this usage note, not the attributes.  The motivation for having the usage note lower is primarily that it is so long.

stellarphot/core.py
mwcraig: Some good news about types: I think we can go with generic python types `float`, `int` and `str`. The `str` column that is created automatically updates the length of the string used to hold the data. The `float` and `int` automatically get converted to numpy dtypes.

This:
```python
t = Table(names=['decimal', 'text', 'integer'], dtype=('float', 'str', 'int'))
```

leads to this:
```
<Table length=0>
decimal text integer
float64 str1  int64
------- ---- -------
```

stellarphot/core.py
JuanCab: So should I keep it as is?

stellarphot/core.py
JuanCab: I did move the usage note  to the end of the parameters.  I wanted to keep it next to the discussion of the `data` parameter since it really is a discussion of the restrictions on that `data` table.

stellarphot/core.py
JuanCab: Can you confirm if you subsequently add new data to the `text` column that is longer than the initial allocation, the string length gets updated?

mwcraig:
I'm not 100% sure the discussion about types went off to, but

> Can you confirm if you subsequently add new data to the text column that is longer than the initial allocation, the string length gets updated?

The answer seems to be complicated. If you start with an empty table the only way to populate it seems to be `add_row` or `add_rows`. If you add data that way, then the string column length gets changed automatically.

If the table already has data then what happens if you try to update a string column depend ons on how you do it.

If you create the table and add a row like this

```python
from astropy.table import Table
t = Table(names=['decimal', 'text'], dtype=('float', 'str'))
t.add_row((10.0, "hello"))
```

then `t["text"]` is string of length 5.

Updating like this changes to length 7:

```python
t["text"] = ["longest'}
```

but if you instead try this it truncates:

```python
t["text"][:] = ["longest"]
```

I think the reason is that `t["text"] = ...` is really replacing the original column, not simply updating the values  The second case updates the vlaues but also warns that they are being truncated.

mwcraig:
Something else just occurred to me -- changing an attribute doesn't change the underlying table, so an attribute and its corresponding table column can get out of sync.

Not sure how to deal with this; one option might be to make properties instead of settable attributes.

To reproduce, run this, and note that at the end `bet.number` and `bet.data["number"]` are different:

```python
import numpy as np
from astropy.table import Table

table_description = np.array([["number", float, None, "number"],
    ["ra", float, "degree", "ra"],
    ["dec", float, "degree", "dec"],
    ["id", int, None, "id"]
])

tab = Table(data=[[1], ["abc"], [1.0], [1.0], [5.1]], names=("id", "tex
    ...: t", "ra", "dec", "number"), units=(None, None, "degree", "degree", None
    ...: ))

bet = BaseEnhancedTable(table_description, data=tab)

bet.number

# Returns
# <Column name='number' dtype='float64' length=1>
# 5.1

bet.number = [6.1]
bet.number

# Returns
# [6.1]

bet.data["number"]

# Returns
# <Column name='number' dtype='float64' length=1>
# 5.1

```

JuanCab:
> Something else just occurred to me -- changing an attribute doesn't change the underlying table, so an attribute and its corresponding table column can get out of sync.
> 
> Not sure how to deal with this; one option might be to make properties instead of settable attributes.
> 
> To reproduce, run this, and note that at the end `bet.number` and `bet.data["number"]` are different:
> 
> ```python
> import numpy as np
> from astropy.table import Table
> 
> table_description = np.array([["number", float, None, "number"],
>     ["ra", float, "degree", "ra"],
>     ["dec", float, "degree", "dec"],
>     ["id", int, None, "id"]
> ])
> 
> tab = Table(data=[[1], ["abc"], [1.0], [1.0], [5.1]], names=("id", "tex
>     ...: t", "ra", "dec", "number"), units=(None, None, "degree", "degree", None
>     ...: ))
> 
> bet = BaseEnhancedTable(table_description, data=tab)
> 
> bet.number
> 
> # Returns
> # <Column name='number' dtype='float64' length=1>
> # 5.1
> 
> bet.number = [6.1]
> bet.number
> 
> # Returns
> # [6.1]
> 
> bet.data["number"]
> 
> # Returns
> # <Column name='number' dtype='float64' length=1>
> # 5.1
> ```

This doesn’t surprise me, you can take any variable and replace its value with a new one.  I just created an attribute that pointed to a astropy table column.  Resetting that attribute to a list breaks that pointer.  

Does making it a property get around this?  Maybe making it immutable would?  You could still define a column’s values as long as you treat it like a pre-existing astropy table column.

JuanCab:
I have uploaded a complete rewrite pair of data classes constructed based on feedback from Astropy folks at SciPy2023.  This is what caused so much grumbling during the sprints, but it is done.

JuanCab:
On the flight I hit on the solution for computing the ‘night’, so that’s taken care of as well.

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    table_description: dict
+        This is a dictionary where each key is a required table column name
+        and the value corresponding to each key is the required dtype
+        (can be None).  This is used to check the format of the input data
+        table.
+
+    data: `astropy.table.Table`
+        A table containing astronomical data of interest.  This table
+        will be checked to make sure all columns listed in table_description
+        exist and have the right runits. However, additional columns that
```
mwcraig, line 98:
```suggestion
        exist and have the right units. Additional columns that
```

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    def __init__(self, table_description, data):
+        # Handle parameters
+        self._table_description = table_description
+        self.data = data.copy()
+
+        # Build the data table
+        if not isinstance(self.data, Table):
+            raise TypeError(f"You must provide an astropy Table as data (it is type {type(self.data)}).")
+        else:
+            # Using the BaseTimeSeries class to handle the data table means required columns
```
mwcraig, line 118: Just to double check, are there any required columns for `BaseEnhancedTable`?

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+            # are checked.
+            super().__init__(data=self.data)
+
+            # Check that self.data exists now
```
mwcraig, line 122: This seems like a check that should be in a unit test. If we have a test that, after intitializing, has a `.data` attribute then we are good to go.

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+            if self.data is None:
+                raise ValueError(f"self.data should exist at this point, but doesn't, crap.")
+
+            # Confirm a proper table description is passed
+            if not isinstance(self._table_description, dict):
```
mwcraig, line 127: I don't think we need to be this restrictive. The only thing we really need is that `table_description` is accessible via keys. We may want to make sure this is a dictionary for us, so we could do something like this up where you currently set `self._table_description = table_description` (lines below are untested, meant to illustrate the idea)

```python
try:
    self._table_description = {k: v for k, v in table_description.items()}
except AttributeError:
    raise TypeError(...)
```

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+    Attributes
+    ----------
+    data: `astropy.table.Table`
+        A table formatted to match the table_description formatting information
+        but containing no data.
+    """
+
+    def __init__(self, table_description, data):
+        # Handle parameters
+        self._table_description = table_description
```
mwcraig, line 111: See below for a suggestion about this. I think the `TypeError` check should be a little different and be up here before the rest of the processing happens.

Also, I think as currently written `self._table_description` will modify the `dict` that the user passes in. Haven't checked that, though.

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        A dictionary containing instrumental passband names as keys and
+        AAVSO passband names as values. This is used to automatically
+        update the passband column to AAVSO standard names if desired.
+
+    retain_user_computed: bool, optional (Default: False)
+        If True, any computed columns (see USAGE NOTES below) that already
+        exist in `data` will be retained.  If False, will throw an error
+        if any computed columns already exist in `data`.
+
+    USAGE NOTES: If you input a data file, it MUST contain the following columns
```
mwcraig, line 170: I still think the "USAGE NOTES" should be in the "Notes" section of the documentation, not here.

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        'noise' : None,
+        'exposure' : u.s,
+        'date-obs' : None,
+        'airmass' : None,
+        'passband' : None,
+        'file' : None
+    }
+
+    def __init__(self, observatory, camera, data,  passband_map=None, retain_user_computed=False):
+        # Set metavariables
```
mwcraig, line 257: What do you mean be metavariables?

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        self.observatory = observatory
+        self.camera = camera
+        self.passband_map = passband_map
+
+        # Check the time column is correct format.
+        # Correct dtype and class
+        if (data['date-obs'].dtype != object) or (data['date-obs'][0].__class__ != Time):
+            raise ValueError(f"data['date-obs'] is not column of astropy.time.Time objects.")
+        # Correct scale is set for first element
+        if (data['date-obs'][0].scale != 'utc'):
```
mwcraig, line 267: Can this be consolidated with the previous `if` like this:

```python
try:
    if (data['date-obs'][0].scale != 'utc'):
        raise ValueError(f"data['date-obs'] astropy.time.Time must have scale='utc', not \'{data['date-obs'][0].scale}\'.")
# Happens if the first item does not have a "scale"
except AttributeError:
    raise ValueError(f"data['date-obs'] is not column of astropy.time.Time objects.")

File: stellarphot/core.py [hunk 34]
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+
+        # Check that self.data exists now
+        if self.data is None:
+            raise ValueError(f"self.data should exist at this point but doesn't, crap.")
+
+        # Compute additional columns (not done yet)
+        computed_columns = ['aperture_area', 'annulus_area', 'snr', 'bjd', 'night',
+                            'mag_inst', 'mag_error']
+
+        # Check if columns exist already, if they do and ignore_computed is False, throw an error
```
mwcraig, line 290:
```suggestion
        # Check if columns exist already, if they do and retain_user_computed is False, throw an error
```

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+        for this_col in computed_columns:
+            if this_col in self.data.colnames:
+                if not retain_user_computed:
+                    raise ValueError(f"Computed column '{this_col}' already exist in data.  If you want to keep them, pass retain_user_computed=True to the initializer.")
+            else:
+                # Compute the columns that need to be computed (switch requries python 3.10)
+                match this_col:
```
mwcraig, line 297: I'm fine using `match` but IIRC there is some "gotcha" involved in the python implementation of there is that can make it tricky. In some subtle way that escapes me at the moment it differs from a switch in C.

stellarphot/core.py
```diff
@@ -62,3 +70,305 @@ def __init__(self, gain=1.0 * u.electron / u.adu,
...
+
+    def add_bjd_col(self):
+        """
+        Returns a astropy column of barycentric Julian date times corresponding to
+        the input observations.  It modifies that table in place.
+
+        Parameters
+        ----------
+
+        phot_table: `astropy.table.Table`
```
mwcraig, line 357: I don't think `phot_table` is actually a parameter...

mwcraig (changes requested):
Excellent progress! There are a couple suggestions inline for some changes. We should also talk briefly about how data gets added to this structure by `stellarphot` as it is doing photometry.

stellarphot/core.py
JuanCab: None in the BaseEnhancedTable, since the approach used by BaseTimeSeries was to use a decorator on the class to handle that.

stellarphot/core.py
JuanCab: I fixed this by copying the dict.

stellarphot/core.py
JuanCab: It is a check I put in to find the bug in the creation of self.data, it is being removed.

stellarphot/core.py
JuanCab: OK, I moved it.

stellarphot/core.py
JuanCab: Originally I thought of these as metadata describing the data .  I just changed the comment to call it attributes.

stellarphot/core.py
JuanCab: OK, fixed.

stellarphot/core.py
JuanCab: Well, thus far, this one seems to be working.

stellarphot/core.py
JuanCab: True enough, I think I was originally thinking of passing it and realized I didn't need to.

JuanCab (commented):
I just discovered the required column checking I thought was being done is not, so still need to debug this.

mwcraig (approved):

PR #138, opened 2023-07-21
New source_detection routine and revised source list class

JuanCab:
So I have uploaded this code so you can look at it.  The source_detection routine is written, but I had to re-work the tests a bit to expect a QTable output.  I disabled those tests that take source_detection output and put it in photutils_stellar_photometry() since those will fail catastrophically.

Issues to consider:
- I have not yet added a padding around the edges to ignore for source detection.
- We want to run the DAOStarFinder on the image with background subtracted, but when I actually subtracted the expected background sky_per_pix_avg (instead of the image mean), we got spurious dection of sources along the edges of the image.


mwcraig:
I'm about to do a force push to this branch with the changes rebased on upstream/main so that this only shows the changes made in source detection.

The easiest next step on your end would be to delete (or rename) your local copy of the branch and then pull from your JuanCab/revise_photometry_module

mwcraig:
Also, I'd like to keep this one to just source detection and handle other changes to photometry in a separate PR after we get this one sorted out.

mwcraig:
> We want to run the DAOStarFinder on the image with background subtracted, but when I actually subtracted the expected background sky_per_pix_avg (instead of the image mean), we got spurious dection of sources along the edges of the image.

Figured this one out I think. The noise image is created in this snippet (I've removed a couple lines for clarity)

```python
        self.mean_noise = self.sources['amplitude'].max() / 100
        self.noise_dev = noise_dev
        # [snip]
        self._noise = make_noise_image(self._stars.shape,
                                       mean=self.mean_noise,
                                       stddev=noise_dev)
```

This makes an image whose pixels follow a Gaussian distribution with mean `self.mean_noise` and deviation `noise_dev`. The appropriate sky per pixel should have been `self.mean_noise` from the beginning, not `noise_dev`. The latter is basically the read noise, while the former is the sky background.

Sorry about that 😬...

stellarphot/photometry/source_detection.py
```diff
@@ -141,52 +145,130 @@ def compute_fwhm(ccd, sources, fwhm_estimate=5,
...
+    sources = daofind(ccd - sky_per_pix_avg)
+    # daofind should be run on background subtracted image
+    # (fails, or at least returns garbage, if sky_per_pix_avg is too low)
+    src_cnt = len(sources)
+    print(f"source_detection: {src_cnt} sources identified.")
+    #print(sources)
+
+    # If image as WCS, compute RA and Dec of each source
+    try:
+        sources['ra'], sources['dec'] = ccd.wcs.all_pix2world(sources['xcentroid'],
```
mwcraig, line 226: The preferred way to call this has changed. It is now

```suggestion
        sources['ra'], sources['dec'] = ccd.wcs.pixel_to_world(sources['xcentroid'],
```

stellarphot/photometry/source_detection.py
```diff
@@ -141,52 +145,130 @@ def compute_fwhm(ccd, sources, fwhm_estimate=5,
...
+    if sky_per_pix_avg is None:
+        sky_per_pix_avg = mean
+        print(f"source_detection: sky_per_pix_avg set to {sky_per_pix_avg:.4f}")
+
+    # Identify sources applying DAOStarFinder to a "sky subtracted"
+    # image.
+    print(f"source_detection: threshold set to {threshold}* standard deviation "
+          f"({std:.4f})")
+    print(f"source_detection: Assuming fwhm of {fwhm} for DAOStarFinder")
+    daofind = DAOStarFinder(fwhm = fwhm, threshold = threshold * std)
```
mwcraig, line 216:
```suggestion
    daofind = DAOStarFinder(fwhm=fwhm, threshold=threshold * std)
```

stellarphot/photometry/source_detection.py
mwcraig: This may also change the return type too -- it may be coordinate objects rather than separate RA and Dec

stellarphot/photometry/source_detection.py
```diff
@@ -70,29 +74,29 @@ def compute_fwhm(ccd, sources, fwhm_estimate=5,
     sources : `astropy.table.Table`
         An astropy table of the positions of sources in the image.
 
-    fwhm_estimate : float, optional
+    fwhm_estimate : float, optional (default=5)
```
mwcraig, line 77: I think we don't need the default here explicitly since it is in the function signature...

mwcraig (changes requested):
Just minor changes 😃

stellarphot/photometry/source_detection.py
JuanCab: OK, I fixed this.

stellarphot/photometry/source_detection.py
```diff
@@ -141,52 +145,130 @@ def compute_fwhm(ccd, sources, fwhm_estimate=5,
...
+    # daofind should be run on background subtracted image
+    # (fails, or at least returns garbage, if sky_per_pix_avg is too low)
+    src_cnt = len(sources)
+    print(f"source_detection: {src_cnt} sources identified.")
+    #print(sources)
+
+    # If image as WCS, compute RA and Dec of each source
+    try:
+        sources['ra'], sources['dec'] = ccd.wcs.all_pix2world(sources['xcentroid'],
+                                                              sources['ycentroid'], 0)
```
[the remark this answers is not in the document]
JuanCab: Fixed.

stellarphot/photometry/source_detection.py
```diff
@@ -141,52 +145,130 @@ def compute_fwhm(ccd, sources, fwhm_estimate=5,
...
+        sources['fwhm_x'][bad_src] = np.nan
+        sources['fwhm_y'][bad_src] = np.nan
+        sources['width'][bad_src] = np.nan
+    else: # add empty columns
+        sources['fwhm_x'] = np.nan * np.ones(src_cnt)
+        sources['fwhm_y'] = np.nan * np.ones(src_cnt)
+        sources['width'] = np.nan * np.ones(src_cnt)
+
+    # Convert sources to SourceListData object by adding
+    # unirs to the columns
```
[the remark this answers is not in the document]
JuanCab: fijed!

JuanCab (commented):
Uploaded fixes to address concerns.

mwcraig (approved):
Thanks, looks good!

PR #139, opened 2023-07-25
Fixed data classes to behave like Tables for slicing, etc.

JuanCab:
I fixed the data classes (and their tests) to allow their superclass 'QTable' behaviors to work by changing it so they just call the superclass __init__ with kwargs if none of our data is passed in.  Renamed our data variable to input_data to avoid conflicts with built-in Table behaviors.


stellarphot/core.py
```diff
@@ -219,19 +234,20 @@ class PhotometryData(BaseEnhancedTable):
...
 
     Parameters
     ----------
 
-    observatory: `astropy.coordinates.EarthLocation`
+    observatory: `astropy.coordinates.EarthLocation` (Default: None)
         The location of the observatory.
 
-    camera: `stellarphot.Camera`
+    camera: `stellarphot.Camera` (Default: None)
```
mwcraig, line 247:
```suggestion
    camera: `stellarphot.Camera`, optional
```

stellarphot/core.py
```diff
@@ -460,15 +487,15 @@ class CatalogData(BaseEnhancedTable):
...
     Parameters
     ----------
-    data: `astropy.table.Table`
+    input_data: `astropy.table.Table` (default: None)
         A table containing all the astronomical catalog data to be validated.
         This data is copied, so any changes made during validation will not
         affect the input data, only the data in the class.
 
-    name: str
+    catalog_name: str (default: None)
```
mwcraig, line 495:
```suggestion
    catalog_name: str, optional
```

stellarphot/core.py
```diff
@@ -460,15 +487,15 @@ class CatalogData(BaseEnhancedTable):
...
         User readable name for the catalog.
 
-    data_source: str
+    catalog_source: str (default: None)
```
mwcraig, line 498:
```suggestion
    catalog_source: str, optional
```

stellarphot/core.py
```diff
@@ -539,7 +579,7 @@ class SourceListData(BaseEnhancedTable):
 
     Parameters
     ----------
-    data: `astropy.table.Table`
+    data: `astropy.table.Table` (Default: None)
```
mwcraig, line 582:
```suggestion
    data: `astropy.table.Table`, optional
```

stellarphot/tests/test_core.py
```diff
@@ -146,31 +146,28 @@ def test_camera_unitscheck():
...
                              fast_reader=False)
 
+
 def test_base_enhanced_table_blank():
-    # This should raise error because you are attempting a blank data set
-    with pytest.raises(TypeError):
-        test_base = BaseEnhancedTable(test_descript)
+    # This should just return a blank BaseEnhancedTable
+    test_base = BaseEnhancedTable()
+    assert type(test_base) == BaseEnhancedTable
```
mwcraig, line 153: The preferred idiom for type check is to use isinstance instead:

```suggestion
    assert isinstance(test_base, BaseEnhancedTable)
```

stellarphot/tests/test_core.py
```diff
@@ -179,25 +176,33 @@ def test_base_enhanced_table_missing_badunits():
...
     with pytest.raises(TypeError):
-        test_base3 = BaseEnhancedTable(test_descript, test_base2)
+        test_base3 = BaseEnhancedTable(table_description=test_descript,
+                                       input_data=test_base2)
+
+
+def test_photometry_blank():
+    # This should just return a blank PhotometryData
+    test_base = PhotometryData()
+    assert type(test_base) == PhotometryData
```
mwcraig, line 198:
```suggestion
    assert isinstance(test_base, PhotometryData)
```

stellarphot/core.py
```diff
@@ -104,14 +104,14 @@ class BaseEnhancedTable(QTable):
     Parameters
     ----------
 
-    table_description: dict or dict-like
+    table_description: dict or dict-like (Default: None)
```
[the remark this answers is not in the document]
JuanCab: OK, I felt like I should not state it is optional since it's only left that way to have QTable behavior inheritance to work right. :/, but yeah, technically true.  Fixing.

stellarphot/core.py
```diff
@@ -104,14 +104,14 @@ class BaseEnhancedTable(QTable):
...
         This is a dictionary where each key is a required table column name
         and the value corresponding to each key is the required dtype
         (can be None).  This is used to check the format of the input data
         table.  Columns will be output in the order of the keys in the dictionary
-        with any additional columns tacked on the end. 
+        with any additional columns tacked on the end.
 
-    data: `astropy.table.QTable`
+    input_data: `astropy.table.QTable` (Default: None)
```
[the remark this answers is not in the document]
JuanCab: Fixed.

stellarphot/core.py
```diff
@@ -124,52 +124,67 @@ class BaseEnhancedTable(QTable):
...
+                    f"table_description is type {type(self._table_description)}).")
+
+            # Check data before copying to avoid recusive loop and non-QTable
+            # data input.
+            if not isinstance(input_data, Table) or isinstance(input_data,
+                                                               BaseEnhancedTable):
+                raise TypeError("You must provide an astropy Table and NOT a "
+                                "BaseEnhancedTable as input_data (currently of "
+                                f"type {type(input_data)}).")
+            else:
```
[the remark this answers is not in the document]
JuanCab: OK, fixed.

stellarphot/core.py
```diff
@@ -219,19 +234,20 @@ class PhotometryData(BaseEnhancedTable):
...
-    additional attributes describing the observatory and camera.
+    format including units.  It returns an `PhotometryData` which is
+    a `astropy.table.QTable` with additional attributes describing
+    the observatory and camera.
 
     Parameters
     ----------
 
-    observatory: `astropy.coordinates.EarthLocation`
+    observatory: `astropy.coordinates.EarthLocation` (Default: None)
```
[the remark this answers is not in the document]
JuanCab: Fixed.

stellarphot/core.py
JuanCab: Fixed.

stellarphot/core.py
```diff
@@ -460,15 +487,15 @@ class CatalogData(BaseEnhancedTable):
 
     Parameters
     ----------
-    data: `astropy.table.Table`
+    input_data: `astropy.table.Table` (default: None)
```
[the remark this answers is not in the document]
JuanCab: Fixed.

stellarphot/core.py
JuanCab: Fixed.

stellarphot/core.py
JuanCab: Fixed.

stellarphot/core.py
JuanCab: Fixed.

stellarphot/core.py
```diff
@@ -219,19 +234,20 @@ class PhotometryData(BaseEnhancedTable):
...
-    observatory: `astropy.coordinates.EarthLocation`
+    observatory: `astropy.coordinates.EarthLocation` (Default: None)
         The location of the observatory.
 
-    camera: `stellarphot.Camera`
+    camera: `stellarphot.Camera` (Default: None)
         A description of the CCD used to perform the photometry.
 
-    data: `astropy.table.QTable`
+    input_data: `astropy.table.QTable` (Default: None)
```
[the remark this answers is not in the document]
JuanCab: Fixed.

JuanCab:
Applied all your suggestions except "can you add a test that tries slicing something with attributes to see if the attributes survive the slicing?"  Working on that now.

JuanCab:
I have now fixed the handling of attributes during slicing.  In a nutshell, I use the approach taken by TableAttributes of storing the information in meta, but I manually created @property functions to report them as attributes.  This avoided the problem with TableAttributes wreaking the __dict__ and not working in general for our setup (still don't know why it doesn't work).

stellarphot/tests/test_core.py
mwcraig: It looks like this change didn't get applied?

mwcraig (changes requested):
Excellent to see this working!

One minor line change, and one bigger question: right now we put in a Camera or an Observatory but can't get those objects back out. The properties you added let us get at the individual attributes of those, but what about adding an additional property that returns the objects?

something like this:

```python
@property
def camera(self):
    return Camera(gain=self.gain, read_noise=self.read_noise, ....)
```

That way we can continue to work with those objects.

mwcraig (approved):

PR #141, opened 2023-07-27
`single_image_photometry()` implemented (including passing previous tests)

### LINKED ISSUE #105: Function `find_too_close()` never used
Function `find_too_close(star_locs, aperture_rad, pixel_scale=0.563)` in `photometry.py` doesn't appear to be called from anywhere in the stellarphot package.  Purge or do we need to use it!

### LINKED ISSUE #131: Is `photutils_stellar_photometry` used anywhere?
This function doesn't appear to be used anywhere and actually produces a table with different tables than `photometry_on_directory`.

JuanCab:
So I have created `single_image_photometry()` and removed `photutils_stellar_photometry()`.  The new function handles input in the form of a `SourceListData` object and a `CCDData` object, sets up all the units properly and ultimately outputs a `PhotometryData` object.

The new function can handle a source list with just x/y positions or just ra/dec positions.  It also handles a range of exposure keywords.

The only concession to it running within another function (the photometry_on_directory() function) is that it returns two items, a `PhotometryData` object and a list of `out_of_bounds_sources`.

In the process of doing all this photometry work, I caught an error in the `PhotometryData` unit handling in that some of our counts are counts per pixel^2, this has been fixed.  This required changes to the tests of core as well.

**Edit**: Closes #105 Closes #131


stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+    anuls = CircularAnnulus(aper_locs, r_in=inner_annulus, r_out=outer_annulus)
+
+    # Perform the aperture photometry
+    photom = aperture_photometry(ccd_image.data, (apers, anuls),
+                                mask=ccd_image.mask, method='center')
+
+    # Add source ids to the photometry table
+    photom['star_id'] = star_ids
+    photom['ra'] = ra
+    photom['ra'].unit = u.deg
```
mwcraig, line 275: Can this be done in a single step: `photom['ra'] = ra * u.deg`?

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+    photom['dec'] = dec
+    photom['dec'].unit = u.deg
+
+    # Drop ID column from aperture_photometry()
+    del photom['id']
```
mwcraig, line 280: 🎉

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+
+    # Add various CCD image parameters to the photometry table
+    if fname is not None:
+        photom['file'] = fname
+    else:
+        photom['file'] = [''] * len(photom)
+
+    # Search for exposure keyword in header and set exposure if found
+    exp_kw = ["EXPOSURE", "EXPTIME", "TELAPSE", "ELAPTIME", "ONTIME", "LIVETIME"]
```
mwcraig, line 289: Let's make this a constant up near the top of the file -- that way it is a little easier to find if we need to edit it. The naming convention for constants is all caps, so `EXP_KW = ....`

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+    matched_kw = None
+    for kw in exp_kw:
+        if kw in ccd_image.header:
+            matched_kw = kw
+            break
+
+    if matched_kw is None:
+        raise ValueError("single_image_photometry: None of the allowed exposure "
```
mwcraig: The function name will be included in the stack trace...

```suggestion
        raise ValueError("None of the allowed exposure "
```

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
-    # annulus objects
-    n_pix_ap = apertures.area
-    n_pix_ann = annulus.area
+    try:
+        photom['date-obs'] = Column(data=Time([ccd_image.header['DATE-OBS']]
+                                              * len(photom),
+                                    format='isot', scale='utc'), name='date-obs')
+    except KeyError:
+        raise ValueError("single_image_photometry: 'DATE-OBS' not found in CCD image "
+                         "header ... SKIPPING THIS IMAGE!")
```
mwcraig: Could these checks that result in skipping the image be move earlier to save a bit of time? No reason to run `aperture_photometry` if we are going to skip it.

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+        except AttributeError:
+            print('single_image_photometry: BAD ANNULUS, stats set to np.nan!')
+            avg_sky_per_pix, med_sky_per_pix, std_sky_per_pix = \
+                np.nan, np.nan, np.nan
+        photom['sky_per_pix_avg'] = avg_sky_per_pix * u.pixel**-1
+        photom['sky_per_pix_med'] = med_sky_per_pix * u.pixel**-1
+        photom['sky_per_pix_std'] = std_sky_per_pix * u.pixel**-1
+        print('                          ...DONE computing clipped sky stats')
+    else: # Don't reject outliers (but why would you do this?)
+        print("single_image_photometry: WARNING: Computing sky per pixel without "
```
mwcraig: If we want a warning we should using `warning`

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
-    try:
-        ra, dec = ccd_image.wcs.all_pix2world(coords[0], coords[1], 0)
-        phot_table['RA_center'] = ra
-        phot_table['Dec_center'] = dec
-    except AttributeError:
-        pass
+            med_pp.append(np.median(annulus_data))
+            std_pp.append(np.std(annulus_data))
+        photom['sky_per_pix_avg'] = photom['annulus_sum'] / photom['annulus_area']
+        photom['sky_per_pix_med'] = np.array(med_pp) * ccd_image.unit * u.pixel**-1
```
mwcraig, line 360: Is there a reason not to use ` / u.pixel` instead of `* u.pixel**-1`?

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+                                    photom['sky_per_pix_avg'].value))
+    photom['aperture_net_cnts'].unit = ccd_image.unit
+
+    # Fit the FWHM of the sources
+    print('single_image_photometry: fitting fwhm of all sources ...')
+    fwhm_x, fwhm_y = compute_fwhm(ccd_image, photom, fit=fwhm_by_fit)
+    print('                          ...DONE fitting fwhm of all sources')
+
+    # Deal with bad FWHM values
+    bad_fwhm = (fwhm_x<1) | (fwhm_y<1) # Set bad values to NaN now
```
mwcraig, line 375:
```suggestion
    bad_fwhm = (fwhm_x < 1) | (fwhm_y < 1) # Set bad values to NaN now
```

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+    photom['fwhm_y'] = fwhm_y * u.pixel
+    photom['width'] = ((fwhm_x + fwhm_y) / 2) * u.pixel
 
-    # Obtain flux error and add column to return table
-    noise = np.sqrt(gain * net_flux + n_pix_ap * (1 + (n_pix_ap / n_pix_ann)) *
-                    (gain * (bkgd_pp + N_dark_pp) + (N_R**2)))
-    phot_table['aperture_sum_err'] = noise
+    # Flag sources with bad counts before computing noise
+    # This can happen, for example, when the object is faint
+    # and centroiding is bad.
```
mwcraig, line 384:
```suggestion
    # and centroiding is bad. It can also happen when the sky background is low
    # but the noise is large.
```

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+    bad_cnts = photom['aperture_net_cnts'].value < 0
+    all_bads = bad_cnts | bad_fwhm
+    photom['aperture_net_cnts'][all_bads] = np.nan
+
+    # Compute and save noise
+    noise = calculate_noise(gain=camera.gain.value,
```
mwcraig, line 390: Seems like we should be passing in the `Camera` object instead of all of its attrtibutes.

mwcraig: Also, can we pass in the `photom` table  and have `calculate_noise` extract what it needs from there?

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,421 @@
...
+    photom['mag_error'] = 1.085736205 / snr
 
-    # Obtain flux error and add column to return table
-    noise = np.sqrt(gain * net_flux + n_pix_ap * (1 + (n_pix_ap / n_pix_ann)) *
-                    (gain * (bkgd_pp + N_dark_pp) + (N_R**2)))
-    phot_table['aperture_sum_err'] = noise
+    # Compute instrumental magnitudes
+    photom['mag_inst'] = \
+        (-2.5 * np.log10(camera.gain.value * photom['aperture_net_cnts'].value /
+                            photom['exposure'].value))
```
mwcraig, line 413: I have a mild preference for

```python
photom['mag_inst'] = (
    -2.5 * np.log10(camera.gain.value * photom['aperture_net_cnts'].value /
                            photom['exposure'].value)
)
```

stellarphot/photometry/tests/fake_image.py
```diff
@@ -0,0 +1,65 @@
...
+        base_data = FakeImage()
+        super().__init__(base_data.image.copy(), unit='adu')
+        # Add some additional features to the CCDData object, like
+        # a header and the sources used to create the image.abs
+        self.header = fits.Header()
+        self.header['EXPOSURE'] = 1.0
+        self.header['DATE-OBS'] = '2018-01-01T00:00:00.0'
+        self.header['AIRMASS'] = 1.2
+        self.header['FILTER'] = 'V'
+        self.sources = base_data.sources.copy()
```
mwcraig, line 62: How about just setting `self.base_data = data` and getting at `noise_dev`, etc via `FakeCCDImage.base-Data.noise_dev`?

mwcraig (changes requested):
Thanks for the PR. 

Feel free to mark as resolved any changes you decide to merge in, but if you could leave onse you respond to with a comment unresolved that will make it easier for me to find them. 😀 

Several line comments but a couple larger picture ones:

+ I'd prefer that ra/dec be used if present over x/y because the current behavior is to prefer ra/dec.
+ In several of the `try`/`except`s that result in a `ValueError` the message says something about "skipping" the image. Should the result here be a warning instead of an error?

JuanCab:
I have been marking resolved those I agreed with you on (or accepted your suggestion).

Also, I threw errors because I was thinking of running this inside multi_image_photometry() and catching the errors.  It would be just as easy to do a warning and return None, None.  Of course, I've never done a formal warning in Python and will have to look that up.

JuanCab:
You can replicate that prefer RA/Dec behavior by dropping the x/y columns (I do that in multi_image_photometry() before calling single_image_photometry()).  I did it this way because I wanted it to be clear the photometry could work for a single field WITHOUT WCS if an amateur wanted to do it with only x/y.  And if the default in source_detection() is to return x/y and computer RA/Dec which makes x/y the measured quantity.  Does that make sense?

mwcraig:
What about adding a `use_coordinates` or something to force the use of that column? That seems easier for the user than dropping columns and them maybe having to add them back.

stellarphot/photometry/photometry.py
JuanCab: photom is not the final PhotometryData instance, so we can't do that, but I re-wrote it to use the Camera object (as well as re-writing all the tests).

stellarphot/photometry/photometry.py
JuanCab: Since I have no preferences here, change made.

stellarphot/photometry/tests/fake_image.py
JuanCab: I'm going to ask you to defer on this discussion until the multi_image_photometry() version gets folded in because this was re-worked a bit there.  I will say, I wanted FakeCCDImage to be hot swappable for FakeImage, which is why I kept somewhat the same structure for the non-CCDData related attributes.

JuanCab (commented):
I've uploaded a set of changes to this branch addressing most of what was discussed.  I would ask you defer on the raising of errors until I give you the multi_image_photometry() branch, since I will have to make the edits there as well.

JuanCab:
> What about adding a `use_coordinates` or something to force the use of that column? That seems easier for the user than dropping columns and them maybe having to add them back.

OK, I will work on this functionality.  I'll get another patch to this branch in shortly.

JuanCab:
I have posted an update to this branch added `use_coordinates` as a parameter for `single_image_photometry()`.

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+        noise for each observation.  If ``False``, only the Poisson noise from
+        the source and the sky will be included.
+
+    passband_map: dict, optional (Default: None)
+        A dictionary containing instrumental passband names as keys and
+        AAVSO passband names as values. This is used to rename the passband
+        entries in the output photometry table to be AAVSO standard versus
+        whatever is in the source list.
+
+    fwhm_by_fit : bool, optional (default: True)
```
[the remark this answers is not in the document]
mwcraig: I think you use `filter_warnings` from the standard lib `warnings` but could be wrong.

I'd be inclined to make the default `False`, though, so users have to opt in to the slow stuff.

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+    if (shift_tolerance<=0):
+        raise ValueError("shift_tolerance must be greater than 0 (should be on order "
+                         "of FWHM)")
+    if (max_adu<=0):
+        raise ValueError("max_adu must be greater than 0")
+
+    # Set hot pixels to NaN
+    ccd_image.data[ccd_image.data > max_adu] = np.nan
+
+    # Extract necessary values from sourcelist structure
```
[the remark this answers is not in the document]
mwcraig: I meant why not do this:

```python
star_ids = sourcelist['star_id']  # no .value
xs = sourcelist['xcenter']  # no .value
...
```

Some of those things will have units but it should work out in the end (maybe).

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+                            reject_too_close=True,
+                            passband_map=None,
+                            reject_background_outliers=True,
+                            fwhm_by_fit=True, fname=None):
     """
-    Perform aperture photometry on an image, with a few options for estimating
-    the local background from an annulus around the aperture.
+    Perform aperture photometry on a single image, with an options for estimating
+    the local background from sigma clipped stats of the counts in an annulus around
+    the aperture.  If the sky positions of the sources are in the sourcelist and not
```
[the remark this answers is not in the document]
mwcraig: Opening an issue to track this

mwcraig (changes requested):
Just a couple minor things -- if there any that are more substantial let's open issues for them. I think I got the two from this review and put them in #144.

stellarphot/photometry/photometry.py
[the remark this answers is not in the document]
JuanCab: OK.

stellarphot/photometry/photometry.py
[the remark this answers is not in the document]
JuanCab: Honestly, I was trying to avoid the issue with units in some of the later steps.  For example, when I remove the .value, I get test failures because the selection of `out_of_bounds` fails since I am mixing values with units and no units.

stellarphot/photometry/photometry.py
```diff
@@ -1,138 +1,408 @@
...
+            # Recentroid the sources using the WCS-derived positions (this is
+            # particularly useful is processing multiple images of the same field
+            # and just passing the same sourcelist when calling single_image_photometry
+            # on each image.
+            try:
+                xcen, ycen = centroid_sources(ccd_image.data,
+                                            imgpos[0],
+                                            imgpos[1],
+                                            box_size=2 * aperture_radius + 1)
+                # Calculate offset between centroid in this image and the positions
```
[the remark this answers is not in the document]
JuanCab: OK, I have addressed this.

stellarphot/photometry/photometry.py
JuanCab: Yeah, I fixed a couple of these as I moved to the multi_image_photometry() code.

stellarphot/photometry/photometry.py
JuanCab: This got revised in the multi_image_photometry to `suggestion` to avoid that confusion.

stellarphot/photometry/photometry.py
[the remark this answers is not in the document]
JuanCab: I fixed the warning messages and the entire problem with crashing in the multi_image_photometry() update. ;)

stellarphot/photometry/photometry.py
JuanCab: They were done here because I was adding them to the photom table which is not created until aperture_photometry is run.  I will fix this in the multi_image_photometry() version of photometry.py.

mwcraig (approved):

PR #150, opened 2023-08-04
Added logging to photometry functions (and two other tweaks)

### LINKED ISSUE #148: Logging in the  photometry functions
The new photometry functions have realy detailed progress output. It would be nice if

- [ ] There was an opt-in or opt-out mechanism (e.g. a `verbose` argument)
- [x] Consider using python logging instead of prints for the messaging

JuanCab:
I was able to add an option to log messages in `single_image_photometry` and `multi_image_photometry` to a file.  In fact, it just uses logging for all the relevant messages and returns the log to the console UNLESS the user has asked for a log file instead.

Turns out my major hassle was `ImageFileCollection` also uses logging and leaves a spare root logging handler active.  It was spewing extra copies of logging messages to screen. :-/

**Question**: Logging is currently to console OR to a file.  Do we want to log to console always?  I don't like not offering any feedback.  With the FITSFixedWarnings suppressed now (see below), the number of console messages isn't really overwhelming.

The two other tweaks:

1. **Removed FITSFixedWarnings**: Code now suppresses those annoying `FITSFixedWarning` warnings that occur every time a FITS file is loaded, they are ignored by the photometry routines since we know their cause is essentially harmless.
2. **Allowed Unmatched Sources**: There is now an `reject_unmatched` parameter (that defaults to True) for the `multi_image_photometry` that allows you to control whether or not to require a source be on all the images to be included in the final returned dataset.  It occurred to me someone could be processing images where their source is close to their observing limit.  In those situations they may want to keep all the observations, even the non-detections.

**EDIT:** Closes #148


stellarphot/photometry/photometry.py
```diff
@@ -187,8 +196,31 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
         raise ValueError(f"input_coordinates ({use_coordinates}) must be either "
                          "'pixel' or 'sky'.")
 
-    # Check CCDData headers before proceeding
+    # Set up logging (retrieve a logger but purge any existing handlers)
+    root_logger = logging.getLogger()
+    for handler in root_logger.root.handlers[:]:
+        root_logger.root.removeHandler(handler)
```
mwcraig, line 202: I assume this was because of the root handler in `ccdproc`? It seems like removing other handlers might mess with other people's stuff? 

I don't have any better ideas, and am not sure what, if anything, to change here.

stellarphot/photometry/photometry.py
```diff
@@ -187,8 +196,31 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
...
+
+    logger = logging.getLogger("single_image_photometry")
+    logger.setLevel(logging.INFO)
+    for handler in logger.handlers[:]:
```
mwcraig, line 206: You have more recent logging experience than me -- is this necessary to remove some default handlers or something?

stellarphot/photometry/photometry.py
```diff
@@ -378,21 +417,24 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
...
 
     if reject_background_outliers:
-        print(f"{logline} Computing clipped sky stats ... ", end="")
+        msg = f"{logline} Computing clipped sky stats ... "
         try:
             avg_sky_per_pix, med_sky_per_pix, std_sky_per_pix = \
                     clipped_sky_per_pix_stats(ccd_image, anuls)
         except AttributeError:
-            print("BAD ANNULUS ('sky_per_pix' stats set to np.nan) ... ", end="")
+            msg += "BAD ANNULUS ('sky_per_pix' stats set to np.nan) ... "
```
mwcraig, line 425:
```suggestion
            msg += "BAD ANNULUS ('sky_per_pix' stats set to np.nan) ... "
```

stellarphot/photometry/photometry.py
```diff
@@ -646,35 +749,37 @@ def multi_image_photometry(directory_with_images,
...
-        # Remove the starid for objects not observed in every image
-        if starid_to_remove:
-            bad_rows = all_phot.loc_indices[starid_to_remove]
-            try:
-                bad_rows = list(bad_rows)
-            except TypeError:
-                bad_rows = [bad_rows]
+    # If requested, eliminate source not detected on every image by building
+    # a set of all the unique star_ids that were missing on at least one image.
+    if (reject_unmatched):
```
mwcraig, line 754: Could we rewrite this block to start with this so we can reduce the indent level?

```suggestion
    if reject_unmatched and (len(missing_sources) > 0):
```

mwcraig (changes requested):
Very nice, I like PRs that reduce lines of code!

Only one important thing to change here, I think, modifying the if statement I commented on below.

I'm not sure about  removing all the handlers but tbh don't remember logging well enough to say anything sensible.

I'll open an issue for the a couple of logging questions/issues.

stellarphot/photometry/photometry.py
JuanCab: I got the impression from the documentation that the root handler isn't really supposed to be used if you are working on logging for a specific function/class or package... just by the fact that [the docs promote to automatically set up loggers for specific modules](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial) saying 

*A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:*
```
logger = logging.getLogger(__name__)
```

That said, I agree, it wasn't necessarily a friendly thing to slam off any root loggers running, since someone might want to import our code into their code... and using the `root` logger is completely acceptable behavior.  But I don't think we can 'disable' the root handler and then re-activate it.

stellarphot/photometry/photometry.py
JuanCab: I think this may be code left over from when I was trying to understand why there were extra logging statements coming to the console.  I removed these lines, I don't think they are necessary... the logger, created the lines before, should not have handlers set up yet.

stellarphot/photometry/photometry.py
JuanCab: Fixed.

stellarphot/photometry/photometry.py
JuanCab: Good catch, fixed.

JuanCab:
BTW, did you want me to turn off console logging if a log file is provided or do you want me to log to both?  I rather think console feedback is necessary.

mwcraig:
> BTW, did you want me to turn off console logging if a log file is provided or do you want me to log to both? I rather think console feedback is necessary.

I think there needs to be a way to disable console logging. That might mean adding way to simply disable it rather than disabling only if a log file is provided.

stellarphot/photometry/photometry.py
mwcraig: Something weird is going on ccdproc uses the thing you suggested when it does logging:

```
logger = logging.getLogger(__name__)
```

JuanCab: Maybe one of the packages ccdproc imports, although a brief review of that code doesn't reveal any packages I don't use regularly.  I suspect when you create a logger for a module, since it is a child of the root logger, a root logger must be created... maybe for some reason you are creating the root handler?  I did check and I didn't see any of the mistakes I saw others point out (e.g. using the BasicConfig setup).

JuanCab:
I just pushed a commit to add an option for console login in addition to file logging.  It is True by default.

mwcraig (approved):
Excellent, thanks!

Issue #122, opened 2023-06-29
Update stellarphot/visualization/photometry_widget_functions.py to use new classes


mwcraig:
Looking at this a little more, this might not need a change. It only gets a path to aperture settings, it doesn't need the Apertures object itself, I think...

JuanCab:
Renamed issue to reflect location of code after re-organization of code.

Issue #121, opened 2023-06-29
Update photometry notebook 03 to use new classes


mwcraig:
It is the only notebook to directly call photometry_on_directory.

In addition, it currently assumes that the apertures are just an astropy table.

PR #151, opened 2023-08-08
Fixed Notebooks 1 to 3 to work with new photometry code and new data classes

JuanCab:
Updated the comparison and photometry setting widgets to work with the new data classes (notably SourceListData) and then re-wrote the first 3 photometry notebooks to work with the new data classes and photometry functions.  This addresses both open issues #121 and #122.

### DIFF
--- a/stellarphot/gui_tools/comparison_functions.py
+++ b/stellarphot/gui_tools/comparison_functions.py
@@ -16,12 +16,14 @@ try:
 except ImportError:
     from astrowidgets.ginga import ImageWidget
 
+from stellarphot import SourceListData
 from stellarphot.differential_photometry import *
 from stellarphot.photometry import *
 from stellarphot.gui_tools.seeing_profile_functions import set_keybindings
 from stellarphot.gui_tools.fits_opener import FitsOpener
 from stellarphot.io import TessSubmission, TOI, TessTargetFile
-from stellarphot.utils.comparison_utils import set_up, crossmatch_APASS2VSX, mag_scale, in_field
+from stellarphot.utils.comparison_utils import (set_up, crossmatch_APASS2VSX,
+                                                mag_scale, in_field)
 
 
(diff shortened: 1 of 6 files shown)


stellarphot/gui_tools/comparison_functions.py
```diff
@@ -378,10 +380,22 @@ def _show_label_button_handler(self, change):
 
     def _save_aperture_to_file(self, button=None, filename=''):
         if not filename:
-            filename = self.aperture_output_file
+            filename = self.photom_apertures_file
         # Export aperture file as CSV (overwrite existing file if it exists)
         try:
-            self.generate_table().write(filename, overwrite=self.overwrite_outputs)
+            # Convert aperture table into a SourceList objects and output
```
mwcraig, line 386: I think most of this (except the last `sources.write....`) should come before the `try` since we are only checking for the `OSError`.

stellarphot/gui_tools/photometry_widget_functions.py
```diff
@@ -68,6 +84,20 @@ def aperture_radius(self):
         """
         return self._aperture_radius
 
+    @property
+    def inner_annulus(self):
```
mwcraig, line 88: This is fine for now, but I'll be pushing up an aperture settings object in a bit.

mwcraig (changes requested):
Just one change

mwcraig (approved):

stellarphot/gui_tools/comparison_functions.py
JuanCab: I see you already pushed this change through.

stellarphot/gui_tools/photometry_widget_functions.py
JuanCab: OK

PR #158, opened 2023-08-17
Fixing non-NaNable ccd images

### LINKED ISSUE #157: Setting saturated pixels to `NaN` fails if image is integer
While testing some of the changes with the new photometry settings I was using an uncalibrated image. `single_image_photometetry` failed on [this line](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/photometry/photometry.py#L259) with 

```python-traceback
File ~/Development/astronomy/stellarphot/stellarphot/photometry/photometry.py:222, in single_image_photometry(ccd_image, sourcelist, camera, observatory_location, aperture_radius, inner_annulus, outer_annulus, shift_tolerance, max_adu, fwhm_estimate, use_coordinates, include_dig_noise, reject_too_close, reject_background_outliers, passband_map, fwhm_by_fit, fname, logline)
    219     return None, None
    221 # Set high pixels to NaN
--> 222 ccd_image.data[ccd_image.data > max_adu] = np.nan
    224 # Extract necessary values from sourcelist structure
    225 star_ids = sourcelist['star_id'].value

ValueError: cannot convert float NaN to integer
```

I think the fix can be just converting the data to `float`.

JuanCab:
I implemented a fix for the ccddata not being NaN-able by converting it to float if needed.  Can you check that your image will now process.

I also updated the CHANGES.rst file to cover recent changes.

Fixes #157

### DIFF
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -250,9 +250,7 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
         return None, None
 
     # Set high pixels to NaN (make sure ccd_image.data is a float array first)
-    if ccd_image.data.dtype != float:
-        ccd_image.data = ccd_image.data.astype(float)
-    ccd_image.data[ccd_image.data > max_adu] = np.nan
+    ccd_image.data[ccd_image.data.astype(float) > max_adu] = np.nan
 
     # Extract necessary values from sourcelist structure
     star_ids = sourcelist['star_id'].value
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -250,7 +250,8 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
         return None, None
 
     # Set high pixels to NaN (make sure ccd_image.data is a float array first)
-    ccd_image.data[ccd_image.data.astype(float) > max_adu] = np.nan
+    ccd_image.data = ccd_image.data.astype(float)
+    ccd_image.data[ccd_image.data > max_adu] = np.nan
 
     # Extract necessary values from sourcelist structure
     star_ids = sourcelist['star_id'].value
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -250,8 +250,8 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
         return None, None
 
     # Set high pixels to NaN (make sure ccd_image.data is a float array first)
-    if ccd_image.data.dtype != np.float:
-        ccd_image.data = ccd_image.data.astype(np.float)
+    if ccd_image.data.dtype != float:
+        ccd_image.data = ccd_image.data.astype(float)
     ccd_image.data[ccd_image.data > max_adu] = np.nan
 
     # Extract necessary values from sourcelist structure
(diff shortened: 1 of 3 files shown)


stellarphot/photometry/photometry.py
```diff
@@ -255,7 +255,9 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
                "SKIPPING THIS IMAGE!")
         return None, None
 
-    # Set high pixels to NaN
+    # Set high pixels to NaN (make sure ccd_image.data is a float array first)
+    if ccd_image.data.dtype != float:
```
mwcraig, line 259: I think you can just cast it as a `float` without checking the type first. If it is already `float` the cast just doesn't do anything. (Also, use `isinstance` for type checks :))

mwcraig (changes requested):

stellarphot/photometry/photometry.py
JuanCab: Fixed.

stellarphot/photometry/photometry.py
```diff
@@ -255,8 +255,8 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
                "SKIPPING THIS IMAGE!")
         return None, None
 
-    # Set high pixels to NaN
-    ccd_image.data[ccd_image.data > max_adu] = np.nan
+    # Set high pixels to NaN (make sure ccd_image.data is a float array first)
```
mwcraig, line 258: I think you still need a line here making the data float, it just doesn't need a type check:

```suggestion
    # Set high pixels to NaN (make sure ccd_image.data is a float array first)
    ccd_image.data = ccd_image.data.astype(float)
```

mwcraig (changes requested):

stellarphot/photometry/photometry.py
JuanCab: Didn't I cast it to float during the NaN replacement? (On iPhone, can't see code)

mwcraig: You cast it as float for the comparison (inside the brackets) but the data itself is still int I think

JuanCab:
OK, it is now fixed.  Sigh.  I would have sworn I had been clever.  Can you PLEASE test this with the integer image, just in case?

Juan


On Aug 17, 2023, at 3:35 PM, Matt Craig ***@***.***> wrote:

@mwcraig commented on this pull request.

________________________________

In stellarphot/photometry/photometry.py<https://github.com/feder-observatory/stellarphot/pull/158#discussion_r1297712936>:

> @@ -255,8 +255,8 @@ def single_image_photometry(ccd_image, sourcelist, camera, observatory_location,
                "SKIPPING THIS IMAGE!")
         return None, None

-    # Set high pixels to NaN
-    ccd_image.data[ccd_image.data > max_adu] = np.nan
+    # Set high pixels to NaN (make sure ccd_image.data is a float array first)


You cast it as float for the comparison (inside the brackets) but the data itself is still int I think

stellarphot/photometry/photometry.py
JuanCab: I think I fixed it now.

mwcraig:
I just pushed a test which I verified fails without your fix 🎉 but unfortunately found out about #161 along the way 😭 .

mwcraig (approved):

PR #162, opened 2023-08-18
Update Character Limit values

Tanner728:
Changed the character limit values for each data column to be a parameter that can be used to truncate the data automatically.


mwcraig (changes requested):
This is a minor nitpick in some sense, but can you please change all og the uppercase "Limit" to lowercase "limit"?

mwcraig (approved):
Thanks!

PR #163, opened 2023-08-22
Added a pydantic-validated Camera class and test.

JuanCab:
I have added a pydantic (v1) validated Camera class.  I also had to modify previous units checks (we now through `ValueError` instead of `TypeError`) and added a new check to see that any 'count' (e.g. count or electron or whatever) is acceptable as long as the units of read_noise, gain, and dark_current are consistent.  

I also confirmed we can json serialize the Camera object, so we can dump its value to disk and should be able to deserialize it easily enough.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -134,7 +134,7 @@ class Camera(BaseModel):
         try:
             rn_unit = Quantity(values['read_noise']).unit
         except KeyError:
-            raise ValueError("Valid keys for values are: ", values.keys())
+            raise ValueError("read_noise must be specified including units.")
 
         # Check that gain and read noise have compatible units, that is that
         # gain is read noise per adu.
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -131,6 +131,7 @@ class Camera(BaseModel):
     # When the switch to pydantic v2 happens, this root_validator will need
     # to be replaced by a model_validator decorator.
     @root_validator(skip_on_failure=True)
+    @classmethod
     def validate_gain(cls, values):
         # Get read noise units
         rn_unit = Quantity(values['read_noise']).unit
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -2,6 +2,9 @@ from astropy import units as u
 from astropy.coordinates import EarthLocation, SkyCoord
 from astropy.table import Column, QTable, Table
 from astropy.time import Time
+from astropy.units import Quantity
+
+from pydantic import BaseModel, root_validator
 
 import numpy as np
 
@@ -9,7 +12,45 @@ __all__ = ['Camera', 'BaseEnhancedTable', 'PhotometryData', 'CatalogData',
            'SourceListData']
 
 
-class Camera:
+# Approach to validation of units was inspired by the GammaPy project
+# which did it before we did:
+# https://docs.gammapy.org/dev/_modules/gammapy/analysis/config.html
+
+class QuantityType(Quantity):
+    # Validator for Quantity type
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        try:
+            v = Quantity(v)
+        except TypeError:
+            raise ValueError(f"Invalid value for Quantity: {v}")
+
+        return v
+
+
+class PixelScaleType(Quantity):
+    # Validator for pixel scale type
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        try:
+            v = Quantity(v)
+        except TypeError:
+            raise ValueError(f"Invalid value for Quantity: {v}")
+        if (len(v.unit.bases) != 2 or
+            v.unit.bases[0].physical_type != "angle" or
+            v.unit.bases[1].name != "pix" ):
+            raise ValueError(f"Invalid unit for pixel scale: {v.unit!r}")
+        return v
+
+class Camera(BaseModel):
     """
     A class to represent a CCD-based camera.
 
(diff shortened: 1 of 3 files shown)


stellarphot/tests/test_core.py
```diff
@@ -32,28 +32,44 @@ def test_camera_unitscheck():
     dark_current = 0.01 * u.electron / u.second
     pixel_scale = 0.563 * u.arcsec / u.pix
 
-    with pytest.raises(TypeError):
+    with pytest.raises(ValueError):
```
mwcraig, line 35: I'm confused about why this passes. pydantic should have turned the `ValueError` into a `ValidationError`.  I'd suggest we check for that , and add some `match` language to make sure we are getting the expected failure:

```suggestion
    with pytest.raises(ValidationError,  match="gain"):
```

stellarphot/tests/test_core.py
```diff
@@ -32,28 +32,44 @@ def test_camera_unitscheck():
...
         c = Camera(gain=gain.value,
                 read_noise=read_noise,
                 dark_current=dark_current,
                 pixel_scale=pixel_scale)
-    with pytest.raises(TypeError):
+    with pytest.raises(ValueError):
```
mwcraig, line 40: As above

stellarphot/tests/test_core.py
```diff
@@ -32,28 +32,44 @@ def test_camera_unitscheck():
...
         c = Camera(gain=gain,
                 read_noise=read_noise.value,
                 dark_current=dark_current,
                 pixel_scale=pixel_scale)
-    with pytest.raises(TypeError):
+    with pytest.raises(ValueError):
```
mwcraig, line 45: Ditto

stellarphot/tests/test_core.py
```diff
@@ -32,28 +32,44 @@ def test_camera_unitscheck():
...
         c = Camera(gain=gain,
                 read_noise=read_noise,
                 dark_current=dark_current.value,
                 pixel_scale=pixel_scale)
-    with pytest.raises(TypeError):
+    with pytest.raises(ValueError):
```
mwcraig, line 50: one more time!

stellarphot/core.py
```diff
@@ -70,31 +111,50 @@ class Camera:
...
+        json_encoders = {
+            Quantity: lambda v: f"{v.value} {v.unit}",
+            QuantityType: lambda v: f"{v.value} {v.unit}",
+            PixelScaleType: lambda v: f"{v.value} {v.unit}"
+        }
+
+    # When the switch to pydantic v2 happens, this root_validator will need
+    # to be replaced by a model_validator decorator.
+    @root_validator()
+    def validate_gain(cls, values):
```
mwcraig, line 132: Also, `root_validator` is a classmethod accoring to the documentation, so let's mark it that way:

```suggestion
    @classmethod
    def validate_gain(cls, values):
```

mwcraig (changes requested):
Despite the extensive comments I think this is really close to ready. Sorry it took a while -- I needed some uninterrupted time so I could try things locally since I'm not that familiar with pydantic or Quantity.

In addition to the line suggestions, I have one other: 

Add an "image_unit" attribute to `Camera` so that the user has the option of using whatever image unit that want (ADU, count, something that is already gain correct so that it is in electrons, say)

What do you think?

stellarphot/core.py
```diff
@@ -2,30 +2,71 @@
...
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        try:
+            v = Quantity(v)
+        except TypeError:
+            raise ValueError(f"Invalid value for Quantity: {v}")
```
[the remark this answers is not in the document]
JuanCab: Good catch, I forgot unitless is allowed.  I've implemented your fix.

stellarphot/core.py
```diff
@@ -2,30 +2,71 @@
...
```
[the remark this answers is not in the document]
JuanCab: Yeah, that's why... at least I assumed it operated that way with all validate functions getting run down the class heirarchy.

stellarphot/core.py
```diff
@@ -70,31 +111,50 @@ class Camera:
...
+            QuantityType: lambda v: f"{v.value} {v.unit}",
+            PixelScaleType: lambda v: f"{v.value} {v.unit}"
+        }
+
+    # When the switch to pydantic v2 happens, this root_validator will need
+    # to be replaced by a model_validator decorator.
+    @root_validator()
+    def validate_gain(cls, values):
+        # Get read noise units
+        try:
```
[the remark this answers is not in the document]
JuanCab: Yeah, I think this is a left over from debugging the code and my trying to figure out what was happening.  Fixing it.

stellarphot/core.py
```diff
@@ -70,31 +111,50 @@ class Camera:
...
+            rn_unit = Quantity(values['read_noise']).unit
+        except KeyError:
+            raise ValueError("read_noise must be specified including units.")
+
+        # Check that gain and read noise have compatible units, that is that
+        # gain is read noise per adu.
+        gain = Quantity(values['gain'])
```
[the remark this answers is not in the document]
JuanCab: Again, left over from writing the code, good catch.

stellarphot/tests/test_core.py
JuanCab: Well, we are throwing `ValueError` in the Camera class if there are bad values, should I have been throwing `ValidationError`?

stellarphot/tests/test_core.py
JuanCab: Noted

stellarphot/tests/test_core.py
JuanCab: Noted.

stellarphot/core.py
JuanCab: [The documentation I looked at](https://docs.pydantic.dev/1.10/usage/validators/) never shows `@classmethod` added in addition to `@root_validator`, can you point mt to your documentation reference?

stellarphot/core.py
```diff
@@ -70,31 +111,50 @@ class Camera:
...
+        extra = "forbid"
+        json_encoders = {
+            Quantity: lambda v: f"{v.value} {v.unit}",
+            QuantityType: lambda v: f"{v.value} {v.unit}",
+            PixelScaleType: lambda v: f"{v.value} {v.unit}"
+        }
+
+    # When the switch to pydantic v2 happens, this root_validator will need
+    # to be replaced by a model_validator decorator.
+    @root_validator()
```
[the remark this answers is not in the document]
JuanCab: Addressed.

stellarphot/tests/test_core.py
JuanCab: OK, I changed the tests, and they still clear.

stellarphot/core.py
JuanCab: OK, you win, changes made.

mwcraig (approved):
Thanks!

PR #164, opened 2023-09-03
Fix calculation of FWHM when the input data has non-finite values

### LINKED ISSUE #161: compute_fwhm fails if there are NaN values in the input
It turns out that `compute_fwhm` fails if there are `NaN`s in the input data. Though there is an option to filter out infinite values (i.e. `np.inf`) but no way in astropy modeling to filter out `np.nan` though the documentation [explains how to do that in a note](https://docs.astropy.org/en/stable/modeling/fitting.html).

mwcraig:
This fixes #161 by properly handling non-finite values in setting up the fitter when the FWHM calculation is done by fitting a 2D gausssian.

Though [this change in #145](https://github.com/feder-observatory/stellarphot/pull/145/files#diff-5f569c8d392db0dcd1a50489fc65d6e74bcad4169101017c22020db1761dba43R55) did indeed add an argument to the call to the fitter to mask out non-finite values, the model to which it was being fit had NaNs in it if the input data had NaNs.

This has been fixed by filtering out non-finite values when generating the parameters for the model to which the fit is done.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,7 @@ Other Changes and Additions
 Bug Fixes
 ^^^^^^^^^
 + Fixed dependence on non-release version of astrowidgets for overwrite capability on output images. [#108]
++ Fixed computation of FWHM when fitting to data that includes NaNs. [#164]
 
 1.3.9 (2023-06-16)
 ------------------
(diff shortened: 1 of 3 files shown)


JuanCab (commented):
Looks good.

PR #172, opened 2023-09-29
Fix formatting of columns in AAVSO tables

mwcraig:
The AAVSO Extended File Format has limits on the number of characters in some columns that effectively require rounding floating point values. This PR tries to ensure that the CSV file generated respects those limits.


mwcraig:
Learning to love self-merging today, I guess...YOLO!

PR #175, opened 2023-10-15
Update photutils and fix a couple broken tests

mwcraig:
This is a pretty straightforward one. The original motivation was to simply update the minimum photutils so taht we can, down the road, pull in some features from it, but I noticed a couple floating point tests were failing locally so I fixed those too.

### DIFF
--- a/stellarphot/tests/test_core.py
+++ b/stellarphot/tests/test_core.py
@@ -245,9 +245,9 @@ def test_photometry_data():
     assert phot_data.camera.read_noise == 10.0 * u.electron
     assert phot_data.camera.dark_current == 0.01 * u.electron / u.second
     assert phot_data.camera.pixel_scale == 0.563 * u.arcsec / u.pix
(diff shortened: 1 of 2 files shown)


stellarphot/tests/test_core.py
```diff
@@ -245,9 +245,9 @@ def test_photometry_data():
     assert phot_data.camera.read_noise == 10.0 * u.electron
     assert phot_data.camera.dark_current == 0.01 * u.electron / u.second
     assert phot_data.camera.pixel_scale == 0.563 * u.arcsec / u.pix
-    assert phot_data.observatory.lat.value == 46.86678
+    np.testing.assert_almost_equal(phot_data.observatory.lat.value, 46.86678)
```
JuanCab, line 248: Cool, nice to know this function exists for checking floats.  Looked over the document, seems to be used fine here (I was concerned about precision, but default precision is 7 decimal places, which is clearly enough here).

JuanCab (approved):
Looks good.

Issue #128, opened 2023-07-10
Is `plot_many_factors` actually used anywhere, seems incomplete?


JuanCab:
The `plot_many_factors` function in `visualization/transit_plots.py` doesn't seem to actually do anything useful (it makes some calculations, but the data is NOT kept before it exits.  Furthermore, the code doesn't seem to be used anywhere.  Purge it?

mwcraig:
I think this was an attempt to move a bunch of code out of photometry notebook 06 and into an importable file. Looks like that was never completed, though...

PR #178, opened 2023-11-17
Update notebooks 

mwcraig:
These changes are not everything that needs to be done, but these are more steps along the way that get us through the exoplanet transit notebook.

### DIFF
--- a/stellarphot/differential_photometry/aij_rel_fluxes.py
+++ b/stellarphot/differential_photometry/aij_rel_fluxes.py
@@ -16,7 +16,7 @@ def add_in_quadrature(array):
 def calc_aij_relative_flux(star_data, comp_stars,
                            in_place=True, coord_column=None,
                            star_id_column='star_id',
-                           counts_column_name='aperture_sum'):
+                           counts_column_name='aperture_net_cnts'):
     """
     Calculate AstroImageJ-style flux ratios.
 
--- a/stellarphot/differential_photometry/aij_rel_fluxes.py
+++ b/stellarphot/differential_photometry/aij_rel_fluxes.py
@@ -6,6 +6,7 @@ import astropy.units as u
 
 __all__ = ['add_in_quadrature', 'calc_aij_relative_flux']
 
+
 def add_in_quadrature(array):
     """
     Add an array of numbers in quadrature.
@@ -23,8 +24,8 @@ def calc_aij_relative_flux(star_data, comp_stars,
     Parameters
     ----------
 
-    star_data : '~astropy.table.Table'
-        Table of star data from one or more images.
+    star_data : 'stellarphot.PhotometryDat'
+        Photometry data from one or more images.
 
     comp_stars : '~astropy.table.Table'
         Table of comparison stars in the field. Must contain a column
@@ -56,7 +57,7 @@ def calc_aij_relative_flux(star_data, comp_stars,
     Returns
     -------
 
-    `astropy.table.Table` or None
+    `stellarphot.PhotometryData` or None
         The return type depends on the value of ``in_place``. If it is
         ``False``, then the new columns are returned as a separate table,
         otherwise the columns are simply added to the input table.
--- a/stellarphot/differential_photometry/aij_rel_fluxes.py
+++ b/stellarphot/differential_photometry/aij_rel_fluxes.py
@@ -16,7 +16,7 @@ def add_in_quadrature(array):
 def calc_aij_relative_flux(star_data, comp_stars,
                            in_place=True, coord_column=None,
                            star_id_column='star_id',
-                           counts_column_name='aperture_net_counts'):
+                           counts_column_name='aperture_sum'):
     """
     Calculate AstroImageJ-style flux ratios.
 
@@ -28,7 +28,7 @@ def calc_aij_relative_flux(star_data, comp_stars,
 
     comp_stars : '~astropy.table.Table'
         Table of comparison stars in the field. Must contain a column
-        called ``RA`` and a column called ``Dec``.
+        called ``ra`` and a column called ``dec``.
         NOTE that not all
         of the comparison stars will necessarily be used. Stars in
         this table are excluded from the comparison set if, in any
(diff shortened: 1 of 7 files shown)


stellarphot/differential_photometry/aij_rel_fluxes.py
```diff
@@ -16,19 +17,19 @@ def add_in_quadrature(array):
...
+                           counts_column_name='aperture_net_cnts'):
     """
     Calculate AstroImageJ-style flux ratios.
 
     Parameters
     ----------
 
-    star_data : '~astropy.table.Table'
-        Table of star data from one or more images.
+    star_data : 'stellarphot.PhotometryDat'
```
JuanCab, line 27: Shouldn't this be 'stellarphot.PhotometryData' (with an 'a' at the end)?

stellarphot/differential_photometry/aij_rel_fluxes.py
```diff
@@ -47,7 +48,7 @@ def calc_aij_relative_flux(star_data, comp_stars,
 
     counts_column_name : str,  optional
         If provided, use this column to find counts.
-        If not provided, the column 'aperture_net_counts' is used.
+        If not provided, the column 'aperture_sum' is used.
```
JuanCab, line 51: This appears to be inconsistent with line 20 above which has the default as actually 'aperture_net_cnts'.

stellarphot/differential_photometry/aij_rel_fluxes.py
```diff
@@ -47,7 +48,7 @@ def calc_aij_relative_flux(star_data, comp_stars,
```
JuanCab, line 48: Missing documentation of following parameters in function definition 

- `in_place=True`
- `coord_column=None`

stellarphot/differential_photometry/tests/test_aij_rel_fluxes.py
JuanCab, line 1: Looks good, just renaming column names and handling import of `PhotometryData`

stellarphot/notebooks/photometry/06-transit-fit-template.ipynb
JuanCab, line 1: In the cell where you have

`side_transit = (photometry['bjd'].jd < start.jd) | (photometry['bjd'].jd > end.jd)`

should this be

`side_transit = (photometry['bjd'] < start.jd) | (photometry['bjd'] > end.jd)`?

After all, the photometry['bjd'] is a column in the table.

JuanCab (commented):
I caught a few missing bits of documentation in the aij_rel_fluxes code and the notebook has (what I believe is) a bad call to check if the observation is in-transit.

JuanCab (changes requested):
I caught a few missing bits of documentation in the aij_rel_fluxes code and the notebook has (what I believe is) a bad call to check if the observation is in-transit.

JuanCab (changes requested):
I caught a few missing bits of documentation in the aij_rel_fluxes code and the notebook has (what I believe is) a bad call to check if the observation is in-transit.

stellarphot/differential_photometry/aij_rel_fluxes.py
mwcraig: I'm confused -- do you mean the default value isn't listed?  I'd prefer to just leave that in the function signature so that it doesn't have to be updated in two places if it changes in the future.

stellarphot/notebooks/photometry/06-transit-fit-template.ipynb
mwcraig: It could probably change to `(photometry['bjd'] < start) | (photometry['bjd'] > end)` (i.e. no `.jd` in either place) but if a `.jd` is present in one side of the comparison it needs to be in the other. Dropping the jd in both cases would make sense though, will give that a try.

stellarphot/differential_photometry/aij_rel_fluxes.py
JuanCab: I didn't see the parameters listed at all in the docs, or am I mistaken?

stellarphot/notebooks/photometry/06-transit-fit-template.ipynb
JuanCab: Not sure how .jd makes sense in that for the photometry table it is a column, not an object with "jd" an attribute, or am I missing something.

mwcraig: That column is a `Time`, so it has a `.jd`. The `.jd` isn't needed in the comparisons, though, so I took it out.

stellarphot/differential_photometry/aij_rel_fluxes.py
mwcraig: They are tehre but were maybe originally not shown because they didn't change?

mwcraig:
I think I've addressed all of the comment, and fixed #128 along the way 😀

JuanCab:
> I think I've addressed all of the comment, and fixed #128 along the way 😀

Well played.  Reviewing Now...

stellarphot/differential_photometry/aij_rel_fluxes.py
JuanCab: I really hate this reviewing interface sometimes. 

I do think the default values should be listed because otherwise a person may think the interface behave differently without the parameter being passed.  Also, I believe I listed defaults for all the docstrings I created/added, which frankly was a lot of them in stellarphot. :)

stellarphot/notebooks/photometry/04-transform-pared-back.ipynb
JuanCab, line 1: Just curious why you have `aavso_band_names` has keys `ip` and `SI` that correspond to same value?  Are you planning for the possibility people will use `ip` or `SI` and not be consistent?  Its just that you don't do it with the other Sloan filters as near as I can tell...

JuanCab (approved):
I think it is good to go.  I asked a question, but not one that holds things up.

JuanCab (approved):
I think it is good to go.  I asked a question, but not one that holds things up.

stellarphot/notebooks/photometry/04-transform-pared-back.ipynb
mwcraig: That is a fine question...ah, I remember. It is because later code in the notebook is knda stupid. It uses `our_filters` to decide what to transform and expects each of those to be a key in `aavso_band_names`. Since the photometry already has set the filter names to the AAVSO ones it ends up being a little silly.

The fix is a partial rewrite of the notebook, but I'll wait until I've done the catalog update for that.

PR #180, opened 2023-11-24
Move catalog functionality to the new data classes

mwcraig:
This has ended up being a little broader than I initially thought it might be because I decided it was useful to put the (currently unused but soon to get used) `catalog_clean` into `BaseEnhancedTable`.

This PR will also update/change `find_apass_stars` and `find_known_variables` to use the `CatalogData` class. That class will acquire the `catalog_search` method too.

To do before marking as not-a-draft:

- [x] Move tests of `in_frame` functionality to `test_core`
- [x] Implement `find_apass_stars` functionality in `CatalogData`
- [x] Remove `find_apass_stars`
- [x] Update calls of `find_apass_stars`
- [x] Implement `find_known_variables` functionality in `CatalogData`
- [x] Update calls of `find_known_variables`
- [x] Remove `in_frame`
- [x] Update calls to `in_frame`
- [x] Remove `catalog_search`
- [x] Update calls to `catalog_search`


mwcraig:
Note to self:

Not sure how to handle the restructuring piece of this. The lookup is easy but translating the output into a single `mag` column with multiple passbands rather than multiple passbands, one per column, is not straightforward and will be often be different for different catalogs.

The initial implementation allows for a callback to be passed in that rearranges the Vizier search result into the proper format.

Not a terrible idea, but and ot simple either....

stellarphot/core.py
```diff
@@ -305,6 +310,86 @@ def _update_passbands(self):
...
+            criteria_re = re.compile(r'({})([-+a-zA-Z0-9]+)'.format(recognized_comparison_ops))
+            results = criteria_re.match(restriction)
+            if not results:
+                raise ValueError("Criteria {}{} not "
+                                "understood.".format(column, restriction))
+            comparison_func = comparisons[results.group(1)]
+            comparison_value = results.group(2)
+            new_keepers = comparison_func(self[column],
+                                        float(comparison_value))
+            keepers = keepers & new_keepers
```
JuanCab, line 389: I worked through the rest of this logic, clever bit of of work to turn the entered operators into comparison_func.  The logic seems sound.  And since you have a docstring test in there, looks positive.  I might suggest testing a few more of the comparisons if I was feeling nasty, although really that that should show is no typos in the names of the functions in comparisons dict.

stellarphot/core.py
```diff
@@ -715,6 +800,225 @@ def catalog_name(self):
...
+        # Find all the magnitude and color columns
+        mag_match = [mag_re.match(col) for col in data.colnames]
+        color_match = [color_re.match(col) for col in data.colnames]
+
+        # create a single list of all the matches
+        matches = [m_match if m_match else c_match
+                   for m_match, c_match in zip(mag_match, color_match)]
+
+        # The passband should be the first group match.
+        passbands = [match[1] for match in matches if match]
```
JuanCab, line 842: Are you grabbing the second character from the 'match' column name?  I *thought* it was usually the first character in the column name...  Also, if you include any color column names at all in matches, is passband a valid idea?

stellarphot/core.py
```diff
@@ -715,6 +800,225 @@ def catalog_name(self):
...
+        mag_re = re.compile(mag_column_regex)
+        color_re = re.compile(color_column_regex)
+
+        # Find all the magnitude and color columns
+        mag_match = [mag_re.match(col) for col in data.colnames]
+        color_match = [color_re.match(col) for col in data.colnames]
+
+        # create a single list of all the matches
+        matches = [m_match if m_match else c_match
+                   for m_match, c_match in zip(mag_match, color_match)]
```
JuanCab, line 839: To you identify the magnitude and color columns and then build a list of matches that consists of the magnitude column name (when available) otherwise a color column name.  Why?  These are column names, if there are 3 magnitudes and 4 color columns, why would you only keep the last color column (ignoring the first three color columns since there are three magnitude columns already).

stellarphot/core.py
```diff
@@ -715,6 +800,225 @@ def catalog_name(self):
...
+        # desired.
+        if clip_by_frame:
+            cat_coords = SkyCoord(ra=cat['ra'], dec=cat['dec'])
+            x, y = header_or_center.wcs.all_world2pix(cat_coords.ra, cat_coords.dec, 0)
+            in_x = (x >= padding) & (x <= header_or_center.wcs.pixel_shape[0] - padding)
+            in_y = (y >= padding) & (y <= header_or_center.wcs.pixel_shape[1] - padding)
+            in_fov = in_x & in_y
+            cat = cat[in_fov]
+
+        return cat
```
JuanCab, line 1020: Logic looks OK as far as I can tell for this from_vizier method.  You have a test for it in test_core.py, so I'll trust it for now as long as the tests passes.

JuanCab (commented):
Logic to move the functions that were in `catalog_search.py` into the `CatalogData` class.  You planning to move the other functions?  `in_frame` makes sense to move to `CatalogData`, although if you have `clip_by_frame` in the `from_vizier` method, you would only need it for NON-Vizier sourced data (like APASS?).  Can you move `in_frame` to `CatalogData` and make the function general enough to use for all `CatalogData` constructs?

stellarphot/core.py
```diff
@@ -305,6 +310,86 @@ def _update_passbands(self):
...
+
+        recognized_comparison_ops = '|'.join(comparisons.keys())
+        keepers = np.ones([len(self)], dtype=bool)
+
+        if remove_rows_with_mask and self.has_masked_values:
+            for c in self.columns:
+                keepers &= ~self[c].mask
+
+        for column, restriction in other_restrictions.items():
+            criteria_re = re.compile(r'({})([-+a-zA-Z0-9]+)'.format(recognized_comparison_ops))
```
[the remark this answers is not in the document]
mwcraig: Yes to the first question, I'm not sure about the second. Either way, I think the compile should be outside the loop since nothing in the compile regex depends on the loop variables.

stellarphot/core.py
mwcraig: I'll add some more tests to the test suite to check some of the operators (and try things like comparing a string). I put the example in in part because the docstring said there was an example but there wasn't 😬.

stellarphot/core.py
mwcraig: The first match is the entire string (or at least the entire matched part of the string) and the next one is the first matching group, with the groups set up by the parentheses in the regex.

stellarphot/core.py
mwcraig: My intent is to capture all of the mag and color columns -- I'll write a test case that checks that that really happens.

Each individual column is either a color column (e.g. name `g-r`) or a magnitude (e.g. `g_mag` or `r_mag`), so for a set of columns named `['RA', 'DEC', 'g_mag', 'r_mag', 'g-r'] the two match lists should be

```python
mag_match =  [None, None, <Match object>, <Match object>,                   None]
color_match = [None, None,                   None,                   None, <Match object>]
```

so the list comprehension is grabbing whichever match is not `None`, if any.

setup.cfg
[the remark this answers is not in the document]
mwcraig: That too, but some change in the astropy configuration system also broke.

mwcraig:
@JuanCab -- I think this is ready for review

stellarphot/core.py
```diff
@@ -305,6 +313,87 @@ def _update_passbands(self):
...
+            '<': np.less,
+            '=': np.equal,
+            '>': np.greater,
+            '<=': np.less_equal,
+            '>=': np.greater_equal,
+            '!=': np.not_equal
+        }
+
+        recognized_comparison_ops = '|'.join(comparisons.keys())
+        criteria_re = re.compile(r'({})([-+a-zA-Z0-9]+)'.format(recognized_comparison_ops))
```
JuanCab, line 376: Might you modify this to allow spaces (e.g. `> 2`)?  Also, I think this criteria will not allow float operands.  While I am not sure supporting spaces is necessary, I would think floats should be allowable.

 I did notice you moved the compile out of the loop.

stellarphot/core.py
JuanCab: Got it.

stellarphot/core.py
JuanCab: Oh that's right, `re` is interesting that way...

stellarphot/core.py
```diff
@@ -715,6 +804,317 @@ def catalog_name(self):
...
+        'recno': 'id', # There is no APASS ID, this is the one generated by Vizier
+        'RAJ2000': 'ra',
+        'DEJ2000': 'dec',
+    }
+    return CatalogData.from_vizier(header_or_center,
+                                   'II/336/apass9',
+                                   radius=radius,
+                                   clip_by_frame=clip_by_frame,
+                                   padding=padding,
+                                   colname_map=apass_colnames)
```
JuanCab, line 1066: Given this construction, I can see why you added this function to core, but it really should probably go a separate file, I think... at least based on it not really being core functionality.  Same for `vsx_vizier`

stellarphot/core.py
```diff
@@ -715,6 +804,317 @@ def catalog_name(self):
...
+
+
+def vsx_vizier(header_or_center,
+               radius=1 * u.degree,
+               clip_by_frame=False,
+               padding=100):
+    """
+    Return the items from APASS DR9 that are within the search radius and
```
JuanCab, line 1074: I suspect you mean VSX and not APASS DR9.

stellarphot/notebooks/photometry/get_apass_comp_mags.ipynb
JuanCab, line 1: Some issues with this notebook:

- Typo in cell 4, you state "L:ook "instead of "Look"
- Pandas throws a deprecation warning with cell 7 `panda = pd.read_json(json, orient = 'records')` saying `FutureWarning: Passing literal json to 'read_json' is deprecated and will be removed in a future version. To read from a literal string, wrap it in a 'StringIO' object.`. You should probably go ahead and address this.
- In Cell 11 the program crashes unless you change `apass['coords'] = SkyCoord(ra=apass['RAJ2000'], dec=apass['DEJ2000'], unit=('degree', 'degree'))` to `apass['coords'] = SkyCoord(ra=apass['ra'], dec=apass['dec'], unit=('degree', 'degree'))`
- In Cell 17 the command `del apass_comp_stars['coords'], apass_comp_stars['B-V'], apass_comp_stars['e_B-V']` crashes because `column "B-V" does not exist`.  You are going to need to change the logic to filter out by the `passband` column value.  Same problem in cell 19 with `KeyyError: 'Vmag'`

stellarphot/tests/test_core.py
```diff
@@ -3,12 +3,15 @@
...
+from astropy.io import ascii, fits
+from astropy.coordinates import EarthLocation, SkyCoord
+from astropy.nddata import CCDData
 from astropy.utils.data import get_pkg_data_filename
+from astropy.wcs import WCS
 from pydantic import ValidationError
+
 from stellarphot.core import (Camera, BaseEnhancedTable, PhotometryData,
-                              CatalogData, SourceListData)
+                              CatalogData, SourceListData, apass_dr9, vsx_vizier)
```
JuanCab, line 14: OK given code changes made BUT not sure these two functions should be in core.

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+def test_base_enhanced_table_clean():
+    # Check that the clean method exists
+    test_base = BaseEnhancedTable(table_description=test_descript, input_data=testdata)
+    # Add a row so that we can clean something
+    test_base_two = test_base.copy()
+    test_base_two.add_row(test_base[0])
+    test_base_two['ra'][1] = - test_base_two['ra'][1]
+    test_cleaned = test_base_two.clean(ra='>0.0')
+    assert len(test_cleaned) == 1
+    assert test_cleaned == test_base
```
JuanCab, line 130: Looks like a reasonable test.

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+
+def test_bet_clean_criteria_none_removed():
+    """
+    If all rows satisfy the criteria, none should be removed.
+    """
+    inp = a_table()
+    criteria = {'a': '>0'}
+    out = inp.clean(**criteria)
+    assert len(out) == len(inp)
+    assert (out == inp).all()
```
JuanCab, line 154: Looks like a reasonable test, again.

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+@pytest.mark.parametrize("condition,input_row",
+                            [('>0', 0),
+                             ('=1', 0),
+                             ('>=1', 0),
+                             ('<-1', 2),
+                             ('=-1', 1)])
+def test_bet_clean_criteria_some_removed(condition, input_row):
+    """
+    Try a few filters which remove the second row and check that it is
+    removed.
```
JuanCab, line 166: Some of these should remove more than just the second row, right?  I think the comment needs an update.

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+    """
+    inp = a_table()
+    criteria = {'b': condition}
+    out = inp.clean(**criteria)
+    assert len(out) == 1
+    assert (out[0] == inp[input_row]).all()
```
JuanCab, line 172: Assuming it works, looks good.

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+                         ({'a': '5'}, "not understood"),
+                         ({'a': '<foo'}, "could not convert string")])
+def test_clean_bad_criteria(criteria, error_msg):
+    """
+    Make sure the appropriate error is raised when bad criteria are used.
+    """
+    inp = a_table(masked=False)
+
+    with pytest.raises(ValueError, match=error_msg):
+        inp.clean(**criteria)
```
JuanCab, line 185: Looks like a reasonable test, again.

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+    inp = a_table(masked=True)
+    # Mask negative values
+    inp['b'].mask = inp['b'] < 0
+    out = inp.clean(remove_rows_with_mask=clean_masked)
+    if clean_masked:
+        assert len(out) == 1
+        assert (np.array(out[0]) == np.array(inp[0])).all()
+    else:
+        assert len(out) == len(inp)
+        assert (out == inp).all()
```
JuanCab, line 200: Looks like a reasonable test, again.

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+    out = inp.clean(remove_rows_with_mask=True, **criteria)
+
+    # Is only one row left?
+    assert len(out) == 1
+
+    # Is the row that is left the same as the second row of the input?
+    assert (np.array(out[0]) == np.array(inp[1])).all()
+
+    # Is the input table unchanged?
+    assert (inp == inp_copy).all()
```
JuanCab, line 225: Looks like a reasonable test, otra vez...

stellarphot/tests/test_core.py
```diff
@@ -115,6 +118,124 @@ def test_base_enhanced_table_from_existing_table():
...
+
+def test_clean_criteria_none_removed():
+    """
+    If all rows satisfy the criteria, none should be removed.
+    """
+    inp = a_table()
+    criteria = {'a': '>0'}
+    out = inp.clean(**criteria)
+    assert len(out) == len(inp)
+    assert (out == inp).all()
```
JuanCab, line 236: Looks like a reasonable test, one more time.

stellarphot/tests/test_core.py
```diff
@@ -379,6 +500,165 @@ def test_catalog_recursive():
...
+
+    # Reference data was sorted by RA, first 20 entries kept
+    # There are 6 magnitude or color columns, so 6 * 20 = 120 rows
+    # in the resulting table.
+    all_apass.sort('ra')
+    all_apass = all_apass[:120]
+
+    # It is hard to imagine the RAs matching and other entries not matching,
+    # so just check the RAs.
+    assert set(ra.value for ra in all_apass['ra']) == set(expected_all['RAJ2000'])
```
JuanCab, line 659: All these catalog tests seem to be reasonable as far as I can tell.

stellarphot/utils/comparison_utils.py
JuanCab, line 1: Looks OK based on what vex_vizier and apass_dr9 are supposed to do.

JuanCab (commented):
OK, you have my feedback.  Biggest annoyance is I caught a few bugs in the notebook that changed and I am not sure about placing vex_vizier and apass_dr9 in the core.py file.

setup.cfg
```diff
@@ -52,8 +52,9 @@ stellarphot.io = *.yml
 stellarphot.tests =
     coveragerc
     data/*
-stellarphot.transit_fitting.gui = data/*
+stellarphot.transit_fitting = data/*
```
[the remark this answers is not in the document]
mwcraig: The `transit_fitting` subpackage is different in that is has data separate from the tests. Eventually that can change -- the data is a couple of json files that could be replaced with pydantic classes eventually

stellarphot/core.py
```diff
@@ -1,14 +1,22 @@
...
+from astroquery.vizier import Vizier
 
+import pandas as pd
 from pydantic import BaseModel, root_validator
 
 import numpy as np
 
 __all__ = ['Camera', 'BaseEnhancedTable', 'PhotometryData', 'CatalogData',
+           'apass_dr9',
+           'vsx_vizier',
```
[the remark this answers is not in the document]
mwcraig: I put it in `core` since they are basically class factories for `CatalogData`.

I'm currently thinking `core` should be split into separate files. Perhaps `core` retains `BaseEnhancedTable` and `PhotometryTable` and the other two classes split off into separate files. Haven't really thought it through.

It would definitely be nice to group the tests better either by splitting them into separate files or using classes in the single file. It would be great if you could see at a glance which class has a problem.

I haven't thought any of this through really, will try to come up with a concrete proposal.

stellarphot/notebooks/photometry/get_apass_comp_mags.ipynb
mwcraig: It might make sense just to drop this one totally....

mwcraig:
I've opened #194 to continue the discussion of what should be in `core.py`

mwcraig:
No idea why [this](https://github.com/feder-observatory/stellarphot/actions/runs/7129306911/job/19413103462?pr=180#step:7:234) is happening 🙄

mwcraig:
> No idea why [this](https://github.com/feder-observatory/stellarphot/actions/runs/7129306911/job/19413103462?pr=180#step:7:234) is happening 🙄

Happily, it is no longer happening 😀

stellarphot/utils/comparison_utils.py
```diff
@@ -72,7 +72,7 @@ def set_up(sample_image_for_finding_stars, directory_with_images="."):
 
     ccd = CCDData.read(path)
     try:
-        vsx = find_known_variables(ccd)
+        vsx = vsx_vizier(ccd.header, radius=0.5 * u.degree)
```
JuanCab, line 75: Do you want to deal with the code coverage warnings?

JuanCab (approved):
Other than the possible issue with how to handle the new code coverage warnings (e.g. adding more tests), looks good.

JuanCab (approved):
Other than the possible issue with how to handle the new code coverage warnings (e.g. adding more tests), looks good.

PR #197, opened 2023-12-09
Adopt pyproject.toml and hatch

### LINKED ISSUE #159: Switch to latest python packaging scheme
This may as well be done prior to releasing 2.0. I think using `hatch` is the way to go.

mwcraig:
This closes #159. Still to do:

- [x] Move pytest configuration to `pyproject.toml`


ofek (approved):
LGTM! Two things I would note:

1. You can now completely get rid of `setup.py`
2. Totally up to your style preference but just FYI all of those indentations in TOML are not required

mwcraig:
Thanks for taking a look!

pyproject.toml
```diff
@@ -1,5 +1,138 @@
...
-build-backend = 'setuptools.build_meta'
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "stellarphot"
+dynamic = ["version"]
+description = "A package for performing stellar photometry."
+readme = "README.md"
+license = { file = "LICENSE.rst" }
```
ofek, line 10: Oh, one other cool thing if you want is that with Hatchling this can be defined as a string because we were very early adopters of the future proposal that was always planned https://hatch.pypa.io/latest/config/metadata/#license

Don't worry about it not being accepted, the efforts to apply the final feedback for acceptance has recommenced: https://discuss.python.org/t/pep-639-round-2-improving-license-clarity-with-better-package-metadata/12622/96

The person was busy for some time working on space stuff 🙂 https://discuss.python.org/t/pep-639-round-2-improving-license-clarity-with-better-package-metadata/12622/81

```suggestion
license = "BSD-3-Clause"
```

PR #205, opened 2023-12-12
Change documentation build to use pyproject.toml and fix minor sphinx issues

mwcraig:
This should have been done as part #197 but we don't have a documentation build as part of the CI, so it wasn't.

Also adding a documentation build to the CI matrix...


mwcraig:
Self-merging is going to be a theme this week I think. Another infrastructure update here with minimal changes to actual code.

PR #214, opened 2023-12-15
Blackify code

mwcraig:
This PR

- Formats the code with black
- Adds black configuration
- Adds a pre-commit hook for black
- Ignores the commit in which the code was reformatted for the purposes of git blame


mwcraig:
I don't think the coverage change is meaningful -- we have increased the number of lines of code a bit by blackifying it, which reduces the coverage when measured by number of lines covered. black promises not to change the functioning of the code, so it seems safe to assume this hasn't added new problems.

mwcraig:
@JuanCab -- this may now automatically apply black to new commits once you have pulled the latest version of the repo. Not 100% sure yet.

PR #231, opened 2023-12-22
Update Camera object

### LINKED ISSUE #165: Modify `Camera` model so that it can generate a JSON schema so that we can generate a UI
Right now this code

```python
from stellarphot import Camera
c = Camera(gain="1 electron/adu", 
           read_noise="10.1 electron", 
           dark_current="0.1 electron/second", 
           pixel_scale="0.556 arcsec/pixel")
c.schema()
```

leads to 

<details>
<summary>this error:</summary>

```python-traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[28], line 1
----> 1 c.schema()

File ~/mambaforge/envs/stelldev2/lib/python3.11/site-packages/pydantic/main.py:664, in pydantic.main.BaseModel.schema()

File ~/mambaforge/envs/stelldev2/lib/python3.11/site-packages/pydantic/schema.py:188, in pydantic.schema.model_schema()

File ~/mambaforge/envs/stelldev2/lib/python3.11/site-packages/pydantic/schema.py:582, in pydantic.schema.model_process_schema()

File ~/mambaforge/envs/stelldev2/lib/python3.11/site-packages/pydantic/schema.py:623, in pydantic.schema.model_type_schema()

File ~/mambaforge/envs/stelldev2/lib/python3.11/site-packages/pydantic/schema.py:256, in pydantic.schema.field_schema()

File ~/mambaforge/envs/stelldev2/lib/python3.11/site-packages/pydantic/schema.py:528, in pydantic.schema.field_type_schema()

File ~/mambaforge/envs/stelldev2/lib/python3.11/site-packages/pydantic/schema.py:952, in pydantic.schema.field_singleton_schema()

ValueError: Value not declarable with JSON Schema, field: name='gain' type=QuantityType required=True
```
</details>

In the short run this doesn't matter --we don't need a schema to read and write the settings to a JSON file. 

If we want to add a UI for setting `Camera` we will need to be able to generate a schema because `ipyautoui` uses it.

Some relevant links: 

+ The [pydantic docs](https://docs.pydantic.dev/1.10/usage/schema/#modifying-schema-in-custom-fields) talk about customizing the schema.
+ This [stack overflow question](https://stackoverflow.com/questions/68893175/error-value-not-declarable-with-json-schema-for-purepath-with-pydantic-and-fasta) also addresses it and has link to a GitHub issue with more discussion.
+ I gather the solution is to just declare it as a string, which works with `Quantity` since they can be initialized with a string.

mwcraig:
This ended up doing too many things -- please let me know if you want me to break it into smaller PRs.

1. Adds "max_adu" as a `Camera` attribute. I named it `max_data_val` instead, because...
2. ...allow any data unit, but require user to say what it is by adding a `data_unit` attribute to `Camera`.
3. Fix #165 by indicating how schema should be created for the types we define.
4. Adds a bit of description to each field and some examples, which end up being displayed in the auto-generated UI.
5. Drops the `copy` method because pydantic provides one
6. Drops the `__repr__` because pydantic provides one

(and did 5 and 6 to avoid having to update them every time a change is made)


stellarphot/core.py
```diff
@@ -30,6 +30,36 @@
 # https://docs.gammapy.org/dev/_modules/gammapy/analysis/config.html
 
 
+class UnitType(Unit):
```
JuanCab, line 33: So I assume this is adding pydantic validators to the astropy units class?  Do you know if it is pydantic v2 compatible? There is a part of me that worries that pydantic v2 is still waiting to bite us in the future.

stellarphot/core.py
```diff
@@ -30,6 +30,36 @@
...
+    # Validator for Unit type
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
```
JuanCab, line 40: What is this validate doing?  It just returns the Unit cast version of v?  Does it actually validate anything?

stellarphot/core.py
```diff
@@ -30,6 +30,36 @@
...
+        return Unit(v)
+
+    @classmethod
+    def __modify_schema__(cls, field_schema, field):
+        # Set default values for the schema in case the field doesn't provide them
```
JuanCab, line 45: I am honestly not sure how to assess what is going on here with `__modify_schema__`, need to dig into pydantic again or trust your tests. :). I will note `__modify_schema__` doesn't appear to be supported in v2 of pydantic (based on https://docs.pydantic.dev/latest/errors/usage_errors/#custom-json-schema).

stellarphot/core.py
```diff
@@ -30,6 +30,36 @@
...
+        name = "Unit"
+        description = "An astropy unit"
+
+        name = field.name or name
```
JuanCab, line 49: Is this setting name to a true/false value based on `field.name or name` as a boolean?

stellarphot/core.py
```diff
@@ -128,13 +206,32 @@ class Camera(BaseModel):
...
+    gain: QuantityType = Field(
+        description="unit should be consistent with data and read noise",
+        examples=["1.0 electron / adu"],
+    )
+    read_noise: QuantityType = Field(
+        description="unit should be consistent with dark current",
+        examples=["10.0 electron"],
+    )
+    dark_current: QuantityType = Field(
+        description="unit consisten with read noise, per unit time",
```
JuanCab, line 225: "unit *consistent*" (typo)

stellarphot/photometry/photometry.py
JuanCab, line 1: Again, all changes consistent with moving max_adu to Camera.

stellarphot/notebooks/photometry/03-photometry-template.ipynb
JuanCab, line 1: All changes consistent with moving max_adu to Camera.

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -22,10 +22,25 @@
 )
 from stellarphot.settings import ApertureSettings
 
+# Constants for the tests
+
 GAINS = [1.0, 1.5, 2.0]
 # Make sure the tests are deterministic by using a random seed
```
JuanCab, line 28: Really not 'using a random seed', rather you are forcing a 'fixed seed for random number generator'

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -22,10 +22,25 @@
...
 SEED = 5432985
 
+SHIFT_TOLERANCE = 6
```
JuanCab, line 31: Good call to move these toward the front of the testing code.

stellarphot/photometry/tests/test_photometry.py
JuanCab, line 1: Looks good, seems a pretty straight forward set of changes to using proper capitalized constants and to removing `max_adu` as an argument.

stellarphot/tests/test_core.py
```diff
@@ -25,74 +25,172 @@
...
     gain = 2.0 * u.electron / u.adu
     read_noise = 10 * u.electron
     dark_current = 0.01 * u.electron / u.second
     pixel_scale = 0.563 * u.arcsec / u.pix
-
-    with pytest.raises(ValidationError, match="gain"):
-        c = Camera(
+    max_adu = 50000 * u.adu
+
+    # All 5 of the attributes after data_unit will be checked for units
```
JuanCab, line 60: Cool, I didn't know you could do this with pytest, counting the number of errors.

stellarphot/tests/test_core.py
```diff
@@ -25,74 +25,172 @@
...
+    assert c.data_unit == data_unit
     assert c.gain == gain
     assert c.dark_current == dark_current
     assert c.read_noise == read_noise
     assert c.pixel_scale == pixel_scale
+    assert c.max_data_value == max_val
+
+
+def test_camera_schema():
+    # Check that we can generate a schema for a Camera and that it
```
JuanCab, line 182: I assume this schema method is a feature of pydantic...

JuanCab (approved):
As far as I can tell, it looks good.  I am not that experienced with pydantic, but otherwise, everything seemed reasonable.  I'll approve it.

stellarphot/core.py
mwcraig: If the creation of the unit fails (e.g. Unit("nonsense")) then it generates a `ValidationError`

stellarphot/core.py
mwcraig: To the extent I understand it, this is providing some of the information needed to generate a json schema for types that are either not built in to Python or pydantic. 

I wish ipyautoui had already made the 1 to 2 transition so we could just code for 2. It looks like there is a replacement for `__modify_schema__` in 2.

stellarphot/core.py
mwcraig: This sets `name` to `field.name` unless `field.name` evaluates to `False`, in which case `name` is used.

stellarphot/photometry/tests/test_photometry.py
mwcraig: That is true...

stellarphot/tests/test_core.py
mwcraig: It isn't counting the errors, `pydantic` is -- `pytest` is just checking that the text of the error message contains `"5 validation errors"`

stellarphot/tests/test_core.py
mwcraig: Yeah; it is used by ipyautoui when generating the UI.

PR #233, opened 2023-12-22
Make camera and observatory  `TableAttribute`s

### LINKED ISSUE #232: Maybe make `Camera` and `Observatory` table attributes
The `PhotometryTable` code has this comment:

```python
# Add the TableAttributes directly to meta (and adding attribute
# functions below) since using TableAttributes results in a
# inability to access the values to due a
# AttributeError: 'TableAttribute' object has no attribute 'name'
```

I was able to work out today how to make things like `Camera` round trip to `ECSV` -- should we make these attributes so that the objects don't have to be unpacked into `meta` the way they are [here](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/core.py#L624)?

mwcraig:
This doesn't have tests for round-tripping yet, but provides proof-of-concept at least. 

Fixes #232

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -2,7 +2,8 @@ import re
 
 from astropy import units as u
 from astropy.coordinates import EarthLocation, SkyCoord
-from astropy.table import Column, QTable, Table
+from astropy.io.misc.yaml import AstropyDumper, AstropyLoader
+from astropy.table import Column, QTable, Table, TableAttribute
 from astropy.time import Time
 from astropy.units import Quantity, Unit, IrreducibleUnit
 from astropy.wcs import WCS
@@ -293,6 +294,19 @@ class Camera(BaseModel):
         if v.value <= 0:
             raise ValueError("max_data_value must be positive")
         return v
+
+
+# Add YAML round-tripping for Camera
+def camera_representer(dumper, cam):
+    return dumper.represent_mapping("!Camera", cam.dict())
+
+
+def camera_constructor(loader, node):
+    return Camera(**loader.construct_mapping(node))
+
+
+AstropyDumper.add_representer(Camera, camera_representer)
+AstropyLoader.add_constructor("!Camera", camera_constructor)
 
 
 class BaseEnhancedTable(QTable):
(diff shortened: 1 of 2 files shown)


stellarphot/core.py
```diff
@@ -201,6 +202,19 @@ def __repr__(self):
         )
 
 
+# Add YAML round-tripping for Camera
```
JuanCab, line 205: So it looks like AstropyDumper handles conversion of objects to their YAML equivalents, interesting.

stellarphot/core.py
```diff
@@ -565,35 +579,21 @@ class PhotometryData(BaseEnhancedTable):
...
         input_data=None,
         observatory=None,
-        camera=None,
         colname_map=None,
         passband_map=None,
         retain_user_computed=False,
         **kwargs,
     ):
-        if (observatory is None) and (camera is None) and (input_data is None):
+        if (observatory is None) and (self.camera is None) and (input_data is None):
```
JuanCab, line 594: Wait, self.camera is a TableAttribute... but it defaults to None, so I assume this means it reports its value as None.  I assume you checked this.

stellarphot/core.py
```diff
@@ -565,35 +579,21 @@ class PhotometryData(BaseEnhancedTable):
...
-                    "astropy.coordinates.EarthLocation object instead "
-                    f"of type {type(observatory)}."
-                )
-            if not isinstance(camera, Camera):
-                raise TypeError(
-                    "camera must be a stellarphot.Camera object instead "
-                    f"of type {type(camera)}."
-                )
-
             # Check the time column is correct format and scale
```
JuanCab, line 597: Caught a typo in line 606 of code, comment should be
`# Happens if first item doesn't have a "scale"`
not
`# Happens if first item dosn't have a "scale"`

JuanCab (approved):
Interesting, will be more interesting once tests exist since I haven't tried this way of storing objects, but the logic seems reasonable.

stellarphot/core.py
mwcraig: I added a couple of lines to one of the tests to check this. As it turns out, the old implementation didn't work when `camera` or `observatory`  were `None` because accessing `phot_table.camera` returned the `property`, which tries to construct a `Camera` even if the there wasn't one when the object was created.

stellarphot/core.py
JuanCab: However, I am just trying to understand the logic here (and inline documentation on this might help us recall this in the future if we choose to add more to these custom AstroPy tables). 

`camera_representer` and `camera_constructor` are currently implemented as functions created when we import core.  From the astropy documentation it seems "[AstropyDumper] is not directly instantiated by user code, but instead is used to maintain the available [representer functions that are called when generating a YAML stream from an object." and AstropyLoader is similar from the constructor end.  It looks like they derived from PyYAML.  Can you clarify why "!Camera" is used in `camera_representer` and AstropyLoader?

I don't see a place where these are actually *implemented*` in `PhotometryData` so I am guessing on any astropy table (or in our case, the subclasses of them) we import/export to ECSV will use AstropyLoader and AstropyDumper to handle the attributes of the table... in which case, do you need similar representers/constructors for `observatory`?

stellarphot/core.py
```diff
@@ -658,36 +672,26 @@ class PhotometryData(BaseEnhancedTable):
         "passband": None,
```
JuanCab, line 672: I just caught that "convenience" is misspelled "convience" in line 511.

JuanCab (commented):
Comments added.

stellarphot/core.py
mwcraig: Fixed

JuanCab (approved):
Looks good.

PR #235, opened 2023-12-26
Add Exoplanet Object

### LINKED ISSUE #186: Make an object to represent an exoplanet
Among other things, this should have:

- Epoch
- Period
- Identifier (TIC ID or name)
- Coordinates
- Depth, if known
- duration, if known
- ???

Tanner728:
Created the exoplanet object and added documentation for it based on the documentation of the aperture settings class above 

Fixes #186

### DIFF
--- a/stellarphot/settings/models.py
+++ b/stellarphot/settings/models.py
@@ -143,6 +143,9 @@ class Exoplanet(BaseModel):
     To create an `Exoplanet` object, you can pass in the epoch,
      period, identifier, coordinate, depth, and duration as keyword arguments:
 
+    >>> from astropy.time import Time
+    >>> from astropy.coordinates import SkyCoord
+    >>> from astropy import units as u
     >>> planet  = Exoplanet(epoch=Time(2455909.29280, format="jd"),
     ...                     period=1.21749 * u.day,
     ...                     identifier="KELT-1b",
@@ -170,25 +173,30 @@ class Exoplanet(BaseModel):
             QuantityType: lambda v: f"{v.value} {v.unit}",
             Time: lambda v: f"{v.value}",
         }
-    @validator
-    @classmethod
-    def validate_period(cls, values):
-        """
-        Checks that the period has physical units of time and raises an error if that is not true. 
-        """
-        if u.get_physical_type(values["period"]) != "time":
-            raise ValueError(
-                f"Period does not have time units,"
-                f"currently has {values['period'].unit} units."
-            )
 
+    @validator("period")
     @classmethod
-    def validate_duration(cls, values):
+    def validate_period(cls, value):
         """
-        Checks that the duration has physical units of time and raises an error if that is not true. 
+        Checks that the period has physical units of time and raises an error
+        if that is not true.
         """
-        if u.get_physical_type(values["duration"]) != "time":
+        if u.get_physical_type(value) != "time":
+            raise ValueError(
+                f"Period does not have time units," f"currently has {value.unit} units."
+            )
+        return value
+
+    @validator("duration")
+    @classmethod
+    def validate_duration(cls, value):
+        """
+        Checks that the duration has physical units of time and raises an error
+        if that is not true.
+        """
+        if u.get_physical_type(value) != "time":
             raise ValueError(
                 f"Duration does not have time units,"
-                f"currently has {values['duration'].unit} units."
+                f"currently has {value.unit} units."
             )
+        return value
--- a/stellarphot/settings/models.py
+++ b/stellarphot/settings/models.py
@@ -6,7 +6,15 @@ from pydantic import BaseModel, Field, conint
 
 from .autowidgets import CustomBoundedIntTex
 
-__all__ = ["ApertureSettings", "PhotometryFileSettings"]
+from astropy.time import Time
+from astropy.coordinates import SkyCoord
+from astropy.units import Quantity
+import astropy.units as u
+
+from ..core import QuantityType
+
+
+__all__ = ["ApertureSettings", "PhotometryFileSettings", "Exoplanet"]
 
 
 class ApertureSettings(BaseModel):
@@ -80,3 +88,97 @@ class PhotometryFileSettings(BaseModel):
     aperture_locations_file: Path = Field(
         filter_pattern=["*.ecsv", "*.csv"], default=""
     )
+
+
+class TimeType(Time):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        return Time(v)
+
+
+class SkyCoordType(SkyCoord):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        return SkyCoord(v)
+
+
+class Exoplanet(BaseModel):
+    """
+    Create an object representing an Exoplanet.
+
+    Parameters
+    ----------
+
+    epoch : TimeType
+        Epoch of the exoplanet as an astropy Time object, not required.
+
+    period : QuantityType
+        Period of the exoplanet as a Quantity with units of time, not required.
+
+    Identifier : Str
+        Identifier of the exoplanet as a string, required.
+
+    coordinate : SkyCoordType
+        Coordinates of the exoplanet as an astropy SkyCoord object, required. 
+    
+    depth : float
+        Depth of the exoplanet object as a float, not required.
+
+    duration : QuantityType
+        Duration of the exoplanet transit as a Quantity with units of time, not required. 
+
+    Examples
+    --------
+
+    To create an `Exoplanet` object, you can pass in the epoch, period, Identifier, coordinate, depth,
+    and duration as keyword arguments:
+
+    >>> planet  = Exoplanet(
+        epoch=Time(0, format="jd"),
+        period=0 * u.min,
+        identifier="",
+        coordinate=SkyCoord(ra="00:00:00.00", dec="+00:00:00.0", frame="icrs", unit=("hour", "degree")),
+        depth=0,
+        duration=0 * u.min,)
+    """
+    
+    epoch: TimeType | None = None
+    period: QuantityType | None = None
+    identifier: str
+    coordinate: SkyCoordType
+    depth: float | None = None
+    duration: QuantityType | None = None
+
+    class Config:
+        validate_all = True
+        validate_assignment = True
+        extra = "forbid"
+        json_encoders = {
+            Quantity: lambda v: f"{v.value} {v.unit}",
+            QuantityType: lambda v: f"{v.value} {v.unit}",
+            Time: lambda v: f"{v.value}",
+        }
+
+    @classmethod
+    def validate_period(cls, values):
+        if u.get_physical_type(values["period"]) != "time":
+            raise ValueError(
+                f"Period does not have time units,currently has "
+                f"{values['period'].unit} units."
+            )
+
+    @classmethod
+    def validate_duration(cls, values):
+        if u.get_physical_type(values["duration"]) != "time":
+            raise ValueError(
+                f"Period does not have time units,currently has "
+                f"{values['duration'].unit} units."
+            )
--- a/stellarphot/settings/models.py
+++ b/stellarphot/settings/models.py
@@ -133,8 +133,7 @@ class Exoplanet(BaseModel):
         Depth of the exoplanet.
 
     duration : `astropy.units.Quantity`, optional
-        Duration of the exoplanet transit as a Quantity with units of time,
-        not required.
+        Duration of the exoplanet transit.
 
     Examples
     --------
@@ -142,7 +141,7 @@ class Exoplanet(BaseModel):
     To create an `Exoplanet` object, you can pass in the epoch,
      period, identifier, coordinate, depth, and duration as keyword arguments:
 
-    >>> planet  = Exoplanet(epoch=Time( 2455909.29280, format="jd"),
+    >>> planet  = Exoplanet(epoch=Time(2455909.29280, format="jd"),
     ...                     period=1.21749 * u.day,
     ...                     identifier="KELT-1b",
     ...                     coordinate=SkyCoord(ra="00:01:26.9169",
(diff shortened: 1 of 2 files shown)


stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+
+class Exoplanet(BaseModel):
+    """
+    Create an object representing an Exoplanet.
+
+    Parameters
+    ----------
+
+    epoch : TimeType
```
mwcraig, line 121: Just `Time` here -- the `TimeType` is something we needed to make to get a `Time` object to play nicely with `pydantic`, but the user passes in a `Time`.

```suggestion
    epoch : `astropy.time.Time`, optional
```

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+        Epoch of the exoplanet as an astropy Time object, not required.
+
+    period : QuantityType
```
mwcraig, line 124:
```suggestion
    period : `astropy.units.Quantity`, optional
```

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+        Period of the exoplanet as a Quantity with units of time, not required.
```
mwcraig, line 125: Instead of putting "not required" here, put "optional in the line above"

mwcraig: Also no need to mention the `Quantity` thing again -- just "Period of the exoiplanet with units of time' would be excellent1

stellarphot/settings/models.py
mwcraig: Adding the backticks and the location of `Quantity` in `astropy` will generate a link in the documentation to astropy `Qunatity` documentation.

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+    Identifier : Str
```
mwcraig, line 127:
```suggestion
    Identifier : str
```

Python types are lower case.

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+class Exoplanet(BaseModel):
+    """
+    Create an object representing an Exoplanet.
+
+    Parameters
+    ----------
+
+    epoch : TimeType
+        Epoch of the exoplanet as an astropy Time object, not required.
```
mwcraig, line 122: Just "Epoch of the exoplanet" would be good here.

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+    period : QuantityType
+        Period of the exoplanet as a Quantity with units of time, not required.
+
+    Identifier : Str
+        Identifier of the exoplanet as a string, required.
```
mwcraig, line 128: Drop the "as a string" and "required" -- all of the optional stuff is indicated as optional so any stuff not optional is required.

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+    coordinate : SkyCoordType
```
mwcraig, line 130:
```suggestion
    coordinate : `astropy.coordinates.SkyCoord`
```

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+        Coordinates of the exoplanet as an astropy SkyCoord object, required.
```
mwcraig, line 131: Update this too (no type, no required)

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+    depth : float
+        Depth of the exoplanet object as a float, not required.
```
mwcraig, line 134: Same here

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+    duration : QuantityType
```
mwcraig, line 136:
```suggestion
    duration : `astropy.units.Quantity`, optional
```

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+        Duration of the exoplanet transit as a Quantity with units of time,
+        not required.
+
+    Examples
+    --------
+
+    To create an `Exoplanet` object, you can pass in the epoch,
+    period, Identifier, coordinate, depth, and duration as keyword arguments:
```
mwcraig, line 144:
```suggestion
    period, identifier, coordinate, depth, and duration as keyword arguments:
```

stellarphot/settings/models.py
```diff
@@ -80,3 +89,107 @@ class PhotometryFileSettings(BaseModel):
...
+
+    >>> planet  = Exoplanet(epoch=Time(0, format="jd"), period=0 * u.min,
```
mwcraig, line 146: It is probably more helpful if you put in information for a real exoplanet. Maybe Kelt-1?

stellarphot/settings/tests/test_models.py
```diff
@@ -1,8 +1,12 @@
+from astropy.time import Time
 from pydantic import ValidationError
 import pytest
 
 from stellarphot.settings.models import ApertureSettings
+from stellarphot.settings.models import Exoplanet
```
mwcraig, line 6: I would just add this to import on the previous line

mwcraig (changes requested):
Excellent! I left a number o minor inline comments below. 

In addition, can you please add a test to check that you get a validation error when you pass in a period and a duration without a unit (or with the wrong kind of unit)?

stellarphot/settings/models.py
Tanner728: updated to use kelt-1b info but all of it may not be correct. kinda estimated values based on the graphs from a paper i found. also the depth is based on normalized flux

mwcraig: You can look up exoplanet data here: https://exoplanetarchive.ipac.caltech.edu/

There are likely several values for kelt-1, maybe use the most recent.

stellarphot/settings/models.py
```diff
@@ -80,3 +89,108 @@ class PhotometryFileSettings(BaseModel):
...
+        Identifier of the exoplanet.
+
+    coordinate : `astropy.coordinates.SkyCoord`
+        Coordinates of the exoplanet.
+
+    depth : float
+        Depth of the exoplanet.
+
+    duration : `astropy.units.Quantity`, optional
+        Duration of the exoplanet transit as a Quantity with units of time,
```
mwcraig, line 137: Please update this docstring

stellarphot/settings/tests/test_models.py
```diff
@@ -1,8 +1,11 @@
+from astropy.time import Time
 from pydantic import ValidationError
 import pytest
 
-from stellarphot.settings.models import ApertureSettings
+from stellarphot.settings.models import ApertureSettings,Exoplanet
```
mwcraig, line 5:
```suggestion
from stellarphot.settings.models import ApertureSettings, Exoplanet
```

mwcraig:
Looking good -- the main thing still missing is a test for invalid values for period and duration.

Tanner728:
For the validation test would that be similar to this: 
```python 
@pytest.mark.parametrize("bad_one", ["radius", "gap", "annulus_width"])
def test_create_invalid_values(bad_one):
    # Check that individual values that are bad raise an error
    bad_settings = DEFAULT_APERTURE_SETTINGS.copy()
    bad_settings[bad_one] = -1
    with pytest.raises(ValidationError, match=bad_one):
        ApertureSettings(**bad_settings)
```

mwcraig:
Yep, something like that would work

stellarphot/settings/tests/test_models.py
```diff
@@ -29,3 +33,43 @@ def test_create_invalid_values(bad_one):
...
+    planet = Exoplanet(**DEFAULT_EXOPLANET_SETTINGS)
+    assert planet.epoch == DEFAULT_EXOPLANET_SETTINGS["epoch"]
+    assert u.get_physical_type(planet.period) == "time"
+    assert planet.identifier == DEFAULT_EXOPLANET_SETTINGS["identifier"]
+    assert planet.coordinate == DEFAULT_EXOPLANET_SETTINGS["coordinate"]
+    assert planet.depth == DEFAULT_EXOPLANET_SETTINGS["depth"]
+    assert u.get_physical_type(planet.duration) == "time"
+
+
+@pytest.mark.parametrize("bad_one", [
```
mwcraig, line 61: Usually `parametrize` is used to when you want to test with multiple values rather than a single value. I could see going two ways here:

1. Test two values, like in the example you mentioned [here](https://github.com/feder-observatory/stellarphot/pull/235#issuecomment-1870454532). That maybe looks like `@pytest.mark.parametrize('bad_one', ('period', 'duration'))` and then in the test set the value ot something bad like in the example.
2. Test with one set of settings in which both duration and period are bad (inappropriate units) and make sure the error message has the right number of `ValidationErrors`. See [this test](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/tests/test_core.py#L52) for an example of that approach.

Either way, you should be able to start with a copy of the default exoplanet and modify values of that rather than reproducing all the settings.

stellarphot/settings/tests/test_models.py
```diff
@@ -29,3 +33,43 @@ def test_create_invalid_values(bad_one):
...
+        "identifier": "bad-planet",
+        "coordinate": SkyCoord(ra="00:00:00.00", dec="+00:00:00.0",
+                               frame="icrs", unit=("hour", "degree")),
+        "depth": 0,
+        "duration": 42 * u.C,
+    }
+])
+def test_create_invalid_exoplanet(bad_one):
+    # Check that individual values that are bad raise an error
+    with pytest.raises(ValidationError, match=".*"):
```
mwcraig, line 74: Typically if `match` is included you should try to match some meaningful part of the error message. This particular value `".*"` means it will always match, which is equivalent to not having `match` there at all.

mwcraig:
A couple broader comments:

+ Please try to fix the linting issues before committing so there aren't a bunch of lint fix commit.
+ Don't rely on what VS Code says is too long a line unless you update the line length to 88. It seems like in several cases you have "overshortened" the lines.

stellarphot/settings/tests/test_models.py
Tanner728: So I redid the tests and just have a question about the order of operation for the tests. Does this pytest run the validation tests that are in the exoplanet class that check the physical units of duration and period? When I gave period and duration units of cm it would pass this pytest.

Tanner728:
should i just ignore the 'line too long' issues or is there something else I can check to see if lines are too long?

mwcraig:
> should i just ignore the 'line too long' issues or is there something else I can check to see if lines are too long?

I would run `tox -e lint` before you commit -- if it doesn't complain about lines being too long then you are OK. When we talked yesterday I thought your vscode was marking lines longer than 79 characters even though the project line length limit is 88.

stellarphot/settings/tests/test_models.py
mwcraig: It should run the validation checks and then raise a single `ValidationError` (rather than several `ValueError`s) with a summary of the errors it encountered.

stellarphot/settings/models.py
```diff
@@ -80,3 +91,107 @@ class PhotometryFileSettings(BaseModel):
...
+        validate_all = True
+        validate_assignment = True
+        extra = "forbid"
+        json_encoders = {
+            Quantity: lambda v: f"{v.value} {v.unit}",
+            QuantityType: lambda v: f"{v.value} {v.unit}",
+            Time: lambda v: f"{v.value}",
+        }
+
+    @classmethod
```
mwcraig, line 175: this needs one more decorator to finish turning it into something that pydantic actually validates, something like `@validator("period")` (after importing `validator` from `pydantic`).

See https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/core.py#L289 for an example. That is why leaving the units off didn't generate a validation error.  Let's ,make sure it does by only making those two values bad.

mwcraig: @Tanner728 -- do you have time to add this change or should I wrap it up instead?

Tanner728: I should have time tonight to do it.

Tanner728:
After adding the validator I get this error message when running the tests and am somewhat stumped about where its coming from. 

py::test_create_exoplanet_correctly failed: def test_create_exoplanet_correctly():
>       planet = Exoplanet(**DEFAULT_EXOPLANET_SETTINGS)

stellarphot\settings\tests\test_models.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for Exoplanet
E   period
E     'Quantity' object with a scalar value does not support indexing (type=type_error)

pydantic\main.py:341: ValidationError

mwcraig:
Can you push the commit the change you made or paste the updarted code into a comment so I can take a look?

mwcraig:
Looks like you pushed a merge commit? You shouldn't need a merge commit....

stellarphot/settings/models.py
mwcraig: Looks like you left out the "period" argument to `@validator`....

mwcraig:
FYI, just updated this to finish off the validators (I think I did, at least, will see what the tests say)

mwcraig (approved):
Thanks, @Tanner728

Tanner728:
Thank you @mwcraig. I am currently driving out to Montana but will send you a draft of the proceedings paper tonight. Thanks for finishing that validator off.

PR #237, opened 2023-12-31
Check camera json encoding

mwcraig:
This fixes an oversight from #231 in which I forgot to add a json encoder for units.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -4,7 +4,7 @@ from astropy import units as u
 from astropy.coordinates import EarthLocation, SkyCoord
 from astropy.table import Column, QTable, Table
 from astropy.time import Time
-from astropy.units import Quantity, Unit
+from astropy.units import Quantity, Unit, IrreducibleUnit
 from astropy.wcs import WCS
 
 from astroquery.vizier import Vizier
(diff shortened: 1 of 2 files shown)


stellarphot/tests/test_core.py
```diff
@@ -193,6 +193,28 @@ def test_camera_schema():
...
+    c = Camera(
+        data_unit=data_unit,
+        gain=gain,
+        read_noise=read_noise,
+        dark_current=dark_current,
+        pixel_scale=pixel_scale,
+        max_data_value=max_val,
+    )
+
+    c2 = Camera.parse_raw(c.json())
```
JuanCab, line 214: Seems like a reasonable test.

stellarphot/core.py
```diff
@@ -240,7 +240,8 @@ class Config:
         json_encoders = {
             Quantity: lambda v: f"{v.value} {v.unit}",
             QuantityType: lambda v: f"{v.value} {v.unit}",
-            UnitType: lambda v: f"{v}",
+            Unit: lambda v: f"{v}",
```
JuanCab, line 243: I assume this sets it up some you store both the units and irreducible units in JSON.   Do you know why this is more 'round-trippable'?  Is it because UnitType isn't an actual astropy object but Unit and IrreducibleUnit are?

JuanCab (approved):
Looks good assuming the test passes.

stellarphot/core.py
mwcraig: I think it is because the encoding is done based on the type of the setting, eg. `Unit` or `IrreducibleUnit`. I could have used just `IrreducibleUnit` here but in principle units could be more complicated, e.g. aand image data unit of `electron / sec` or something.

PR #239, opened 2024-01-01
Fix photometry test data

mwcraig:
There was an error in the way the photometry test data was being created that led to a `Time` column not being savable.


mwcraig:
Self-merging since this is pretty small...

JuanCab (commented):
Looks reasonable.

PR #240, opened 2024-01-01
Fix photometry error message

mwcraig:
Fixes a minor error in `BaseEnhancedTable` initialization


mwcraig:
Self-merging since this is pretty small...

PR #242, opened 2024-01-02
Fix bug in catalog search functionality and make comparison notebook compatible with `CatalogData`

### LINKED ISSUE #241: CatalogData.from_vizier should accept a WCS as its first argument
Right now `CatalogData.from_vizier` accepts either a coordinate or a FITS header as its first argument. It needs to also accept a WCS. The reason is that when you read a FITS image with `CCDData.read` the WCS information is automatically stripped from the header and used to create a WCS object.

mwcraig:
This PR started as a way to get the comparison notebook working again. Along the way I found a couple of issues in `CatalogData.from_vizier` that needed fixing.

Fixes #241


stellarphot/core.py
```diff
@@ -1106,10 +1106,10 @@ def from_vizier(
         Parameters
         ----------
 
-        header_or_center : FITS header or `astropy.coordinates.SkyCoord`
-            Either a FITS header with WCS information or a `SkyCoord` object.
-            The center of the frame or the input coordinate is the center
-            of the cone search.
+        location : `astropy.coordinates.SkyCoord`, `astropy.wcs.WCS`, or FITS header or
```
JuanCab, line 1109: Might want to edit some more, this documentation is not grammatically correct (I suspect the last 'or' in line 1109 should not be there).  Also extra space between words 'or' and 'a' at beginning of line 1111.  Honestly, I think the last sentence could be clearer that if you entered SkyCoord it represents center of frame or otherwise the center of the frame is extrapolated from the WCS information provided, just to clarify where the information is obtained.  But you know me, I am wordy. :)

stellarphot/core.py
JuanCab, line 1: For consistency, do you want to rename the `header_or_center` parameter of `apass_dr9` to `location` (of `field_center`, see below) and update its documentation  to indicate it accepts ` Either a `~astropy.coordinates.SkyCoord` object, a `~astropy.wcs.WCS` object or  a FITS header with WCS information.` (that is, just copy the documentation of `from_vizier`)?

stellarphot/tests/test_core.py
```diff
@@ -1015,6 +1041,16 @@ def test_from_vizier_with_coord_and_frame_clip_fails():
         _ = CatalogData.from_vizier(cen_coord, "B/vsx/vsx", clip_by_frame=True)
 
 
+def test_from_vizier_with_header_no_wcs_raise_error():
```
JuanCab, line 1044: I was going to see if you added a test for the no-header situation, glad you did.

stellarphot/tests/test_core.py
```diff
@@ -963,10 +963,36 @@ def test_catalog_from_vizier_search_apass():
...
+        # The projection doesn't matter because we only need the center
+        # coordinate. The code below is more or less copied from
+        #
+        wcs = WCS(naxis=2)
+        # Put the center in the right place
+        wcs.wcs.crpix = [100, 100]
+        wcs.wcs.crval = [coordinate.ra.degree, coordinate.dec.degree]
+        wcs.array_shape = [200, 200]
+
+        # The rest of these calues shouldn't matter for this test
```
JuanCab, line 985: 'calues' should be 'values'

stellarphot/tests/test_core.py
```diff
@@ -963,10 +963,36 @@ def test_catalog_from_vizier_search_apass():
...
     # Do a cone search with a small enough radius to return exaclty one star,
     # DQ Psc, which happens to already be in the test data.
     coordinate = SkyCoord(ra=359.94371 * u.deg, dec=-0.2801 * u.deg)
+
+    if location_method == "coord":
+        center = coordinate
+    else:
+        # We need a WCS for each of these methods, so make that first.
+        # The projection doesn't matter because we only need the center
+        # coordinate. The code below is more or less copied from
```
JuanCab, line 977: "copied from " where? I want to know....

stellarphot/tests/test_core.py
```diff
@@ -963,10 +963,36 @@ def test_catalog_from_vizier_search_apass():
 
 
 @pytest.mark.remote_data
-def test_catalog_from_vizier_search_vsx():
+@pytest.mark.parametrize("location_method", ["coord", "header", "wcs"])
+def test_catalog_from_vizier_search_vsx(location_method):
     # Do a cone search with a small enough radius to return exaclty one star,
     # DQ Psc, which happens to already be in the test data.
     coordinate = SkyCoord(ra=359.94371 * u.deg, dec=-0.2801 * u.deg)
+
```
JuanCab, line 971: The logic of the test seems sound for testing your three location determining approaches.

stellarphot/core.py
```diff
@@ -1089,7 +1089,7 @@ def _tidy_vizier_catalog(data, mag_column_regex, color_column_regex):
     @classmethod
     def from_vizier(
         cls,
-        header_or_center,
+        location,
```
JuanCab, line 1092: Given we use `location` elsewhere in the context of observatory, might I suggest 'field_center'?  Not critical, but seemed like a way to avoid confusion.

stellarphot/tests/test_core.py
JuanCab, line 1: It all looks good, logic is sound.  If you change name of `location` to `field_center` you might rename the test parameter, but that's probably overkill.

stellarphot/core.py
JuanCab, line 1: `apass_dr9` and `vsx_vizier` should have their documentation revised to indicate their return type is `stellarphot.CatalogData`.

stellarphot/utils/comparison_utils.py
JuanCab, line 1: You need to fix the documentation on `crossmatch_APASS2VSX`  to indicate `apass` value returned is `stellarphot.CatalogData` instead of an astropy table. 

Since you didn't state it in the notes on this PR, was the bug in line 81, telling it units of RA were hours when CatalogData uses degrees?

`mag_scale` documentation should probably note you are filtering to only keep `r` bandpass and WHY you are doing this (since you added it here).

JuanCab (commented):
See the comments, the logic was sound but there were several documentation issues that I could see.

stellarphot/core.py
mwcraig: Good idea, done!

stellarphot/core.py
mwcraig: Excellent suggestion, done!

stellarphot/utils/comparison_utils.py
mwcraig: Done

stellarphot/tests/test_core.py
mwcraig: Added it :)

mwcraig:
I think I've addressed all of your comments, thanks for the review!

stellarphot/core.py
JuanCab, line 1: Looks like you addressed everything.

stellarphot/utils/comparison_utils.py
JuanCab, line 1: Still would like documentation of reason for the use of the r-band filter (aka `good_filter`) in `mag_scale`.

JuanCab (commented):
Almost there, missing a clarification of use of r-band filter in `mag_scale` in `comparison_utils.py`.

stellarphot/utils/comparison_utils.py
mwcraig: Good catch -- I just turned it into an argument with a default value of "r". The reason for using "r" is mostly historical (and weird): it is the filter we have most often used here.

JuanCab (approved):
Looks good!

PR #253, opened 2024-01-11
Tidy up Camera-related tests and add a name property

mwcraig:
My original intent here was to just a dd a name property; that allows storing at least a little information that is human-readable to identify the camera if looking at the settings much later.

Since that would have meant many, many changes to the tests I first rewrote the tests a bit so that `Camera` only needed to be changed for the new property in one or two places.

I think that ended up making the tests a little bit clearer because it is now more explicit what property of the camera is being changed/modified in a particular ttest.

### DIFF
--- a/stellarphot/photometry/tests/test_photometry.py
+++ b/stellarphot/photometry/tests/test_photometry.py
@@ -29,9 +29,12 @@ SEED = 5432985
 
 SHIFT_TOLERANCE = 6
 FWHM_ESTIMATE = 5
+
+# A camera with not unreasonable settings
 FAKE_CAMERA = Camera(
     data_unit=u.adu,
     gain=1.0 * u.electron / u.adu,
+    name="test camera",
     read_noise=0 * u.electron,
     dark_current=0.1 * u.electron / u.second,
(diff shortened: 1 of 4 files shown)


mwcraig:
@JuanCab -- it may be easier to look at the two commits separately. The first commit condenses the tests, the second commit add the name property.

stellarphot/photometry/tests/test_photometry.py
JuanCab, line 1: Rather like this approach of having a ZERO_CAMERA and then just copying it and changing the attributes.  Cleaner test code!

JuanCab (approved):
It looked good, in fact, I rather like the 'trimming' you have been doing tightening up this code (particularly the testing code).  I also like how the test for alt units being ok (`test_camera_altunitscheck`) is now just to assert the dictionary for the camera is still the same one you used to create it (I assume this is thanks to pydantic's checking).

stellarphot/photometry/tests/test_photometry.py
mwcraig: 😀 thanks!

PR #256, opened 2024-01-12
Properly set title in FitsOpener

mwcraig:
Turns out #255 introduced a bug, because the ipyautoui FileChooser discards the title argument in its initializer.


mwcraig:
Test failures are all due to remote data, merging...

PR #258, opened 2024-01-14
Calculate an ID for APASS following IAU guidelines

### LINKED ISSUE #257: Vizier ID numbers for APASS really shouldn't be used as an ID because  they change
We picked up a new test failure, I think because the ID we were using from Vizer in the test apparently changes. The search was a 0.03 deg search about re, dec of 0, 0 that returns one star. The ID number had been `17672748` on `2023-11-28` according to the comment in `test_catalog_from_vizier_search_apass`. Today the number is `17672747`; see [this run log](https://github.com/feder-observatory/stellarphot/actions/runs/7512788423/job/20453843485#step:7:105) reproduced in part below.

To be fair, Vizier warns that the ID is assigned by Vizier and should not be used to identify things...

```python-traceback
=================================== FAILURES ===================================
____________________ test_catalog_from_vizier_search_apass _____________________

    @pytest.mark.remote_data
    def test_catalog_from_vizier_search_apass():
        # Nothing special about this point...
        sc = SkyCoord(ra=0, dec=0, unit="deg")
    
        # Small enough radius to get only one star
        radius = 0.03 * u.deg
    
        apass_colnames = {
            "recno": "id",  # There is no APASS ID, this is the one generated by Vizier
            "RAJ2000": "ra",
            "DEJ2000": "dec",
        }
    
        apass = CatalogData.from_vizier(
            sc,
            "II/336/apass9",
            radius=radius,
            colname_map=apass_colnames,
            clip_by_frame=False,
        )
        assert len(apass) == 6
    
        # Got the values below from Vizier on 2023-11-28
>       assert apass["id"][0] == 17672748
E       assert 17672747 == 17672748

stellarphot/tests/test_core.py:763: AssertionError
```

mwcraig:
Fixes #257

This code had used the "recno" from Vizier as the ID column for an APASS catalog despite Vizier's clear warning that it "Should not be used for identification." Turns out that recno does not stay the same over time. Instead, we calculate an identifier based on the coordinates following https://cds.unistra.fr/Dic/iau-spec.html


JuanCab (approved):
Seems reasonable to use the IAU guidelines and the test confirms it works.

PR #263, opened 2024-01-24
Add more tests of our pydantic models prior to upgrading to pydantic 2

mwcraig:
I think I know how to upgrade to pydantic 2 now, but before I do that PR I thought it was prudent to write more tests of our pydantic models and their uses in notebooks. 

If you have a chance to look at this, @JuanCab, great, but I think it is fine to self-merge since this mostly adds tests.

The tests I wrote did uncover a couple bugs -- places where we hadn't properly added handling of the new `FWHM` property on `PhotometryApertures`.

### DIFF
--- a/stellarphot/gui_tools/fits_opener.py
+++ b/stellarphot/gui_tools/fits_opener.py
@@ -128,3 +128,4 @@ class FitsOpener:
 
         self._fc.reset(directory, file)
         self._fc._apply_selection()
+        self._fc._value = self._fc.selected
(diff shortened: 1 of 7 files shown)


stellarphot/gui_tools/fits_opener.py
```diff
@@ -128,3 +128,4 @@ def set_file(self, file, directory=None):
 
         self._fc.reset(directory, file)
         self._fc._apply_selection()
+        self._fc._value = self._fc.selected
```
JuanCab, line 131: I am a little wary of using "protected" attributes/methods of a class... I have done it, but I am wary of this.

JuanCab (approved):
As far as I can tell the tests seem reasonable.  Only issue (and not really that) is I noticed you accessing private attributes/methods.  Not worth blocking the merge over that.

stellarphot/gui_tools/fits_opener.py
mwcraig: I had the same reservations, but figured it could wait....

PR #268, opened 2024-02-05
Add an Observatory class

### LINKED ISSUE #98: Consider removing Feder specific code in stellarphot
I poked around and identified the following places where the code either assumes or refers to Feder Observatory in the current version of `stellarphot`.  I found the following:

- [x] `photometry.py`
  - **Assumes only Feder!** `add_to_photometry_table()`: Turns out code only adds BJD field to `phot` if the observatory is Feder (lines 381-383: Codes for Feder Observatory BJD)
  - `find_times()`: latitude and longitude defaults to Feder (but does allow other choices)
- [x] `visualization/comparison_functions.py` is hardcoded for Feder!
  - [x] **Assumes only Feder!**  In `ComparisonViewer/_set_object` line 513 sets `telescope_code` for `tess_submission` to Feder
  - [x] **Assumes only Feder!**  In `ComparisonViewer/_make_tess_save_box` line 625 sets displayed telescope code to Feder, no other option.
- [x] `visualization/seeing_profile_functions.py` is hardcoded for Feder!
  -  [x] **Assumes only Feder!** In `SeeingProfileWidget/_make_tess_save_box` line 447 sets displayed telescope code to Feder, no other option.
  -  [x] **Assumes only Feder!** In `SeeingProfileWidget/_update_plots` the read noise and gain are set to the values for our camera (CG16M).
```[tasklist]
### Tasks
```

mwcraig:
This pull request 

- adds an Observatory class
- Moves the functionality needed to serialize models to an astropy table into a base class on which others build
- Consolidates a bunch of tests into a test class that is parameterized with the models defined in the code
- Uses the `Observatory` class in a couple of cases to replace Feder-specific code

A couple comments about the motivation for adding the Observatory class, and at least one alternative I looked at:

+ Right now we represent an observatory with an `EarthLocation`. There is some additional information, like AAVSO observer code and TESS designation that it makes sense to me to include in an observatory definition.
+ I started with an `EarthLocation` as one of the properties of `Observatory` but had trouble getting that to work with auto-generation of a widgets UI to enter it, so I went with taking lat, long and altitude as input and providing a property with the `EarthLocation`.
+ I looked at the [`Observer` class](https://astroplan.readthedocs.io/en/stable/api/astroplan.Observer.html#astroplan.Observer) from astroplan. That has much more than we need, some of which might be useful, but would mean adding a new dependency and I wasn't sure how it would work with pydantic. One thought is to offer it as a property with astroplan as an optional dependency. Or we just punt the whole issue down the road.

@JuanCab -- if it is possible to do this oen this week that would be great, but if not let me know. 

Fixes #98

### DIFF
--- a/stellarphot/gui_tools/comparison_functions.py
+++ b/stellarphot/gui_tools/comparison_functions.py
@@ -278,7 +278,7 @@ class ComparisonViewer:
 
     def _init(self):
         """
-        Handles aspects of initialization that need to be defered until
+        Handles aspects of initialization that need to be deferred until
         a file is chosen.
         """
         if self.tess_submission is not None:
@@ -365,10 +365,10 @@ class ComparisonViewer:
                 planet=1,
             )
         except ValueError:
-            # Guess not, time to turn on the coordinates box
-            # self._turn_on_coordinates()
+            # Not a TESS object, so turn off the TESS interface
             _settings_for_no_tess()
         else:
+            # This is a TESS object, so set up the TESS interface
             self.tess_save_toggle.disabled = False
             self.tess_save_toggle.layout.visibility = "visible"
             self.toi_info = TOI(self.tess_submission.tic_id)
--- a/stellarphot/gui_tools/comparison_functions.py
+++ b/stellarphot/gui_tools/comparison_functions.py
@@ -195,6 +195,13 @@ class ComparisonViewer:
     overwrite_outputs: bool, optional
         Whether to overwrite existing output files. Defaults to True.
 
+    observatory : `stellarphot.settings.Observatory`, optional
+        Observatory information. If this is set and the observatory has
+        a `stellarphot.settings.Observatory.TESS_telescope_code` attribute,
+        then the TESS submission information will be extracted from the header
+        and a button will appear to save images of the field for a TESS
+        submission.
+
     Attributes
     ----------
 
@@ -242,6 +249,7 @@ class ComparisonViewer:
         object_coordinate=None,
         photom_apertures_file=None,
         overwrite_outputs=True,
+        observatory=None,
     ):
         self._label_name = "labels"
         self._circle_name = "target circle"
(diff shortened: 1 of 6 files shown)


stellarphot/gui_tools/comparison_functions.py
```diff
@@ -346,8 +350,27 @@ def _set_object(self):
...
+            self.tess_submission = TessSubmission.from_header(
+                self._file_chooser.header,
+                telescope_code=self.observatory.TESS_telescope_code,
+                planet=1,
+            )
+        except ValueError:
+            # Guess not, time to turn on the coordinates box
+            # self._turn_on_coordinates()
+            _settings_for_no_tess()
         else:
```
JuanCab, line 371: Since I always forget this, I might add a comment line after `else` to indicate this bit of code gets run if `try` succeeds.  Or not.  I constantly do reminders like this for features I use less often.

stellarphot/gui_tools/comparison_functions.py
```diff
@@ -484,11 +507,14 @@ def _make_tess_save_box(self):
...
         )
+        self.tess_save_toggle.layout.visibility = "hidden"
         self._tess_save_box = ipw.VBox()
         self._tess_save_box.layout.visibility = "hidden"
 
+        tess_scope = self.observatory.TESS_telescope_code if self.observatory else None
+        tess_scope = "" if tess_scope is None else tess_scope
         scope_name = ipw.Text(
-            description="Telescope code", value="Paul-P-Feder-0.4m", style=DESC_STYLE
+            description="Telescope code", value=tess_scope, style=DESC_STYLE
```
JuanCab, line 517: Finally!

JuanCab (dismissed):
It all looks good as far as I can tell.  I only caught a few issues with comments which honestly don't really need to be fixed unless you wish to address them.  They are very minor.

stellarphot/gui_tools/comparison_functions.py
mwcraig: You have no idea the amount of pain in setting up a form for the observatory 😬 😀

PR #269, opened 2024-02-06
Format required columns as RST tables for type in core

mwcraig:
Currently, the docstrings of `PhotometryData`, `CatalogData`, and `SourceList` each contain a list of required columns. These render in the documentation quite badly. The list is in the yellow box below:

PhotometryData_—_stellarphot_v0_1_dev233_g383d1c1_d20240205

This pull request formats these as rst tables by adding a few equal signs here and there. After this, the same section of the documentation will look like this:

PhotometryData_—_stellarphot_v1_3_10_dev590_g383d1c1_d20240206

@JuanCab, I'll self-merge this unless you object....

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -307,7 +307,8 @@ class PhotometryData(BaseEnhancedTable):
     simply means it can be any unit, but it must be the same for all the columns with
     'consistent count units'.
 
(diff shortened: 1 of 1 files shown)


JuanCab (approved):
Completely reasonable

PR #270, opened 2024-02-07
Update notebook 3 for change in photometry arguments

mwcraig:


mwcraig:
Going extra wild today and self-merging before tests, since the only change was in a notebook we don't test.

PR #272, opened 2024-02-08
Rearrange photometry settings and add object for photometry options

### LINKED ISSUE #149: Function signatures for photometry functions
The new photometry routines, like the old ones, have a fairly large number of required arguments. It would be nice to simplify that, by either grouping things into objects, or making them keyword arguments so that there is at least some indication in the function call what the arguments are.

Edit:

```[tasklist]
### Tasks
- [x] Provide a settings object for the arguments with defaults in photometry
- [x] Maybe make interface class-based
- [ ] Consolidate settings into a single settings object by composing the sub-settings
```

mwcraig:
This is another fairly big one. It does a few things:

- Rearrange the order of keyword arguments in `single_image_photometry` and `multi_image_photometry` so the order is the same, with keywords specific to each function at the end of each signature.
- Adds a `PhotometryOptions` object that includes `shift_tolerance` and all of the keyword options that were common to `single_image_photometry` and `multi_image_photometry`. Also adds tests of the new object.
- Uses that object in place of `shift_tolerance` and the keyword options that were common to both.
- Modifies tests as needed to accommodate the new arguments.
- Fixes a few bugs where arguments in `multi_image_photometry` were not being passed through to `single_image_photometry` 😀 

@JuanCab -- my step will be to use these to turn the photometry functions into a class. I'm going to start building that based on this, but can of course modify in response to suggestions here. 

If you can review this one by early next week that would be great. The next one, "class"-ifying photometry, will be much less extensive than this I think.

Closes #149

### DIFF
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -589,7 +589,7 @@ def multi_image_photometry(
     sourcelist,
     camera,
     observatory_location,
-    photometry_settings,
+    photometry_apertures,
     shift_tolerance,
     include_dig_noise=True,
     reject_too_close=True,
@@ -631,7 +631,7 @@ def multi_image_photometry(
         Location of the observatory where the images were taken.  Used for calculating
         the BJD.
 
-    photometry_settings : `stellarphot.settings.PhotometryApertures`
+    photometry_apertures : `stellarphot.settings.PhotometryApertures`
         Radius, inner and outer annulus radii settings and FWHM.
 
     shift_tolerance : float
@@ -772,7 +772,7 @@ def multi_image_photometry(
             sourcelist,
             camera,
             observatory_location,
-            photometry_settings,
+            photometry_apertures,
             shift_tolerance,
             use_coordinates="sky",
             include_dig_noise=include_dig_noise,
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -83,7 +83,8 @@ def single_image_photometry(
         Radius, inner and outer annulus radii settings and FWHM.
 
     photometry_options : `stellarphot.settings.PhotometryOptions`
-        Several options for the details of performing the photometry.
+        Several options for the details of performing the photometry. See the
+        documentation for `~stellarphot.settings.PhotometryOptions` for details.
 
     passband_map: dict, optional (Default: None)
         A dictionary containing instrumental passband names as keys and
@@ -581,7 +582,8 @@ def multi_image_photometry(
         Radius, inner and outer annulus radii settings and FWHM.
 
     photometry_options : `stellarphot.settings.PhotometryOptions`
-        Several options for the details of performing the photometry.
+        Several options for the details of performing the photometry. See the
+        documentation for `~stellarphot.settings.PhotometryOptions` for details.
 
     passband_map: dict, optional (Default: None)
         A dictionary containing instrumental passband names as keys and
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -45,12 +45,12 @@ def single_image_photometry(
     include_dig_noise=True,
     reject_too_close=True,
     reject_background_outliers=True,
-    passband_map=None,
     fwhm_by_fit=True,
-    fname=None,
-    logline="single_image_photometry:",
     logfile=None,
     console_log=True,
+    passband_map=None,
+    fname=None,
+    logline="single_image_photometry:",
 ):
     """
     Perform aperture photometry on a single image, with an options for estimating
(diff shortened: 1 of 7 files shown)


stellarphot/photometry/tests/test_photometry.py
```diff
@@ -514,14 +535,26 @@ def test_photometry_on_directory():
...
-        # positions.
-
-        # Here we just check whether any difference is consistent with
-        # less than the expected one sigma deviation.
-        assert np.abs(expected_flux - obs_avg_net_cnts) < expected_deviation
+        # We have two cases to consider: use_coordinates="sky" and
+        # use_coordinates="pixel". In the former case, the expected
+        # result is the test below. In the latter case, the expected
+        # result is that either obs_avg_net_cnts is nan or the difference
+        # is bigger than the expected_deviation.
```
JuanCab, line 542: Am I misreading this?  You say you expect the different is BIGGER than the expected_deviation, but below (line 556) you assert it is smaller, no?

Also, just for my sake, can you clarify why the noise will be different in sky coords versus pixel coord?

stellarphot/settings/models.py
```diff
@@ -410,6 +412,68 @@ def earth_location(self):
         )
 
 
+class PhotometryOptions(BaseModelWithTableRep):
```
JuanCab, line 415: Are we providing a function for setting the PhotometryOptions or does pydantic do that for us?

stellarphot/settings/tests/test_models.py
```diff
@@ -288,6 +304,16 @@ def test_observatory_lat_long_as_float():
     assert obs == Observatory(**DEFAULT_OBSERVATORY_SETTINGS)
 
 
+def test_photometry_settings_negative_shift_tolerance():
```
JuanCab, line 307: Should we test the other parameters of `PhotometryOptions`?

JuanCab (commented):
Mostly good (like usual).  I caught a couple of issues.  Two big ones:
1. Do we need more tests of PhotometryOptions class?
2. Do we need to provide a function to the user to set the PhotometryOptions?

Other than that, it's all the small documentation issues I usually see.  I do think we should be clearer to users when listing `photometry_options`  as a parameter for the Single image and Multi-image photometry who the user can set those parameters.  I would also consider listing them in the documentation for the photometry routines.

stellarphot/photometry/photometry.py
```diff
@@ -75,69 +73,30 @@ def single_image_photometry(
...
-
-    reject_too_close : bool, optional (Default: True)
-        If ``True``, any sources that are closer than twice the aperture radius
-        are rejected.  If ``False``, all sources in field are used.
-
-    include_dig_noise : bool, optional (Default: True)
-        If ``True``, include the digitization noise in the calculation of the
-        noise for each observation.  If ``False``, only the Poisson noise from
-        the source and the sky will be included.
+    photometry_options : `stellarphot.settings.PhotometryOptions`
```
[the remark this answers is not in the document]
mwcraig: Not your fault photometry is complicated

mwcraig:
> Mostly good (like usual). 

Thanks 😀 

> Do we need more tests of PhotometryOptions class?

I don't think so because it gets all of the tests that are [here](https://github.com/feder-observatory/stellarphot/pull/272/files#diff-748a2d067b95b98944915d9fb04079350dc103bee2b594a1f1ba10115f1d2bf7L60) in addition for the test for the one parameter that has a constraint. 

> Do we need to provide a function to the user to set the PhotometryOptions?

I'm adding an example of creating one to the `PhotometryOptions` documentation, and added a second link to those from the single/multi image docstrings.

stellarphot/photometry/tests/test_photometry.py
mwcraig: Two things

+ I rearranged the comments to make clear what the passing condition is in each case.
+ I think there is a bug somewhere, either
    - In `list_of_image` or in the shifting stuff that calls. The reason I say that is I think we should be able to use either pixel or sky coordinates and have this pass. In practice, the image first image is fine but the second gets negative net counts and bad fwhms.
    - Maybe in `multi_image_photometry` since that treats the first image different than the rest.

I'll track down what exactly the error is later this morning, I hope.

mwcraig: > You say you expect the different is BIGGER than the expected_deviation, but below (line 556) you assert it is smaller, no?

oops, now fixed

mwcraig:
Good news, there is no bug. The reason the test in question doesn't find sources when `use_coordinates="pixel"` is that the pixel coordinates are set by the position of sources in the first image, but the rest of the test images are deliberately offset from those. I added something to that effect in the comments.

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -514,14 +535,34 @@ def test_photometry_on_directory():
...
+            # Here we just check whether any difference is consistent with
+            # less than the expected one sigma deviation.
+            assert np.abs(expected_flux - obs_avg_net_cnts) < expected_deviation
+        else:
+            # In this case we are trying to do photometry in pixel coordinates,
+            # using the pixel location of the sources as found in the first image --
+            # see the line where found_sources is defined.
+            #
+            # However, the images are shifted with respect to each other by
+            # list_of_fakes, so there are no long stars at those positions in the
```
JuanCab, line 557: "no long" --> "no longer"?

JuanCab (approved):

stellarphot/photometry/tests/test_photometry.py
mwcraig: 🤣 I wish we could take care of collimation issues in the software! Will fix in the next batch of changes.

PR #275, opened 2024-02-16
This rearranges some of the photometry settings and adds an `AperturePhotometry` class

mwcraig:
There are still a couple things to be resolved:

+ Should single/multi image photometry become methods of `AperturePhotometry`?
+ Is `AperturePhotometry` even needed?

Marking this as a draft for the moment, but if you get a chance to take a peek, @JuanCab, feedback would be welcome. WOn't get back to it myself until Sunday.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,8 +1,70 @@
 Documentation
 =============
 
-This is the documentation for stellarphot.
-This is a package for doing stellar photometry that relies on astropy.
+Stellarphot is a package for performing photometry on astronomical images. It
+provides a simple interface for performing aperture photometry of either a single
+image or a directory with multiple images.
+
+
+Getting Started
+===============
+
+Installation
+------------
+
+Install it with pip or conda.
+
+Overview
+--------
+
+Using ``stellarphot`` starts with defining a bunch of configuration settings. Some of these,
+like details about your observing location and camera, will change infrequently.
+Others, like what size apertures to use for photometry and where in each image those
+apertures should be placed, may change from night to night.
+
+Provide your observatory information
+-------------------------------------
+
+TBD
+
+Provide your camera information
+-------------------------------
+
+TBD
+
+Provide a source list
+---------------------
+
+TBD
+
+Some optional settings
+-----------------------
+
+TBD
+
+Performing photometry
+---------------------
+
+Once you have made your settings doing photometry is a two line process. First, you
+create a photometry object:
+
+```python
+from stellarphot.photometry import AperurePhotometry
+phot = AperurePhotometry(photometry_settings)
+```
+
+Then you can perform photometry on a single image:
+
+```python
+phot(image)
+```
+
+If you have a directory of images you can perform photometry on all of them at once like this:
+
+```python
+phot(directory, object_of_interest="M13")
+```
+
 
 .. toctree::
   :maxdepth: 3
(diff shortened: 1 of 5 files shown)


docs/index.rst
JuanCab, line 1: Obviously this will need some more fleshing out, but I like that you are getting the documentation in place.  You may want to mention for situations where the users may not be comfortable yet with the command line or Python programming, a Jupyter-notebook-based set of tools is available to configure stellarphot and run the photometry.

Also, do we want to note this is meant to be used on REDUCED images (or am I wrong here)?  I don't remember any processing of flats/darks/etc, which I believe is because we were working with reduced images.

stellarphot/photometry/photometry.py
```diff
@@ -39,14 +42,72 @@
...
+    ) -> PhotometryData:
+        """
+        Perform aperture photometry on a single image or a directory of images.
+
+        Parameters
+        ----------
+        file_or_directory : str or Path
+            The file or directory on which to perform aperture photometry.
+
+        fname : str, optional (Default: None)
```
JuanCab, line 64: Isn't `fname` redundant if the value of `file_or_directory` already automatically switches whether it is single image processing or multi-image processing based on it being a file or directory.  Do you want to support wildcard (e.g. glob) processing of filename?

stellarphot/settings/models.py
```diff
@@ -68,6 +72,9 @@ class BaseModelWithTableRep(BaseModel):
     Class to add to a pydantic model YAML serialization to an Astropy table.
     """
 
+    # NOTE WELL that this will set the configuration for all subclasses of this
```
JuanCab, line 75: OK, Not sure what the point of "NOTE WELL" is here.  Do you really want me to pay attention that this sets the configuration of all subclasses of `BaseModelWithTableRep`?

stellarphot/settings/models.py
```diff
@@ -412,13 +441,25 @@ def earth_location(self):
         )
 
 
-class PhotometryOptions(BaseModelWithTableRep):
+class SourceLocationSettings(BaseModelWithTableRep):
```
JuanCab, line 444: Good name choice.

stellarphot/settings/models.py
JuanCab, line 1: I like the organization of the settings you have put together.

JuanCab (commented):
So far looks like a strong first step re-organizing the settings to be more clearly delineated and also to work toward the object-oriented model of photometry we talked about.  Not "approving" because it is a draft, but looks like a lot of good stuff here.

mwcraig:
Next up will be better documentation and fleshing out the couple of widgets that still need to be done.

mwcraig:
@JuanCab  what do you think about these two questions:

> + Should single/multi image photometry become methods of AperturePhotometry?
> + Is AperturePhotometry even needed?

stellarphot/settings/models.py
JuanCab: Ah

JuanCab:
> @JuanCab what do you think about these two questions:
> 
> > * Should single/multi image photometry become methods of AperturePhotometry?
> > * Is AperturePhotometry even needed?

Unless you somehow want to leave the possibility open for other kinds of photometry to be done as single or multi-image, it seems to make sense to me to keep `AperturePhotometry` as descriptive of the task it handles.  I could see that just to avoid being overly long with the function names which might call them `single_image` and `multi_image` since the 'photometry' would be redundant if they are in the `AperturePhotometry` class.

JuanCab (approved):
I had to use Sublime Merge to see the changes since GitHub wasn't as clear about what changed since the last update.  But the changes all look reasonable.

PR #278, opened 2024-02-23
UI work

mwcraig:
@JuanCab -- if this passes tests I'm going to merge it so we can have a couple folks try out some of the settings forms.

### DIFF
--- a/autoui_widget_viewer.ipynb
+++ b/autoui_widget_viewer.ipynb
@@ -14,6 +14,17 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "id": "6ecbdeea-e2f0-46f6-bcc4-33edc5c7e2a4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from ipyautoui import AutoUi\n",
+    "import ipywidgets as ipw"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "id": "b291300e-1bb2-4f00-96c7-4236286d0c2e",
    "metadata": {},
    "outputs": [],
@@ -25,8 +36,10 @@
     "    PhotometryApertures, \n",
     "    Exoplanet, \n",
     "    PhotometrySettings, \n",
-    "    PassbandMap\n",
+    "    PassbandMap,\n",
     ")\n",
+    "from stellarphot.settings.models import _extract_short_description\n",
+    "from stellarphot.settings import ui_generator\n",
     "from pydantic import BaseModel"
    ]
   },
@@ -39,7 +52,12 @@
    "source": [
     "classes = [Camera, Observatory, PhotometryApertures, LoggingSettings, PassbandMap, Exoplanet, PhotometrySettings]\n",
     "\n",
-    "uis = [AutoUi(c) for c in classes]"
+    "uis = [ui_generator(c) for c in classes]\n",
+    "\n",
+    "# for ui in uis:\n",
+    "#     ui.show_validation = False\n",
+    "#     ui.show_null = True\n",
+    "#     ui.description = _extract_short_description(ui.description)"
    ]
   },
   {
@@ -53,6 +71,47 @@
     "tabs.children = uis\n",
     "tabs.titles = [c.__name__ for c in classes]\n",
     "tabs"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "51403657-ca4e-4615-bc59-65b93e3fad6d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ui = uis[-1]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1c6610f7-1973-4a9b-b755-49bbacac7e78",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ui.value"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4292410c-206d-4a1b-a30c-b227f22517d7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ui_generator(PhotometrySettings)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b00ce1bf-6406-495c-94eb-9eae16b9ded3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ui.show_validation = False\n",
+    "ui.show_null = True"
    ]
   },
   {
@@ -156,6 +215,50 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "83b25581-c0a1-44e7-8d8f-20d7d756836d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from typing import Annotated\n",
+    "from pydantic import BaseModel, Field\n",
+    "\n",
+    "from ipyautoui import AutoUi\n",
+    "\n",
+    "class NestedStr(BaseModel):\n",
+    "    \"\"\"A nested model\"\"\"\n",
+    "    string: str\n",
+    "\n",
+    "class NestedFloat(BaseModel):\n",
+    "    \"\"\"Nested float\"\"\"\n",
+    "    number: float | None\n",
+    "\n",
+    "class Compound(BaseModel):\n",
+    "    \"\"\"My compound model\"\"\"\n",
+    "    nest_str_1: Annotated[\n",
+    "        NestedStr,\n",
+    "        Field(description=\"Overridden description nest_str_1\")\n",
+    "    ]\n",
+    "    nest_str_2: NestedStr = Field(description=\"Non-annotated overridden description nest_str_2\")\n",
+    "    nest_str_3: NestedStr = Field(title=\"NestedStr\", description=\"Overridden, and title doesn't work nest_str_3\")\n",
+    "    nest_str_4: NestedStr = Field(title=\"Nested Str\", description=\"Overridden, this title works nest_str_4\")\n",
+    "    nest_float: NestedFloat\n",
+    "\n",
+    "AutoUi(Compound)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "53599a47-ee36-466a-807f-8f8e5c1c7a7d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "AutoUi(Compound)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c2cdc0a5-7788-49a8-974e-e2a5891c2c53",
    "metadata": {},
    "outputs": [],
    "source": []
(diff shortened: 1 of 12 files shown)


mwcraig:
Actually, just kidding about the merging, there is a better way to handle it. In any event, another bug needs to get fixed first.

mwcraig:
@JuanCab -- the thing to focus on here is the new `PassbandMap` model. The old one (a stright-up `dict`) didn't have an automatically generate-able UI. The new one does. On the other hand, the new is a PITA to work with, so I made it settable from a `dict`, and added just enough methods to make it behave like a `dict`. Not at all sure this is the best way to go, but we can always change it later I suppose.

I'll get folks looking at the UI separately from getting this merged.

autoui_widget_viewer.ipynb
JuanCab, line 1: Did note review notebook beyond GitHub changes interface, looks like mostly markdown changes anyway.

docs/index.rst
```diff
@@ -13,7 +13,27 @@ Getting Started
...
+Testing
++++++++
+
+If you are testing a pre-release version of stellarphot we recommend setting up
+a virtual environment and installing stellarphot in this environment.
+
+Creating an environment with `conda` or `mamba`:
+
+```bash
+mamba create -n stellarphot-test python=3.11
```
JuanCab, line 25: So does stellarphot-test have all the dependencies defined?  I have my own testing environment for stellarphot under conda and I should probably compare the two setups.

stellarphot/core.py
```diff
@@ -171,9 +171,14 @@ def _update_passbands(self):
         # in case any of the new names are longer than the longest of the old names.
         # If that happens, astropy by default just truncates the names.
         new_filter_name = [
-            self._passband_map[orig_pb] if orig_pb in self._passband_map else orig_pb
+            (
+                self._passband_map[orig_pb]
+                if orig_pb in self._passband_map.keys()
```
JuanCab, line 176: Isn't the default iterator for a dictionary keys?  Oh, I see, you defined a method keys for the PassbandModel that returns the keys.  Later I do see where you perform `self._passband_map = passband_map.model_copy()` to set `self._passband_map`... but `passband_map` is a `dict` according to the docstrings for`CatalogData` and `PhotometryData` . Does this need to get changed to `PassbandMap` object?

Did you place this function the `BaseEnhancedTable` so it is available to be inherited by `CatalogData` and `PhotometryData`?  I ask because the passband mapping is not used in the `BaseEnhancedTable`.

Finally, what is `self["passband"]`?   I can't find where `self["passband"]` is even defined in `BaseEnhancedTable`?

stellarphot/settings/models.py
```diff
@@ -545,27 +564,81 @@ class PhotometryOptions(BaseModelWithTableRep):
     fwhm_by_fit: bool = True
 
 
+class PassbandMapEntry(BaseModel):
+    """
+    long loing liong liong liong long long
```
JuanCab, line 569: What is this?

stellarphot/core.py
```diff
@@ -541,7 +546,7 @@ def __init__(
 
             # Apply the filter/passband name update
             if passband_map is not None:
-                self._passband_map = passband_map.copy()
+                self._passband_map = passband_map.model_copy()
```
JuanCab, line 549: if `passband_map` is the `dict` passed in as a parameter, how can we apply `model_copy()` which I though was from pydantic?  Since this file is not importing the `PassbandMap` model from `.settings/models.py` I am having a hard time seeing how that model gets applied.

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -87,10 +88,12 @@
 )
 
 # Passband map for the tests
-PASSBAND_MAP = {
-    "B": "B",
-    "rp": "SR",
-}
+PASSBAND_MAP = PassbandMap(
```
JuanCab, line 91: So this is using the pydantic model... but stellarphot/core wasn't?

stellarphot/settings/models.py
```diff
@@ -545,27 +564,81 @@ class PhotometryOptions(BaseModelWithTableRep):
...
     """
     Class to represent a mapping from one set of filter names to another.
 
     Parameters
     ----------
-    yours_to_aavso : dict[str, str]
-        A dictionary containing instrumental passband names as keys and
-        AAVSO passband names as values. This is used to rename the passband
-        entries in the output photometry table.
+    your_filter_names_to_aavso : list[`stellarphot.settings.PassbandMapEntry`]
```
JuanCab, line 592: OK, this makes sense, although again, not seeing how `stellarphot/core.py` knows about this?

stellarphot/settings/models.py
```diff
@@ -52,6 +54,21 @@
 )
 
 
+def _extract_short_description(docstring: str) -> str:
```
JuanCab, line 57: Clever way to keep up with the docstring changes. :)

JuanCab (commented):
I *think* I followed this in the end, but would appreciate answers to some of my questions.  You need to fix all the mentions of `passband_map` in the docstrings to indicate it is now a `PassbandMap` object!  Otherwise, while ugly, if it passes the tests, then we can say it works well enough.

docs/index.rst
mwcraig: No -- this section should be modified/removed. For testing of this branch I sent the instructions below. It creates a bare-bones environment then `pip` installs everything from a github URL so people can try out the UI independent of whether we've done a pre-release or not and dodging all of the intricacies of git:

---
This first look will be to try out a couple of jupyter notebook-based forms for entering a bunch of settings.

I’d recommend, for now, that when you are testing you make a python environment for testing in, the delete the environment when you are done instead of trying to frequently update stellarphot. Instructions for doing that if you use conda are below, happy to provide non-conda instructions too.

To make an environment for testing called sp-test:

```
conda create -n sp-test python=3.11
```

When the environment is created, activate it:

```
conda ativaste sp-test
```

The command to install the version of stellarphot to try this time around is:

```
pip install git+https://github.com/mwcraig/stellarphot.git@ui-work
```

When this is done, please save the attached notebook to a folder, and navigate into that folder in the terminal. Run “`jupyter lab`” to start up jupyter and try going through the notebook. After several cells of set up it will present forms for you to enter settings. Don’t worry about using realistic values for now.

I’m expecting you will have questions and that it needs more instructions, but figured getting a quick look sooner was better than waiting longer.

When you are done you can deactivate the environment and then remove it:

```
conda deactivate
conda remove --all --yes sp-test
```

stellarphot/settings/models.py
mwcraig: oops, meant to edit that 😬 it was keyboard mashing.

stellarphot/settings/models.py
mwcraig: Yeah, also avoids `ipyautoui` using _the entire docstring_ as the brief description...

stellarphot/core.py
mwcraig: Taking a look in a moment -- the intent was to have every place use a `PassbandMap` object but to make that object act sort of like a dict.

mwcraig: Also, you are write that the default iterator is over keys, but I wasn't sure which dunder method I needed to use to make that happen. Can dig it up in _Fluent Python_ probably.

mwcraig: > Did you place this function the BaseEnhancedTable so it is available to be inherited by CatalogData and PhotometryData? I ask because the passband mapping is not used in the BaseEnhancedTable.

I think `_update_passbands` has been in `BaseEnhancedTable` for a while: https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/core.py#L166

I do think you are correct about the reason it is in `BaseEnhancedTable`

mwcraig: > Finally, what is self["passband"]?

Took me a sec to remember, it is one of the columns in the table.

mwcraig:
@JuanCab -- I think this is ready for another look

stellarphot/core.py
```diff
@@ -284,7 +285,7 @@ class PhotometryData(BaseEnhancedTable):
         names as values.  This is used to automatically update the column
         names to the desired names before the validation is performed.
 
-    passband_map: dict, optional (Default: None)
+    passband_map: `stellarphot.settings.PassbandMap`, optional (Default: None)
         A dictionary containing instrumental passband names as keys and
```
JuanCab, line 289: Do you want to call it a `dictionary` in the documentation?  I guess It behaves like a dictionary in many ways, but you need to create it in a very different way, am I not correct?  What happens if they just pass in a `dict`?  It should fail.  Maybe we add an Exception that reports they passed in a dict and tells them how to convert it into a proper PassbandMap object?

stellarphot/core.py
```diff
@@ -603,7 +604,7 @@ class CatalogData(BaseEnhancedTable):
         names as values.  This is used to automatically update the column
         names to the desired names BEFORE the validation is performed.
 
-    passband_map: dict, optional (Default: None)
+    passband_map: `stellarphot.settings.PassbandMap`, optional (Default: None)
         A dictionary containing instrumental passband names as keys and
```
JuanCab, line 608: Same quibble with calling it a dictionary as above.

stellarphot/io/tests/test_tess_submission.py
```diff
@@ -88,6 +88,10 @@ def test_target_file():
     except ConnectionError:
         server_down = True
         tess_target = None  # Assure tess_target is defined so that we can delete it
+    except ValueError:
+        # The server is teechnically back but producing garbage....
```
JuanCab, line 92: "technically" not "teechnically"...

stellarphot/settings/models.py
```diff
@@ -481,12 +498,14 @@ class SourceLocationSettings(BaseModelWithTableRep):
...
-    source_list_file: str
+    source_list_file: Annotated[
+        str, Field(json_schema_extra=dict(autoui="ipyautoui.custom.FileChooser"))
+    ]
     use_coordinates: Literal["sky", "pixel"] = "sky"
     shift_tolerance: NonNegativeFloat = 5.0
 
 
-class PhotometryOptions(BaseModelWithTableRep):
+class PhotometryOptionalSettings(BaseModelWithTableRep):
```
JuanCab, line 508: I agree, better name.

stellarphot/settings/models.py
JuanCab, line 1: I like how you added support to make PassbandMap behave like a dict in many ways.

JuanCab (commented):
My only issue is in the docs of `stellarphot/core.py` you refer to PassbandMap as a dictionary.  I do think you have a point in that it certainly behaves like a dictionary, BUT it isn't created like `dict`... don't know if that will confuse people who actually read the docstrings but are not familiar with pydantic.

stellarphot/core.py
mwcraig: Oops, that was completely an oversight on my part. I'll change the wording here and below.

stellarphot/settings/models.py
mwcraig: That was partly bcc it was easier to do that than it was to rewrite code that used `PassbandMap` 🤣

mwcraig:
@JuanCab -- made the changes you suggested. If test pass I think this can be merged/

JuanCab (approved):
OK, looks good.

PR #280, opened 2024-02-23
Handle long passband names and test for it

### LINKED ISSUE #279: Creating `PhotometryData` fails if the pasbband_map new passband names are longer than the longest of the original passband names
Consider this setup:

+ The input data to `PhotometryData` has a `passband` column whose contents are `['i']`.
+ The 1passband_map` is `{'i': 'SI'}`

This will issue a warning and fail to update the filter value because the original dtype of the column is based on the length of the longest string in the column, in this case `U1` so unicode, 1 character. Settings the value to a longer string, e.g. `SI`, results in truncation of the value to one character, in this case setting the value to `S`.

The warning is:

```
E           astropy.table.column.StringTruncateWarning: truncated right side string(s) longer than 1 character(s) during assignment
```

Test for this and fix is incoming.

mwcraig:
This fixes #279. I checked that the tests failed before I made the fixes....

@JuanCab if you are able to look over this one by monday that would be great.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -166,9 +166,15 @@ class BaseEnhancedTable(QTable):
     def _update_passbands(self):
         # Converts filter names in filter column to AAVSO standard names
         # Assumes _passband_map is in namespace.
-        for orig_pb, aavso_pb in self._passband_map.items():
-            mask = self["passband"] == orig_pb
-            self["passband"][mask] = aavso_pb
+
+        # Create a list of new passband names instead of trying to change names in place
+        # in case any of the new names are longer than the longest of the old names.
+        # If that happens, astropy by default just truncates the names.
+        new_filter_name = [
+            self._passband_map[orig_pb] if orig_pb in self._passband_map else orig_pb
+            for orig_pb in self["passband"]
+        ]
+        self["passband"] = new_filter_name
 
     def clean(self, remove_rows_with_mask=False, **other_restrictions):
         """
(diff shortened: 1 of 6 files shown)


stellarphot/core.py
```diff
@@ -166,9 +166,15 @@ def _update_colnames(self, colname_map, data):
     def _update_passbands(self):
         # Converts filter names in filter column to AAVSO standard names
         # Assumes _passband_map is in namespace.
-        for orig_pb, aavso_pb in self._passband_map.items():
-            mask = self["passband"] == orig_pb
-            self["passband"][mask] = aavso_pb
+
+        # Create a list of new passband names instead of trying to change names in place
```
JuanCab, line 170: Took me a bit of effort to comprehend the list comprehension... but I think it is sound.

stellarphot/core.py
```diff
@@ -166,9 +166,15 @@ def _update_colnames(self, colname_map, data):
...
+        # in case any of the new names are longer than the longest of the old names.
+        # If that happens, astropy by default just truncates the names.
```
JuanCab, line 172: I assume you mean that the truncation is a bad thing...

Also, took me a bit of effort to comprehend the list comprehension... but I think it is sound assuming what you want is for all the `orig_pb` to be kept, using the existing `_passband_map` assigned to `orig_pb` if there is one, otherwise using `orig_pb` itself (I assume instead of letting astropy truncate the old name).

JuanCab (dismissed):
Assuming the test passes, it looks good to me.

mwcraig:
Fixed the tests -- RTD seemed to be failing because of a partially downloaded version of a dependency, and one of the remote data tests because a server at the university of loiusville is down for maintenance. Merging, since both have workarounds for now.

PR #284, opened 2024-03-11
Fix observatory widget

### LINKED ISSUE #281: Observatory input form is unusable
The issue comes up in this form after a valid lat/lon/elevation have been entered. When that is done ipyautoui by default serilaizes the object and puts those values in the form. In particular, lat/lon are converted to dms format regardless of how they were entered.

That is not the end of the world.

But when you try to edit those values, at every keystroke ipyautoui tries to update the value as you are typing it, which is not helpful and makes it impossible to edit.

I think a short term fix is to set `continuous_update` to false on each of the entry widgets.

mwcraig:
This fixes #281 b y disabling `continuous_update` for most entry widgets.

### DIFF
--- a/autoui_widget_viewer.ipynb
+++ b/autoui_widget_viewer.ipynb
@@ -39,7 +39,7 @@
     "    PhotometryApertures, \n",
     "    Exoplanet, \n",
     "    PhotometrySettings, \n",
-    "    PhotometryOptions,\n",
+    "    PhotometryOptionalSettings,\n",
     "    PassbandMap,\n",
     "    SourceLocationSettings,\n",
     ")\n",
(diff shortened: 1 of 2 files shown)


JuanCab (approved):
The changes seem to be mostly documentation and for the auto_widget_viewer.ipynb notebook just changing the name of the PhotometyOptions to PhotometryOptionalSettings (the new name for those settings).  Seems completely reasonable stuff.

mwcraig:
This is really the only functional bit:

```python
    for widget in ui.di_widgets.values():
        if hasattr(widget, "continuous_update"):
            widget.continuous_update = False
```

It prevents the AutoUi widget from trying to force the string representation of values into the widget as you are trying to type. That was really annoying and made it essentially impossible to edit a value like latitude that had correct units but an incorrect value.

PR #285, opened 2024-03-12
Add descriptions to pydantic fields for display in UI

### LINKED ISSUE #283: Add some description of what each of the `PhotometryOptionalSettings` means
Right now it is not obvious what they mean

mwcraig:
This fixes #283 by adding a brief description to each of the optional settings. It also adds descriptions for fields in the rest of the models too. In a couple of cases (e.g. observatory latitude, longitude and elevation) it adds examples of valid values.

### DIFF
--- a/stellarphot/settings/models.py
+++ b/stellarphot/settings/models.py
@@ -336,15 +336,29 @@ class PhotometryApertures(BaseModelWithTableRep):
 
     radius: Annotated[
         PositiveInt,
-        Field(default=1, json_schema_extra=dict(autoui="ipywidgets.BoundedIntText")),
+        Field(
+            default=1,
+            description="radius of circular annulus, in pixels",
+            json_schema_extra=dict(autoui="ipywidgets.BoundedIntText"),
+        ),
     ]
     gap: Annotated[
         PositiveInt,
-        Field(default=1, json_schema_extra=dict(autoui="ipywidgets.BoundedIntText")),
+        Field(
+            default=1,
+            description="Size of gap between aperture and annulus, in pixels",
+            json_schema_extra=dict(autoui="ipywidgets.BoundedIntText"),
+        ),
     ]
     annulus_width: Annotated[
         PositiveInt,
-        Field(default=1, json_schema_extra=dict(autoui="ipywidgets.BoundedIntText")),
+        Field(
+            default=1,
+            description=(
+                "distance between inner and outer radii of annulus, " "in pixels"
+            ),
+            json_schema_extra=dict(autoui="ipywidgets.BoundedIntText"),
+        ),
     ]
     # Disable the UI element by default because it is often calculate from an image
     fwhm: Annotated[PositiveFloat, Field(disabled=True, default=1.0)]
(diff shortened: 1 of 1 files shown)


Tanner728 (approved):
Everything looks good to me. Can the `elevation` also accept imperial units or is it only metric? I do not think most people will try to use imperial units but who knows.

mwcraig:
Imperial is only allowed if they have separately activated it, and I wouldn't guarantee it will work. That was why I made all the examples metric.

stellarphot/settings/models.py
```diff
@@ -336,15 +336,29 @@ class PhotometryApertures(BaseModelWithTableRep):
...
+            json_schema_extra=dict(autoui="ipywidgets.BoundedIntText"),
+        ),
     ]
     annulus_width: Annotated[
         PositiveInt,
-        Field(default=1, json_schema_extra=dict(autoui="ipywidgets.BoundedIntText")),
+        Field(
+            default=1,
+            description=(
+                "distance between inner and outer radii of annulus, " "in pixels"
```
JuanCab, line 358: This line seems malformed, It should work since the parenthesis around the strings effectively collects them together, but WHY was this done here?  The line is not as far as I can see to long to fit within 80 characters.

stellarphot/settings/models.py
```diff
@@ -435,18 +449,50 @@ class Observatory(BaseModelWithTableRep):
...
+        ),
     ]
     longitude: Annotated[
         Longitude,
         _UnitQuantTypePydanticAnnotation,
         BeforeValidator(add_degree_to_float),
+        Field(
+            description="Longitude of the observatory",
+            examples=[
+                "-96.7678",
```
JuanCab, line 474: I know it is a little silly, but given the target audience includes new users, should we clarify negative is used for W longitude and the 360 version is always E of Greenwich?  It would make the description too long, so maybe we put this in the documentation or in the Python notebooks themselves?

stellarphot/settings/models.py
```diff
@@ -435,18 +449,50 @@ class Observatory(BaseModelWithTableRep):
...
+                "263d13m55.92s",
+            ],
+        ),
+    ]
+    elevation: Annotated[
+        QuantityType,
+        WithPhysicalType("length"),
+        Field(
+            description="Elevation of the observatory",
+            examples=["1000 m", "1 km", "6.685e-9 au"],
```
JuanCab, line 487: I like the inclusion of AU, but given this is a *stellar* photometry package maybe pc instead. :)

stellarphot/settings/models.py
```diff
@@ -435,18 +449,50 @@ class Observatory(BaseModelWithTableRep):
...
+        ),
     ]
-    elevation: Annotated[QuantityType, WithPhysicalType("length")]
-    AAVSO_code: str | None = None
-    TESS_telescope_code: str | None = None
+    AAVSO_code: Annotated[str | None, Field(description="AAVSO code for observer")] = (
+        None
+    )
+    TESS_telescope_code: Annotated[
+        str | None, Field(description="AAVSO code for observer")
```
JuanCab, line 494: This description appears to be an accidental duplicate of the AAVSO_code description.

stellarphot/settings/models.py
```diff
@@ -499,10 +545,28 @@ class SourceLocationSettings(BaseModelWithTableRep):
...
+                autoui="ipyautoui.custom.FileChooser",
+                filter_pattern=["*.ecsv", "*.csv"],
+            )
+        ),
     ]
-    use_coordinates: Literal["sky", "pixel"] = "sky"
-    shift_tolerance: NonNegativeFloat = 5.0
+    use_coordinates: Annotated[
+        Literal["sky", "pixel"],
+        Field(description="coordinates to use for locating sources"),
```
JuanCab, line 558: Do we need to add "in images" after "locating sources"?  Probably just me being overly specific.

stellarphot/settings/models.py
```diff
@@ -558,10 +622,42 @@ class PhotometryOptionalSettings(BaseModelWithTableRep):
...
+                "Should the digitization noise be included in the calculation of "
+                "the noise for each observation?"
+            )
+        ),
+    ] = True
+
+    reject_too_close: Annotated[
+        bool,
+        Field(
+            description="Should sources that are too close be excluded from photometry?"
```
JuanCab, line 638: I would say "too close together"... but again, that's me. On first read I imagined (OK, so excluding the moon and orbiting satellites...

stellarphot/settings/models.py
```diff
@@ -558,10 +622,42 @@ class PhotometryOptionalSettings(BaseModelWithTableRep):
...
+        ),
+    ] = True
+
+    reject_background_outliers: Annotated[
+        bool,
+        Field(
+            description=(
+                "Should extreme pixels (e.g. from a star in the annulus) "
```
JuanCab, line 646: For readability, I would move the parenthetical to the end of the sentence and before the question mark.

JuanCab (dismissed):
I really like this PR... making the GUI more usable with built-in help is excellent!  A lot of nit-picking on my part, but happy this is being done.

stellarphot/settings/models.py
mwcraig: Good catch, it had originally been longer

stellarphot/settings/models.py
mwcraig: I'm thinking an example. in the documentation -- for now I'll add a bit of discussion to the `Examples` section of the `Observatory` docstring.

stellarphot/settings/models.py
mwcraig: 😮 

oops

stellarphot/settings/models.py
mwcraig: Yes, I think that is helpful

stellarphot/settings/models.py
```diff
@@ -433,20 +453,70 @@ class Observatory(BaseModelWithTableRep):
...
     ]
-    elevation: Annotated[QuantityType, WithPhysicalType("length")]
-    AAVSO_code: str | None = None
-    TESS_telescope_code: str | None = None
+    elevation: Annotated[
+        QuantityType,
+        WithPhysicalType("length"),
+        Field(
+            description="Elevation of the observatory",
+            examples=["1000 m", "1 km", "3.241e-14 pc"],
```
JuanCab, line 511: Nice fix. :)

JuanCab (approved):
Looks good!

PR #288, opened 2024-03-25
Add method argument to photometry options

### LINKED ISSUE #273: single/multi image photometry should accept a "method" argument that gets passed into photutils
Right now the method is [hardwired to `"center"`](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/photometry/photometry.py#L409). This will get fixed after #272 is in.

mwcraig:
This fixes #273 by adding a `method` field to `PhotometryOptionalSettings`. The method determines how photutils handles partial pixels. 

The old default in stellarphot used to be "center", but "exact" is arguably the better choice so I changed it.

### DIFF
--- a/stellarphot/photometry/tests/test_photometry.py
+++ b/stellarphot/photometry/tests/test_photometry.py
@@ -48,6 +48,8 @@ DEFAULT_SOURCE_LOCATIONS = SourceLocationSettings(
 )
 
 PHOTOMETRY_OPTIONS = PhotometryOptionalSettings()
+# This used to be the default; it has switched to exact
+PHOTOMETRY_OPTIONS.method = "center"
 
 # A camera with not unreasonable settings
 FAKE_CAMERA = Camera(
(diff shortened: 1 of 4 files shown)


stellarphot/photometry/tests/test_photometry.py
JuanCab, line 1: Test seems reasonable although it might be be nice to construct a test where the "exact" value was known to be greater/less than "center" so we could be a little more precise... but I can't think of an obvious way to predict this for a circular aperture... its probably not worth the time.

stellarphot/settings/models.py
```diff
@@ -617,6 +617,15 @@ class PhotometryOptionalSettings(BaseModelWithTableRep):
         the star. If ``False``, the FWHM will be calculated by finding the
         second order moments of the light distribution. Default is ``True``.
 
+    method : `typing.Literal["exact", "center", "subpixel"]`, optional
```
JuanCab, line 620: If this is optional, I think you want to state 'exact' is the default if it is unset as part of the documentation.

JuanCab (approved):

mwcraig:
Assuming tests pass I'll go ahead and merge this since the only change was to add the default value to the documentation.

mwcraig:
Not sure what is up with RTD this morning, but merging since the previous commit built ok.

PR #290, opened 2024-03-31
Add a system for saving some settings for easier reuse

mwcraig:
The motivation of this pull request is to add a system for saving some settings -- those with a ``name`` field -- in a way that makes reuse easier.

The main additions are in the file `stellarphot/settings/saved_settings.py`, in particular the `SavedSettings` class.  I also fleshed out some of the settings code a little more.

This does add a new dependency, on (`platformdirs`)[https://github.com/platformdirs/platformdirs] which handles figuring out where user data should be stored on each platform. On Mac this ends up being inside `~/Library/Application Support/stellarphot/2`, where `2` refers the the version number of the code.

Once this in I'll open a PR with some notebooks that use this to make re-using settings easier. The idea is that camera, observatory, and passband map will be represented by dropdowns with the current cameras, etc, and an option to create a new camera.

Not sure I'll work up a UI to manage the saved settings more extensively or not, but adding this resolves one of the things I'd been confused about: how to make it easy to re-use camera, observatory or passband map without having to do a bunch of copy/paste.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ pip-wheel-metadata
 .pytest_cache
 .tmp
 .ruff_cache
+.hypothesis
 
 # Sphinx
 docs/api
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -38,32 +38,7 @@ pip install --pre --upgrade stellarphot
 Overview
 --------
 
-Using ``stellarphot`` starts with defining a bunch of configuration settings. Some of these,
-like details about your observing location and camera, will change infrequently.
-Others, like what size apertures to use for photometry and where in each image those
-apertures should be placed, may change from night to night.
 
-All the settings can be made through a graphical interface, via the command line, or by editing
-an existing settings file.
-
-The settings are grouped into these categories:
-
-- Observatory
-- Camera
-- Passband Map
-- Photometry Apertures
-- Source Location Settings
-- Optional Settings
-- Logging Settings
-
-The first three categories each include a ``name`` property, which is used to identify the
-settings and provides a shortcut to re-using those settings in the future.
-
-A copy of the settings are stored in a file called `stellarphot_settings.json` in the working directory
-where you are using stellarphot. It is these settings that are used when you run the photometry.
-
-Settings can be generated using a jupyter notebook with a graphical interface, by using the command line,
-or by editing a settings file directly.
 
 Graphical interface for generating settings
 -------------------------------------------
@@ -88,32 +63,6 @@ Editing a settings file directly
 
 The settings file is a JSON file that can be edited in any text editor.
 
-
-Provide your observatory information
--------------------------------------
-
-.. autopydantic_field:: stellarphot.settings.Observatory.latitude
-    :field-show-constraints: False
-..     :model-show-config-summary: False
-..     :model-show-field-summary: False
-.. .. autopydantic_model:: stellarphot.settings.Observatory
-..     :model-show-config-summary: False
-..     :model-show-field-summary: False
-
-Provide your camera information
--------------------------------
-
-TBD
-
-Provide a source list
----------------------
-
-TBD
-
-Some optional settings
------------------------
-
-TBD
 
 Performing photometry
 ---------------------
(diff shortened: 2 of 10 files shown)


JuanCab:
> This does add a new dependency, on (`platformdirs`)[https://github.com/platformdirs/platformdirs] which handles figuring out where user data should be stored on each platform. On Mac this ends up being inside `~/Library/Application Support/stellarphot/2`, where `2` refers the the version number of the code.

We should (if you haven't already) probably add documentation stating where the preferences are stored on the three major platforms, just so people know.

.gitignore
```diff
@@ -20,6 +20,7 @@ pip-wheel-metadata
 .pytest_cache
 .tmp
 .ruff_cache
+.hypothesis
```
JuanCab, line 23: Does this mean you are adding hypothesis to the suite of tools used for testing?

docs/index.rst
```diff
@@ -38,33 +38,31 @@ pip install --pre --upgrade stellarphot
...
-TBD
+To generate settings using the command line, run the following command:
 
-Provide a source list
----------------------
+```bash
+stellarphot-settings
+```
+
+This will generate a settings file in the directory in which you run the command
```
JuanCab, line 58: Is this separate from the one you mentioned going in the OS-dependent preferences directory or does it over-ride it?  You discuss this in the settings.rst file, but didn't know if it made sense to distinguish it here or say to check the settings documentation for more details.

docs/index.rst
```diff
@@ -38,33 +38,31 @@ pip install --pre --upgrade stellarphot
...
-There is a graphical interface for making all of the settings below. They can also be set
-programmatically in Python or by editing a file in your favorite text editor.
 
-Provide your observatory information
--------------------------------------
+Graphical interface for generating settings
+-------------------------------------------
 
-TBD
+To generate settings using a graphical interface, start Jupyter lab. In the launcher will be a section called
```
JuanCab, line 46: Is it "Jupyter lab"?  The [documentation](https://docs.jupyter.org/en/latest/) suggests it is "JupyterLab"... So does [this YouTube video](https://www.youtube.com/watch?v=dQw4w9WgXcQ).

docs/stellarphot/settings.rst
```diff
@@ -0,0 +1,142 @@
...
+- `~stellarphot.settings.PhotometryApertures`
+- `~stellarphot.settings.SourceLocationSettings`
+- `~stellarphot.settings.OptionalSettings`
+- `~stellarphot.settings.LoggingSettings`
+
+
+A copy of the settings are stored in a file called `stellarphot_settings.json` in the working directory
+where you are using stellarphot. It is these settings that are used when you run the photometry.
+
+Settings can be generated using a jupyter notebook with a graphical interface, by using the command line,
```
JuanCab, line 29: I think "Jupyter" like "Python" should probably be capitalized throughout...

mwcraig:
> We should (if you haven't already) probably add documentation stating where the preferences are stored on the three major platforms, just so people know.

I think I documented (in the narrative documentation) that they can get the location from a `SavedSettings` object with the `settings_path` attribute. My preference would be to document it like that (i.e. provide a clear way to get the path) rather than having to maintain a mirror of what platformdirs says is the way.

docs/stellarphot/settings.rst
JuanCab, line 1: Nice first draft, very readable and clear.

stellarphot/settings/settings_files.py
```diff
@@ -0,0 +1,236 @@
...
+            del instance.as_dict[name]
+            instance.save()
+        else:
+            # Delete the entire file
+            file_path.unlink(missing_ok=True)
+
+
+class Cameras(SavedFileOperations, BaseModel):
+    # Using the ClassVar annotation means this is treated as a class variable rather
+    # than a pydantic field. We don't pydantic storing the name of the settings file in
```
JuanCab, line 82: We don't WANT pydantic storing... (missing word?)

JuanCab (changes requested):
Looks good but for a few typos.  The tests seem to cover everything I can think of that could go wrong with this scheme and then some.

.gitignore
mwcraig: No, but for some reason I keep getting a hypothesis directory when I run tests, so I wanted it to go away.

docs/index.rst
mwcraig: I think you are right...."JupyterLab"

PR #291, opened 2024-04-13
Impose restrictions on name field

mwcraig:
I realized while working on the form for entering `Camera` and other named objects that it allowed a bunch of bad names, like `""` or `"  "`. Since leading or trailing whitespace in a name also seemed like a bad idea, this PR adds (and tests) this constraint on the name field:

+ Must be one or more non-whitespace characters
+ Cannot have leading or trailing whitespace

A quick review on this would be helpful...

### DIFF
--- a/stellarphot/io/tess.py
+++ b/stellarphot/io/tess.py
@@ -474,7 +474,13 @@ class TessTargetFile:
         result = requests.get(
             self.aperture_server + "cgi-bin/gaia_to_aij/upload_request.cgi",
             params=params,
+            timeout=15,  # If no response in 15 seconds we won't ever get one...
         )
+        if result.status_code != 200:
+            raise requests.ConnectionError(
+                f"Failed to retrieve target file: {result.text}"
+            )
+
         links = re.search(
             'href="(.+)"',
             result.text.replace("\n", ""),
--- a/stellarphot/io/tests/test_tess_submission.py
+++ b/stellarphot/io/tests/test_tess_submission.py
@@ -3,7 +3,7 @@ import warnings
 
 import pytest
 from astropy.coordinates import SkyCoord
-from requests import ConnectionError
+from requests import ConnectionError, ReadTimeout
 
 from stellarphot.io.tess import TessSubmission, TessTargetFile
 
@@ -85,7 +85,7 @@ def test_target_file():
 
     try:
         tess_target = TessTargetFile(tic_742648307, magnitude=12, depth=10)
-    except ConnectionError:
+    except (ConnectionError, ReadTimeout):
         server_down = True
         tess_target = None  # Assure tess_target is defined so that we can delete it
     except ValueError:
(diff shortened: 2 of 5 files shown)


stellarphot/io/tess.py
```diff
@@ -474,7 +474,13 @@ def _retrieve_target_file(self):
         result = requests.get(
             self.aperture_server + "cgi-bin/gaia_to_aij/upload_request.cgi",
             params=params,
+            timeout=15,  # If no response in 15 seconds we won't ever get one...
         )
+        if result.status_code != 200:
```
JuanCab, line 479: I don't think this has to do with name field restrictions. :)

stellarphot/io/tests/test_tess_submission.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 from astropy.coordinates import SkyCoord
-from requests import ConnectionError
+from requests import ConnectionError, ReadTimeout
```
JuanCab, line 6: I'm guessing this has to do with allowing timeout of TESS access attempt, not the name field restrictions.

stellarphot/settings/models.py
```diff
@@ -54,6 +56,32 @@
...
+#
+# String should match pattern "^\\S$|^\\S.*\\S$"
+#
+# Instead, the custom validation function below checks for errors and
+# raises a ValueError with a more user-friendly message.
+def _non_empty_string_validator(value):
+    if not value.strip():
+        raise ValueError("name must not be empty or contain only whitespace.")
+
+    if not re.search(r"^\S$|^\S.*\S$", value):
```
JuanCab, line 73: So you are matching a SINGLE non-whitespace character as the string name or any string that starts and ends with non-whitespace characters?  I *guess* this makes sense, as I think of ways to avoid leading or trialing spaces... you have to allow for the possibility of a single character or multiple characters.  It does look odd, but the logic is sound.

stellarphot/settings/models.py
```diff
@@ -221,9 +249,10 @@ class Camera(BaseModelWithTableRep):
     model_config = MODEL_DEFAULT_CONFIGURATION
 
     name: Annotated[
-        str,
+        NonEmptyStr,
```
JuanCab, line 252: Took me a bit, but I see this is now using that NonEmptyStr validator you set up above...

JuanCab (approved):
I think this looks good and ready to merge, ALTHOUGH a few of the code changes in this commit have nothing to do with checking the name field for spaces and instead seem to be related to handling TESS catalog timeouts...

stellarphot/io/tess.py
mwcraig: Yeah, we were getting a test failure on two tests because of this

stellarphot/io/tests/test_tess_submission.py
mwcraig: Correct...I wanted to make the lights green

stellarphot/settings/models.py
mwcraig: Yes, that is the intent. I thought about just stripping out whitespace silently but it seemed but to flag it instead.

PR #300, opened 2024-04-28
Improve check of unit consistency in Camera

### LINKED ISSUE #299: `Camera` model validation code contains an error
In writing #298 I stumbled on what is, I think, an error in the validation logic for the `Camera` model.

Consider this code, which *should* be a valid `Camera` but is not:

```python
from stellarphot.settings import Camera
c = Camera(
     name="should_work",
      data_unit="electron",
      gain="1",
      dark_current="0.01 electron / second",
      read_noise="1.2 electron",
      pixel_scale="0.55 arcsec / pix",
      max_data_value="80000 electron"
 )
```

It fails because of the way that we check that the the `gain` has units `read_noise` / `data_unit` is this:

```python
            len(gain.unit.bases) != 2
            or gain.unit.bases[0] != rn_unit
            or gain.unit.bases[1] != self.data_unit
```

That fails here because `1` fails the first test.

Opening a PR shortly to fix this....

mwcraig:
This fixes #299 by checking that units of `gain` can be converted to units of `read_noise /data_unit` without requiring that the units of `gain` have a particular structure.

### DIFF
--- a/stellarphot/settings/models.py
+++ b/stellarphot/settings/models.py
@@ -7,7 +7,7 @@ from typing import Annotated, Any, Literal
 from astropy.coordinates import EarthLocation, Latitude, Longitude, SkyCoord
 from astropy.io.misc.yaml import AstropyDumper, AstropyLoader
 from astropy.time import Time
-from astropy.units import Quantity, Unit
+from astropy.units import Quantity, Unit, UnitConversionError
 from astropy.utils import lazyproperty
 from pydantic import (
     AfterValidator,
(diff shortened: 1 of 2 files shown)


stellarphot/settings/models.py
```diff
@@ -294,15 +294,14 @@ def validate_gain(self):
         # Check that gain and read noise have compatible units, that is that
         # gain is read noise per data unit.
         gain = self.gain
-        if (
-            len(gain.unit.bases) != 2
-            or gain.unit.bases[0] != rn_unit
-            or gain.unit.bases[1] != self.data_unit
-        ):
+
```
JuanCab, line 297: I have a vague memory that this may be my code you are fixing... good, it looks ugly now after seeing so much pydantic in your commits over the last few months. :). Do we want to change the dark current check a few lines below this to take a similar approach?

JuanCab (approved):
See comments, but otherwise, I approve.

stellarphot/settings/models.py
mwcraig: > Do we want to change the dark current check a few lines below this to take a similar approach?

Maybe...opening a separate issue for that

PR #306, opened 2024-04-29
Add better unicode test

### LINKED ISSUE #305: Some unicode names fail on Windows
In particular, Korean fails: 크레이그

Reported by @Tanner728

mwcraig:
It turns out that, on Windows, the default encoding used for text files is not UTF-8. Whatever the encoding is includes Greek characters, like π, that were used in the original test for unicode names, but not Korean characters as in 크레이그.

When complete, this pull request will close #305. The first commit here I am hoping to see a failure on Windows to make sure we are actually testing a case that fails.

### DIFF
--- a/stellarphot/settings/settings_files.py
+++ b/stellarphot/settings/settings_files.py
@@ -12,6 +12,8 @@ __all__ = ["SavedSettings", "SETTINGS_FILE_VERSION"]
 # or never.
 SETTINGS_FILE_VERSION = "2"  # value chosen to match amjor version of stellarphot
 
+ENCODING = "utf-8"
+
 
 class SavedFileOperations:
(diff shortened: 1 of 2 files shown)


mwcraig:
@Tanner728 -- would you ming trying this out? This should install the code in this PR:

`pip install git+https://github.com/mwcraig/stellarphot.git@fix-windows-unicode-issue`

If this works you will be able to load/save names with any unicode characters that UTF-8 understands (which ought to be most of them)

Tanner728 (approved):
After testing it appears that this fixed this issue. This time around I tested Arabic, Japanese, Russian, Korean, and Chinese. All of the languages tested worked for me on windows.

mwcraig:
Excellent! Can you please mark it as approved?

Tanner728 (approved):

PR #307, opened 2024-04-29
Fix "Edit" button issues

mwcraig:
This PR addresses the edit button issues mentioned in #303. The other issues will be addressed in a separate PR.

PR #308, opened 2024-04-29
Better handle disabling widgets when displaying

### LINKED ISSUE #303: Fix some issues with the custom widget UI
A few things that @Tanner728 found:

- [x] You can type stuff in nullable fields (e.g. AAVSO code for observatory) even if the "switch" is off if you have hit the "Edit" button
- [x] The "Status" button in the ipyautoui save button bar (the leftmost button) is clickable when editing and it shouldn't be.
- [x] "Edit" button appears if there are no items in a group (e.g. no observatories) and it shouldn't.
- [ ] ~~Might be possible to enter an invalid name (e.g. trailing space) in the form.~~
- [x] The edit button doesn't appear after making a new item.

mwcraig:
This addresses the first two items in #303, so when merged it fixes #303.

The original method of disabling and re-enabling the fields of a model when displaying vs editing or making a new item was a little too aggressive, and ended up enabling something elements that should not have been enabled. 

This fixes that by only disabling `AutoObject`s and also hiding the add/remove buttons that appear in the `PassbandMap` object.

### DIFF
--- a/stellarphot/settings/custom_widgets.py
+++ b/stellarphot/settings/custom_widgets.py
@@ -4,6 +4,7 @@
 import ipywidgets as ipw
 import traitlets as tr
 from ipyautoui.autoobject import AutoObject
+from ipyautoui.custom.iterable import ItemBox
 
 from stellarphot.settings import (
     Camera,
@@ -239,8 +240,16 @@ class ChooseOrMakeNew(ipw.VBox):
             State that ``disabled`` should be set to.
         """
 
-        if hasattr(top, "disabled") or isinstance(top, AutoObject):
+        if isinstance(top, AutoObject):
             top.disabled = value
+        elif isinstance(top, ItemBox):
+            if value:
+                # Disabled, so do not show the add/remove buttons
+                top.add_remove_controls = "none"
+            else:
+                # Enabled, so show the add/remove buttons
+                top.add_remove_controls = "add_remove"
+
         try:
             for child in top.children:
                 self._set_disable_state_nested_models(child, value)
--- a/stellarphot/settings/tests/test_custom_widgets.py
+++ b/stellarphot/settings/tests/test_custom_widgets.py
@@ -1,5 +1,6 @@
 import ipywidgets as ipw
 import pytest
+from ipyautoui.custom.iterable import ItemBox, ItemControl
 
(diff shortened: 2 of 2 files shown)


stellarphot/settings/custom_widgets.py
```diff
@@ -239,8 +240,16 @@ def _set_disable_state_nested_models(self, top, value):
             State that ``disabled`` should be set to.
         """
 
-        if hasattr(top, "disabled") or isinstance(top, AutoObject):
+        if isinstance(top, AutoObject):
             top.disabled = value
+        elif isinstance(top, ItemBox):
```
JuanCab, line 245: I'll admit I am trusting your tests because I don't know `pyautoui` well.

stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -187,6 +188,36 @@ def test_choosing_different_item_updates_display(self, tmp_path):
...
+        saved = SavedSettings(_testing_path=tmp_path)
+        passband_map = PassbandMap(**DEFAULT_PASSBAND_MAP)
+        saved.add_item(passband_map)
+
+        choose_or_make_new = ChooseOrMakeNew("passband_map", _testing_path=tmp_path)
+
+        # There is no great way to get to the ItemBox widget that contains and controls
+        # the add/remove buttons, so we keep going down through widget children until we
+        # get to an ItemBox and then check that the buttons are disabled.
+        # Recursion is the easiest way to do that, so recurse we will..
```
JuanCab, line 203: Wow, an actual use for recursion that isn't factorials or the Fibonacci sequence or binary search...

JuanCab (approved):
Looks good as tests go.  I am not an `pyautoui` expert, so I am relying on the tests as described working as commented.

Tanner728:
Doing some more testing here and noticed some odd interaction.
When creating a new passband map for the first time you are able to click the `Edit this Passband Map` button without having entered a passband name.

 This results in you not being able to enter a name for the passband map and therefore unable to save it. Is this something we need to address? Beyond this I did not find anymore bugs... yet. 

image

Tanner728 (approved):

mwcraig:
> When creating a new passband map for the first time you are able to click the Edit this Passband Map button without having entered a passband name.

Good catch -- I think that was fixed by #307 but will check once all of these are merged

PR #310, opened 2024-04-30
Connect state of save/revert buttons to something sensible

mwcraig:
I think the save and revert buttons should only be enabled when the value of the widget is valid (in the pydantic model sense) and there are unsaved changes.

This PR does that, and adds a test for it.

### DIFF
--- a/stellarphot/settings/tests/test_views.py
+++ b/stellarphot/settings/tests/test_views.py
@@ -1,0 +1,68 @@
+from stellarphot.settings import Camera, ui_generator
+from stellarphot.settings.tests.test_models import TEST_CAMERA_VALUES
+
+
+class TestUiGenerator:
+    def test_camera(self):
+        ui = ui_generator(Camera)
+        # The description should be the beginning of the docstring
+        assert Camera.__doc__.strip().startswith(ui.description.split()[0])
+
+        # We always want to show nullable fields
+        assert ui.show_null
+
+        # Which means we don't need the button to show/hide them
+        assert ui.bn_shownull.layout.display == "none"
+
+        # For now we do not show the validaation output because it is painfully
+        # verbose and not very helpful.
+        assert not ui.show_validation
+
+        # We always display nested models
+        assert ui.open_nested
+
+        # Finally, we don't want the widget to update continuously because it
+        # will overwrite the value the user entered.
+        for widget in ui.di_widgets.values():
+            if hasattr(widget, "continuous_update"):
+                assert not widget.continuous_update
+
+    def test_disabled_state_save_revert_button(self):
+        # We want the save and revert buttons to be enabled only when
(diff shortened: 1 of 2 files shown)


stellarphot/settings/views.py
```diff
@@ -43,4 +46,34 @@ def ui_generator(model):
...
+    """
+    Return a callback that will enable/disable the save and revert buttons based
+    on the validity of the value and whether there are unsaved changes.
+    """
+
+    def handler(_):
+        """
+        A handler must take an argument but we don't use it here.
+        """
+        needs_to_save = widget.is_valid and widget.savebuttonbar.unsaved_changes
```
JuanCab, line 76: Clever bit of logic here.

stellarphot/settings/tests/test_views.py
```diff
@@ -0,0 +1,68 @@
...
+        # The description should be the beginning of the docstring
+        assert Camera.__doc__.strip().startswith(ui.description.split()[0])
+
+        # We always want to show nullable fields
+        assert ui.show_null
+
+        # Which means we don't need the button to show/hide them
+        assert ui.bn_shownull.layout.display == "none"
+
+        # For now we do not show the validaation output because it is painfully
```
JuanCab, line 17: Typo of "validation", somewhat ironic since it passed initial review by the author....

stellarphot/settings/tests/test_views.py
```diff
@@ -0,0 +1,68 @@
...
+        # ...and the save button should still be disabled
+        assert ui.savebuttonbar.bn_save.disabled
+        assert ui.savebuttonbar.bn_revert.disabled
+
+        # Set a valid value
+        ui.value = TEST_CAMERA_VALUES
+
+        # So it turns out that the validation stuff only updates when changes are made
+        # in the UI rather than programmatically. Since we know we've set a valid value,
+        # and that we've made changes we just manually set the relevant values.
```
JuanCab, line 52: This is rather sad, forcing the test to work, but I suppose it checks that changing `ui.savebuttonbar.unsaved_changes` and ` ui.is_valid.value` have the intended effects on the save button.

JuanCab (approved):

JuanCab (changes requested):
Might want to fix the typo, otherwise all good.

stellarphot/settings/views.py
mwcraig: I had to break it into steps because I was confusing myself when I tried to make it a single line 😬

stellarphot/settings/tests/test_views.py
mwcraig: Yeah, this was not my first choice. I tried calling the handlers that I thought `ipyautoui` was invoking, but i) they were private (started with `_`) and ii) didn't do anything.

PR #311, opened 2024-05-01
Remove save button code made unnecessary by #310

### LINKED ISSUE #309: Save button is enabled after clicking edit even though no changes have been made
This sequence of events leads to an save button that is enabled even though no changes have been made:

1. Make a new camera or edit an existing one.
2. Click save (after making a change if necessary)
3. Click confirm if necessary
4. Click Edit
5. The save button will be enabled even though no changes have been made.

The root cause is [these lines](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/settings/custom_widgets.py#L114-L118) which are an attempt to make to prevent the save button dfrom being enabled unless the model is valid.

That is useful, but whether there are any unsaved changes also needs to be taken into account. That is available in `SaveButtonBar.unsaved_changes`.

mwcraig:
This fixes #309 by adding a test for that case and removing the problematic code in `custom_widgets`. The correct fix for enabling/disabling the save and revert buttons was made in #310.

### DIFF
--- a/stellarphot/settings/tests/test_custom_widgets.py
+++ b/stellarphot/settings/tests/test_custom_widgets.py
@@ -271,8 +271,8 @@ class TestChooseOrMakeNew:
         # Immediately after the edit button has been clicked, the save button should be
         # disabled because no changes have been made yet.
         #
-        # That should remain true even after the user makes a change and then saves it
-        # and then edits again.
+        # That should remain true even after the user makes a change, then saves it
+        # and then edits again but has not yet made any changes.
         saved = SavedSettings(_testing_path=tmp_path)
         camera = Camera(**TEST_CAMERA_VALUES)
         saved.add_item(camera)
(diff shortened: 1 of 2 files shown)


stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -267,6 +267,42 @@ def test_edit_button_returns_after_making_new_item(self, tmp_path):
         # The edit button should now be displayed
         assert choose_or_make_new._edit_button.layout.display != "none"
 
+    def test_save_button_disabled_when_no_changes(self, tmp_path):
+        # Immediately after the edit button has been clicked, the save button should be
+        # disabled because no changes have been made yet.
+        #
+        # That should remain true even after the user makes a change and then saves it
```
JuanCab, line 274: This statement is unclear, I assume you mean this should remain true after a user makes a change and then saves it UNTIL they start editing and have made new changes or something like that.

JuanCab (changes requested):
Looks good although I suggest fixing a comment first.

mwcraig:
Merging, since the comment was updated...

PR #314, opened 2024-05-01
Improve handling of revert button enabled state

mwcraig:
This PR fixes a mistake I think made in #310. in handling whether the save and revert buttons in the button bar were enabled.

In that PR I made it so that the save and revert buttons were enabled only if:

+ there were unsaved changes AND
+ the widget held a valid pydanitc model.

That makes sense for the save button -- you should only be able to save a valid model.

Requiring the widget hold a valid pydantic for the *revert* button makes little sense. One of the reasons you might want to revert, after all, is that you have accidentally entered one or more invalid fields.

This PR changes the behavior for the revert button so that it is enabled whenever there are unsaved changes.

@JuanCab and @Tanner728 -- a look at the code would be great, but confirmation that this new logic makes sense would be enough...I'm reasonably confident the code does what it should.

### DIFF
--- a/stellarphot/settings/tests/test_views.py
+++ b/stellarphot/settings/tests/test_views.py
@@ -27,22 +27,24 @@ class TestUiGenerator:
             if hasattr(widget, "continuous_update"):
                 assert not widget.continuous_update
 
-    def test_disabled_state_save_revert_button(self):
-        # We want the save and revert buttons to be enabled only when
+    def test_disabled_state_save_button(self):
+        # We want the save button to be enabled only when
         # 1. The user has made a change, AND
         # 2. The value in the widget is a valid pydantic model
 
         ui = ui_generator(Camera)
         # The save button should be disabled
         assert ui.savebuttonbar.bn_save.disabled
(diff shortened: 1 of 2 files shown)


Tanner728 (approved):
All of the code looks good to me, although I don't have much experience with UI stuff. I did some quick testing of it and all of it seems to work as expected. I know we had talked about the revert button earlier but have forgotten what you had decided to do with it as it still does nothing. Just to let you know.

mwcraig:
> it still does nothing. Just to let you know.

A PR to fix that is coming, but I needed to address enabling the revert button first (and am trying to keep the PRs smaller and easier to review)

JuanCab (approved):
The code looks good.  I also looked at the notebook, specifically:

I executed the following commands:

```
conda create -n test-314 python=3.11 astroscrappy
conda activate test-314
pip install git+https://github.com/mwcraig/stellarphot.git@improve-default-revert-enabled
```
and then tried the `saveable-settings.ipynb` notebook.  It behaved as advertised (as far as I could tell through creating one entry and then changing it).

PR #315, opened 2024-05-01
Make the revert button reset to saved value

### LINKED ISSUE #312: Need way to cancel an edit of a camera, etc
Right now there is no easy way to stop editing something if you decide you don't want to make any changes. The obvious way to do that is with the "Revert" button (the yellow counter-clockwise arrow button).

Clicking on that button should set the UI to the same state it would be in if you had selected that item.

mwcraig:
This PR fixes #312 by making the revert button end an edit and reset the widget value back whatever settings were there pre-edit.

Note that the revert button doesn't do anything if you click while making a new item. That is deliberate (at least for now) because either i) there are no items so the user needs to make a valid one and can edit fields as needed to get there or ii) the user can use the dropdown to select an existing item.

@Tanner728 -- the revert button should do something now 😀

### DIFF
--- a/stellarphot/settings/custom_widgets.py
+++ b/stellarphot/settings/custom_widgets.py
@@ -224,6 +224,9 @@ class ChooseOrMakeNew(ipw.VBox):
         # Update the current state of the widget
         self._editing = True
 
+        # Enable the revert button so that the user can cancel the edit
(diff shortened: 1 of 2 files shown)


Tanner728 (approved):
Seems all good to me.

mwcraig:
Force pushed to resolve the conflict, merging....one more widget PR coming today

PR #318, opened 2024-05-03
Properly set widget value when making a new item

### LINKED ISSUE #317: Making a new item confuses the widget
If you have at least one item, e.g. camera, and you select the "Make new..." option and then select the original item again then the widget display is blank. The reason is that the private method `_init_ui` from ipyautoui resets the values of each individual text box (or whatever) but does not change the value of the widget as a whole. 

As a result, after selecting "Make new..." with widget value is still the value for the existing item. So when this custom widget tries to redisplay after you select the existing item again, the UI isn't updated because the UI thinks nothing has changed, so no handlers are triggered.

This description is probably a little confusing, but trust me it is an issue. Fix incoming momentarily.

mwcraig:
This fixes #317 by setting the ipyautoui widget to the appropriate value when making a new item.

@JuanCab @Tanner728 -- I'm going to merge this in the morning if tests pass so that I can finally get this custom widget wrapped up, but if you want to try this out here are the steps:

1. Make a camera. (or observatory or..._
2. Save it.
3. Once the dropdown updates, select "Make a new..."
4. Select the item you made in step 1.

Prior to this fix you got a bunch of blank fields in step 4. Now it works.

### DIFF
--- a/stellarphot/settings/custom_widgets.py
+++ b/stellarphot/settings/custom_widgets.py
@@ -72,7 +72,7 @@ class ChooseOrMakeNew(ipw.VBox):
             message=f"Replace value of this {self._display_name}?",
         )
 
-        self._item_widget = self._make_new_widget()
+        self._item_widget, self._widget_value_new_item = self._make_new_widget()
 
         self._edit_button.on_click(self._edit_button_action)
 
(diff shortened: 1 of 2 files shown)


Tanner728 (approved):
Did a quick test of this and it works as expected.

PR #325, opened 2024-05-09
Add update argument to save method

mwcraig:
This pull request completes the mechanism for saving files in the working directory by adding an `update` argument to `save` that causes a save to add additional partial settings to whatever is on disk.

### DIFF
--- a/stellarphot/settings/settings_files.py
+++ b/stellarphot/settings/settings_files.py
@@ -394,18 +394,18 @@ class PhotometryWorkingDirSettings:
                     # Grab a dict of the settings, only keeping values that are not
                     # None. Making it dict so that the update method can be used to
                     # merge the two sets of settings.
-                    temp_settings_arg = {
+                    passed_partial_settings = {
                         k: v for k, v in settings.model_dump().items() if v is not None
                     }
-                    temp_settings_file = self._partial_settings.model_dump()
+                    existing_partial_settings = self._partial_settings.model_dump()
 
                     # The order matters here. Keys in the argument version of the
                     # settings will overwrite the keys in the file version of the
                     # settings. Note that update works in-place.
-                    temp_settings_file.update(temp_settings_arg)
+                    existing_partial_settings.update(passed_partial_settings)
 
                     settings = PartialPhotometrySettings.model_validate(
-                        temp_settings_file
+                        existing_partial_settings
                     )
                 # set variable file to point to appropriate (partial or full)
                 # settings file location.
--- a/stellarphot/settings/settings_files.py
+++ b/stellarphot/settings/settings_files.py
@@ -343,7 +343,7 @@ class PhotometryWorkingDirSettings:
         else:
             return True
 
-    def save(self, settings):
+    def save(self, settings, update=False):
         """
         Save the partial or full settings to the working directory. Note well that
         this removes any partial settings file if called with valid full settings.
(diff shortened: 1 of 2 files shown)


stellarphot/settings/settings_files.py
```diff
@@ -365,13 +389,32 @@ def save(self, settings):
                     raise ValueError(
                         "Cannot save partial settings when full settings already exist."
                     )
+                # Are we updating or replacing partial settings?
+                if update and self._partial_settings is not None:
+                    # Grab a dict of the settings, only keeping values that are not
+                    # None. Making it dict so that the update method can be used to
+                    # merge the two sets of settings.
+                    temp_settings_arg = {
```
JuanCab, line 397: I know I am being a stickler, but I don't like the names of these variables because it is not clear which are the settings currently in the partial settings in the object and which are the updates.  Might I suggest instead of `temp_settings_arg` you called it `passed_partial_settings` and instead of `temp_settings_file` you call it `existing_partial_settings` or some variant like that.  Have settings from two sources both starting with `temp` makes the code hard to read.

stellarphot/settings/tests/test_settings_file.py
```diff
@@ -486,3 +486,52 @@ def test_load_one_setting_present(self):
 
         settings = settings_file.load()
         assert settings == partial_settings
+
+    def test_save_updates_instead_of_replacing(self):
+        # Test that saving settings adds to whatever partial settings have
```
JuanCab, line 491: Simple but clear test of updating partial settings.

stellarphot/settings/settings_files.py
```diff
@@ -352,8 +352,32 @@ def save(self, settings):
...
             The settings to save.
+
+        update : bool, optional
+            If True, the settings are updated -- see Note below for more description.
+            If False, the settings are overwritten.
+
+        Notes
+        -----
+
+        If ``update`` is True, then the settings are updated. This means that if the
```
JuanCab, line 363: So should it be an error to have update `True` when the settings are full settings?  Is there anything actually wrong with updating full settings?  Because I understand the logic below, full settings will not get updated.

stellarphot/settings/tests/test_settings_file.py
```diff
@@ -486,3 +486,52 @@ def test_load_one_setting_present(self):
...
+        )
+        settings_file.save(partial_settings_obs, update=True)
+        from_file2 = settings_file.load()
+        # Make sure the camera is still there
+        assert from_file2.camera == DEFAULT_PHOTOMETRY_SETTINGS["camera"]
+        # Make sure the observatory is there
+        assert from_file2.observatory == DEFAULT_PHOTOMETRY_SETTINGS["observatory"]
+
+    def test_save_update_completing_partial_makes_full(self):
+        # Test that saving a partial settings file and then updating it to a
```
JuanCab, line 517: Logic of this test also makes sense.

JuanCab (changes requested):
Good idea, but I have some questions in my comments...

stellarphot/settings/settings_files.py
mwcraig: Good suggestion!

stellarphot/settings/settings_files.py
mwcraig: Correct, full settings would override anything saved on disk.

JuanCab (approved):

PR #327, opened 2024-05-13
Add option to show/hide item details

mwcraig:
This adds the option to make the details of a selected object hideable. The motivation is to at least have the option of saving space in the photometry notebooks 1 and 2.

After this change the widget looks like this if hideable is enabled; if it is not enabled then there is no change in appearance:

corm-play-BA…__2__-_JupyterLab

Cursor_and_corm-play-BA…__2__-_JupyterLab

### DIFF
--- a/stellarphot/settings/custom_widgets.py
+++ b/stellarphot/settings/custom_widgets.py
@@ -14,6 +14,8 @@ from stellarphot.settings import (
     ui_generator,
 )
 
+__all__ = ["ChooseOrMakeNew", "Confirm"]
+
 DEFAULT_BUTTON_WIDTH = "300px"
 
 
@@ -39,7 +41,9 @@ class ChooseOrMakeNew(ipw.VBox):
         PassbandMap.__name__,
     ]
 
-    def __init__(self, item_type_name, *arg, _testing_path=None, **kwargs):
+    def __init__(
+        self, item_type_name, *arg, details_hideable=False, _testing_path=None, **kwargs
+    ):
         if item_type_name not in self._known_types:
             raise ValueError(
                 f"Unknown item type {item_type_name}. Must "
@@ -60,14 +64,25 @@ class ChooseOrMakeNew(ipw.VBox):
         # and track if we are making a new item
         self._making_new = False
 
+        # keep track of whether there is a "show details" checkbox
+        self._show_details_shown = details_hideable
+
         self._display_name = item_type_name.replace("_", " ")
 
         # Create the child widgets
+
+        # Descriptive title
         self._title = ipw.HTML(
             value=(f"<h2>Choose a {self._display_name} " "or make a new one</h2>")
         )
 
+        # Selector for existing items or to make a new one
         self._choose_existing = ipw.Dropdown(description="")
+
+        # Option to show/hide details, only displayed if user wants it.
+        self._show_details_ui = ipw.Checkbox(description="Show details", value=True)
+        self._show_details_ui.layout.display = "flex" if details_hideable else "none"
+        self._show_details_cached_value = self._show_details_ui.value
 
         self._edit_delete_container = ipw.HBox(
             # width below was chosen to match the dropdown...would prefer to
@@ -83,19 +98,29 @@ class ChooseOrMakeNew(ipw.VBox):
             description=f"Delete this {self._display_name}",
         )
 
+        # Put almost everything into a VBox, which we can perhapds wrap in an
+        # accordian widget.
+        self._details_box = ipw.VBox()
+
         self._edit_delete_container.children = [self._edit_button, self._delete_button]
 
         self._confirm_edit_delete = Confirm()
 
         self._item_widget, self._widget_value_new_item = self._make_new_widget()
 
+        # Put all of the details into a box that can be easily hidden
+        self._details_box.children = [
+            self._edit_delete_container,
+            self._confirm_edit_delete,
+            self._item_widget,
+        ]
+
         # Build the main widget
         self.children = [
             self._title,
             self._choose_existing,
-            self._edit_delete_container,
-            self._confirm_edit_delete,
-            self._item_widget,
+            self._show_details_ui,
+            self._details_box,
         ]
 
         # Set up the dropdown widget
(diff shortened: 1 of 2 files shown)


stellarphot/settings/custom_widgets.py
```diff
@@ -83,19 +98,29 @@ def __init__(self, item_type_name, *arg, _testing_path=None, **kwargs):
             description=f"Delete this {self._display_name}",
         )
 
+        # Put almost everything into a VBox, which we can perhapds wrap in an
```
JuanCab, line 101: misspelling of "perhaps" in comments... and I assume you are not using 'accordion' widget, so maybe drop mention of that....

stellarphot/settings/custom_widgets.py
```diff
@@ -138,6 +163,17 @@ def __init__(self, item_type_name, *arg, _testing_path=None, **kwargs):
...
+        # Set up an observer to show/hide the details box if the check box
+        # is clicked
+        self._show_details_ui.observe(self._show_details_handler, names="value")
+
+    @property
+    def value(self):
+        """
+        The value of the widget.
+        """
+        return self._item_widget.model(**self._item_widget.value)
```
JuanCab, line 175: This last line of the `value` method took me some effort to understand.  It should be dictionary unpacking the `self._item_widget.value` and passes it to `self._item_widget.model`.  Since `_item_widget` is produced by `ipyautoui`, I assumed `model` is from pydantic, but that suggests it creates a model, NOT returning a value.  Can you clarify this?

stellarphot/settings/custom_widgets.py
```diff
@@ -217,6 +259,13 @@ def _handle_selection(self, change):
             # but does no harm in the other cases
             self._set_disable_state_nested_models(self._item_widget, True)
 
+            # We may have arrived here by choosing a different item while
```
JuanCab, line 262: Weird that you have to do this....

stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -550,6 +550,141 @@ def test_weird_sequence_of_no_clicks(self, tmp_path):
         # ...the edit/delete buttons should be displayed
         assert choose_or_make_new._edit_delete_container.layout.display != "none"
 
+    @pytest.mark.parametrize("hideable", [True, False])
+    def test_details_hideable_or_not(self, tmp_path, hideable):
```
JuanCab, line 554: Logic for this test looks reasonable.

stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -550,6 +550,141 @@ def test_weird_sequence_of_no_clicks(self, tmp_path):
...
+            "camera", details_hideable=True, _testing_path=tmp_path
+        )
+
+        assert choose_or_make_new._details_box.layout.display != "none"
+
+        # Set "show details" box to unchecked so details are hidden
+        choose_or_make_new._show_details_ui.value = False
+
+        # Note whether the "show details" box is checked or not
+        show_state = False
```
JuanCab, line 600: I'm confused by this, it just seems to create a variable `show_state` which is only used once later to assert `False` ... why create the variable?

stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -550,6 +550,141 @@ def test_weird_sequence_of_no_clicks(self, tmp_path):
...
+        new_camera["name"] = "new camera"
+        choose_or_make_new._item_widget.value = new_camera
+        choose_or_make_new._item_widget.savebuttonbar.bn_save.click()
+
+        # "show details" should be visible
+        assert choose_or_make_new._show_details_ui.layout.display != "none"
+
+        # Details should be hidden
+        assert choose_or_make_new._details_box.layout.display == "none"
+        assert choose_or_make_new._show_details_ui.value == show_state
```
JuanCab, line 626: Why not just `assert choose_or_make_new._show_details_ui.value == False` since `show_state` is only defined above to use in this one test?

stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -550,6 +550,141 @@ def test_weird_sequence_of_no_clicks(self, tmp_path):
...
+
+    def test_details_hideable_not_set_preserves_old_behavior(self, tmp_path):
+        # The details should not be hidden if hideable is false
```
JuanCab, line 629: The logic of this test seems sound as well.

JuanCab (approved):
Looks good assuming the tests pass.  Only issue I had is I got lost in some of the details of the code.  I noted them.

stellarphot/settings/custom_widgets.py
mwcraig: It was easier than adding a cancel button...

stellarphot/settings/custom_widgets.py
mwcraig: Fixewd

stellarphot/settings/custom_widgets.py
mwcraig: It is creating an instance of the `model`.

`model` is the attribute that `ipyautoui` sets that contains the `pydantic` model the ui is based on. Would it be clearer if the code was changed to 

```python 
return self._item_widget.model.model_validate(self._item_widget.value)
```

stellarphot/settings/tests/test_custom_widgets.py
mwcraig: seemed like a good idea at the time, but I don't remember why. Removing it.

stellarphot/settings/tests/test_custom_widgets.py
mwcraig: Fixed

stellarphot/settings/tests/test_custom_widgets.py
mwcraig: Fixed

stellarphot/settings/custom_widgets.py
JuanCab: Doesn't `model_validate` return the entire model (assuming it is valid based on the input value).  I guess given most of our `item_widgets` are effectively automatically generated widgets for controlling collections of values, the name `value` for the function seems a bit confusing... not sure I can offer a better name right now.  Maybe in the comments just state it returns an instance of the model corresponding to the widget's current values.

stellarphot/settings/tests/test_custom_widgets.py
JuanCab: Good.

JuanCab (approved):

PR #329, opened 2024-05-14
Fix center profile

### LINKED ISSUE #328: `CenterAndProfile` fails if the image has a background
It turns out that `profiles.CenterAndProfile` fails in many cases if the image has a background. It works if the star is bright enough (or anytime the peak is much, much higher than the background) but fails otherwise because it is fitting a Gaussian  to the profile that expects to go to zero at large enough radii.

PR incoming...

### LINKED ISSUE #330: `CenterAndProfile` does not return a background-subtracted radial profile
An incorrect profile is below. It turns out that none of our profile tests had a background....

lab_-_JupyterLab

mwcraig:
This fixes #328 by subtracting a rough background in the initial creation of the radial profile. It also fixes #330 by subtracting the background.

### DIFF
--- a/stellarphot/photometry/profiles.py
+++ b/stellarphot/photometry/profiles.py
@@ -178,8 +178,10 @@ class CenterAndProfile:
             profile_radius = cutout_size // 2
 
         radii = np.linspace(0, profile_radius, profile_radius + 1)
-        # Get a rough profile without background subtraction
-        self._radial_profile = RadialProfile(self._data, self._cen, radii)
+        # Get a rough profile with rough background subtraction -- note that
+        # NO background subtraction does not work.
+        background = sigma_clipped_stats(self._cutout.data)[1]
+        self._radial_profile = RadialProfile(self._data - background, self._cen, radii)
 
         self._sky_area = None
 
--- a/stellarphot/photometry/profiles.py
+++ b/stellarphot/photometry/profiles.py
@@ -247,16 +247,31 @@ class CenterAndProfile:
         """
         radii = []
         pixel_values = []
-        for rad, ap in zip(
-            self.radial_profile.radius, self.radial_profile.apertures, strict=True
-        ):
+        for ap in self.radial_profile.apertures:
+            # Calculate the distance of each pixel from the center
+            grid_x, grid_y = np.mgrid[: self._data.shape[0], : self._data.shape[1]]
+            dist_from_cen = np.sqrt(
+                (grid_x - self._cen[1]) ** 2 + (grid_y - self._cen[0]) ** 2
+            )
+
+            # Get only the data in this aperture
             ap_mask = ap.to_mask(method="center")
             ap_data = ap_mask.multiply(self._data)
+            dist_from_cen = ap_mask.multiply(dist_from_cen)
+
+            # Drop any data where the aperture mask was zero and flatten
             good_data = ap_data != 0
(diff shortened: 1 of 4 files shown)


mwcraig:
@JuanCab -- this is actually ready for a look now

stellarphot/photometry/profiles.py
```diff
@@ -245,17 +247,32 @@ def pixel_values_in_profile(self):
...
         radii = []
         pixel_values = []
-        for rad, ap in zip(
-            self.radial_profile.radius, self.radial_profile.apertures, strict=True
-        ):
+        for ap in self.radial_profile.apertures:
+            # Calculate the distance of each pixel from the center
+            grid_x, grid_y = np.mgrid[: self._data.shape[0], : self._data.shape[1]]
+            dist_from_cen = np.sqrt(
+                (grid_x - self._cen[1]) ** 2 + (grid_y - self._cen[0]) ** 2
```
JuanCab, line 254: Is this correct, you have `grid_x` determined by `: self._data.shape[0]` but you compute the distance from center using `(grid_x - self._cen[1])`?

stellarphot/photometry/tests/test_profiles.py
```diff
@@ -72,6 +72,8 @@ def test_find_center_no_star():
 
 
 def test_radial_profile():
+    # Test that both curve of growth and radial profile are correct
+
```
JuanCab, line 76: Looks like a reasonable test

stellarphot/photometry/tests/test_profiles.py
```diff
@@ -102,3 +112,32 @@ def test_radial_profile_exposure_is_nan():
     assert np.isnan(rad_prof.noise(c, np.nan)[-1])
     assert all(np.isfinite(rad_prof.curve_of_growth.profile))
     assert all(np.isfinite(rad_prof.radial_profile.profile))
+
+
+def test_radial_profile_with_background():
+    # Regression test for #328 -- image with a background level
+    image = make_gaussian_sources_image(SHAPE, STARS)
+    image = image + +make_noise_image(
```
JuanCab, line 120: Superfluous + sign, I think...

stellarphot/photometry/tests/test_profiles.py
```diff
@@ -102,3 +112,32 @@ def test_radial_profile_exposure_is_nan():
     assert np.isnan(rad_prof.noise(c, np.nan)[-1])
     assert all(np.isfinite(rad_prof.curve_of_growth.profile))
     assert all(np.isfinite(rad_prof.radial_profile.profile))
+
+
+def test_radial_profile_with_background():
+    # Regression test for #328 -- image with a background level
```
JuanCab, line 118: Again, I believe the logic here is sound, though it took a bit for me to understand it.

JuanCab (dismissed):
Looks good... but check my comment about the computation of `list_from_cen` and make sure that is correct...

stellarphot/photometry/profiles.py
mwcraig: Yeah; this is the array vs image thing.

stellarphot/photometry/tests/test_profiles.py
mwcraig: Fixed

mwcraig:
I fixed the "+" sign and double-checked the `1`/`0` index thing to confirm it is correct, going to merge.

PR #361, opened 2024-05-31
Fix description of aperture radius

### LINKED ISSUE #360: Fix description of aperture radius
@Tanner728 [pointed out](https://github.com/feder-observatory/stellarphot/pull/358#pullrequestreview-2086846247) in #358 that the aperture radius is described as the "radius of the circular annulus", which is confusing. It should say "circular paerture" instead.

mwcraig:
Fixes #360


Tanner728 (approved):
Looks good to me!

PR #364, opened 2024-06-04
Update comparison star notebook 

### LINKED ISSUE #183: Notebook 02 problems
- [ ] Add visual indication that aperture and variable files need to be saved or autosave them
- [x] Add settings for file names
- [x] Bring back some way to save TIC information?
- [ ] Offer a choice of source selection (maybe)

mwcraig:
This is still a hot mess, but attempts to fix #183 by adding a source location widget to the comparison viewer and autosaving the aperture file.

Issues to be dealt with before merging:

- [x] Add some tests.
- [x] Decide on a nicer layout
- [x] Chuck the TESS stuff?
- [x] What should be done about the variable information? Punting for now

### DIFF
--- a/stellarphot/gui_tools/comparison_functions.py
+++ b/stellarphot/gui_tools/comparison_functions.py
@@ -17,6 +17,7 @@ from stellarphot import SourceListData
 from stellarphot.gui_tools.fits_opener import FitsOpener
 from stellarphot.gui_tools.seeing_profile_functions import set_keybindings
 from stellarphot.io import TOI, TessSubmission, TessTargetFile
+from stellarphot.settings import SourceLocationSettings, ui_generator
 from stellarphot.utils.comparison_utils import (
     crossmatch_APASS2VSX,
     in_field,
@@ -567,6 +568,7 @@ class ComparisonViewer:
         """
         )
 
+        self.source_locations = ui_generator(SourceLocationSettings)
         iw = ImageWidget()
         out = ipw.Output()
         set_keybindings(iw)
@@ -581,7 +583,9 @@ class ComparisonViewer:
         self._make_tess_save_box()
         box = ipw.VBox()
         inner_box = ipw.HBox()
-        inner_box.children = [iw, legend]
+        source_legend_box = ipw.VBox()
+        source_legend_box.children = [self.source_locations, legend]
+        inner_box.children = [iw, source_legend_box]  # legend]
         box.children = [
             self._file_chooser.file_chooser,
             self.object_name,
--- a/stellarphot/gui_tools/comparison_functions.py
+++ b/stellarphot/gui_tools/comparison_functions.py
@@ -22,6 +22,7 @@ from stellarphot.settings import (
     SourceLocationSettings,
     ui_generator,
 )
+from stellarphot.settings.custom_widgets import SettingWithTitle
 from stellarphot.utils.comparison_utils import (
     crossmatch_APASS2VSX,
     in_field,
@@ -272,7 +273,10 @@ class ComparisonViewer:
 
         self.box, self.iw = self._viewer()
 
-        self.photom_apertures_file = photom_apertures_file
+        if photom_apertures_file is not None:
+            # Set the source location file name to the name passed in
+            self._set_source_location_file_to_value(photom_apertures_file)
+
         self.overwrite_outputs = overwrite_outputs
 
         if file:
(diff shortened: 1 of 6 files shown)


mwcraig:
@Tanner728 @JuanCab this is still a work in progress but any feedback you have based on what is here so far is welcome....

mwcraig:
@Tanner728  -- thanks for taking a look!

> marked some minor typos but overall looks good. 

Can you comment on those in github so I can fix them?

> My one question is why you removed the TESS stuff, is it becoming too much work and causing issues?

I'd like to mostly move that into another widget, because the comparison viewer is pretty complicated even without it. Though it doesn't exist yet, there would be some sort of "generate your TESS stuff" widget for generating those things. I have kept the TESS code in a separate branch.

> Also where is this new widget located?

It hasn't been added yet, will do that...

mwcraig:
While this could still be beautified, I think it is ready for review. Further beautification can wait until a later pull request

mwcraig:
FYI, I think the codecov report on this PR i incorrect -- all but one of the lines marked as uncovered in the patch are now covered with the most recent commit.

Tanner728:
> @Tanner728 -- thanks for taking a look!
> 
> > marked some minor typos but overall looks good.
> 
> Can you comment on those in github so I can fix them?
> 
> > My one question is why you removed the TESS stuff, is it becoming too much work and causing issues?
> 
> I'd like to mostly move that into another widget, because the comparison viewer is pretty complicated even without it. Though it doesn't exist yet, there would be some sort of "generate your TESS stuff" widget for generating those things. I have kept the TESS code in a separate branch.
> 
> > Also where is this new widget located?
> 
> It hasn't been added yet, will do that...

Sorry had forgotten to hit submit on the review.

mwcraig:
I think this is go to go -- thanks for catching the types @Tanner728 and for the review/off-line discussion @JuanCab

stellarphot/gui_tools/comparison_functions.py
```diff
@@ -330,82 +333,46 @@ def _set_object(self):
...
-            self.object_name.value = self._file_chooser.header["object"]
+            self.object_name.value = (
+                f"<h2> Object: {self._file_chooser.header['object']}</h2>"
+            )
+            self._object = self._file_chooser.header["object"]
         except KeyError:
-            # No object, will show empty box for name
+            # No object, will show placeholder
+            self.object_name.value = "<h2>Object unknown</h2>"
             self.object_name.disabled = False
```
JuanCab, line 343: Seems weird that `self.object_name.disabled = False` here, is that correct?  I assume this is because you used to have an empty box here... shouldn't you just remove the `self.object_name.disabled` variable or is it used elsewhere?

stellarphot/gui_tools/comparison_functions.py
```diff
@@ -560,10 +505,12 @@ def _viewer(self):
 
         legend = ipw.HTML(
             value="""
-        <h3>Green circles -- Gaia stars within 2.5 arcmin of target</h3>
-        <h3>Red circles -- APASS stars within 1 mag of target</h3>
-        <h3>Blue circles -- VSX variables</h3>
-        <h3>Red × -- Exclude as target or comp</h3>
+        <ul>
+        <li>Green circles -- Gaia stars within 2.5 arcmin of target</li>
+        <li>Red circles -- APASS stars within 1 mag of target</li>
```
JuanCab, line 510: OK, I'm going to be difficult here, since I should have thought of this earlier.  A not insignificant subset of the population is color-blind... can't we use different shapes as well as colors?

JuanCab (approved):
See my comments, looks good for the most part although I did catch a potential accessibility issue that might need to be addressed.

JuanCab (changes requested):
Looks good for the most part, but I noticed an accessibility issue for the color-blind.  I noted it.  You can fix it now if easy otherwise consider adding it as an issue for the future and push this merge through.

stellarphot/gui_tools/comparison_functions.py
mwcraig: Yeah, should be axed

stellarphot/gui_tools/comparison_functions.py
mwcraig: I'm getting to open an issue for this -- I think the markers are also maybe too small. You are correct though that straight up using red/green is about the worst possible choice from an accessibility POV.

mwcraig:
Thanks @JuanCab -- I opened #368 to keep track of the symbol color issue and addressed the rest of your comments, I think.

JuanCab (approved):
LGTM

PR #366, opened 2024-06-05
Updated ruff and black precommit revs and added codespell to precommit.

JuanCab:
I have updated the precommit config yaml to include codespell in the recommit as well as manually updating the repo rev for ruff and black in the precommit.

I also added precommit.ci as an app for stellarphot (both for feder-observatory and JuanCab), which should do weekly checks on the precommit rev versions and issue PR when there is an update to the rev on one of our recommits.


mwcraig (commented):
This looks good -- should we make the changes to spelling that codespell suggests? And maybe add some codespell exceptions:

- It is spell checking the SVG logo file
- It is trying to change TOI to something else. 

I think there is a way to add exceptions to pyproject.toml...

JuanCab:
> This looks good -- should we make the changes to spelling that codespell suggests? And maybe add some codespell exceptions:
> 
> * It is spell checking the SVG logo file
> * It is trying to change TOI to something else.
> 
> I think there is a way to add exceptions to pyproject.toml...

I can work on the exceptions, but where are you seeing codespell's suggestions?

PR #370, opened 2024-06-11
Make name of field match snake case version of class name

mwcraig:
For reasons I cannot recall, the name for the `SourceLocationSettings` in `PhotometrySettings` is `source_locations`. For all of the fields in `PhotometrySettings` the name of the field is the snake case version of the class name. So `PhotometryApertures` is `photometry_apertures`, etc.

It will turn out to be convenient for the last set of widgets if the name is *always* the snake case version o f the class name.]

This PR does that.

@JuanCab -- it would be great if this could get a quick review.

### DIFF
--- a/stellarphot/gui_tools/comparison_functions.py
+++ b/stellarphot/gui_tools/comparison_functions.py
@@ -458,7 +458,7 @@ class ComparisonViewer:
         """
         self.photometry_settings.save(
             PartialPhotometrySettings(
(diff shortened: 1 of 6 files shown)


JuanCab (approved):
Assuming the tests pass, this looks good (and it is just a variable name change).  I assume you just did a multi-file search and replace here (and you did look for cases where it was part of a variable name, which is good), hopefully you didn't miss an edge case.

PR #371, opened 2024-06-12
Add a review widget

mwcraig:
This pull request adds what will be the first of the new photometry notebooks (setting camera, observatory, passband map) and the final review-of-settings notebook. 

Remaining to be done:

- [x] add unit tests for `ReviewSettings`
- [x] add notebook for first step
- [x] add notebook for last step
- [x] add launcher entries for notebooks 

Fixes #377 which was uncovered during development of this new widget.
Fixes #182 -- the notebooks were not combined but things have been streamlined a bit.

PR #372, opened 2024-06-13
Use pytest-mock to create testing path for saved settings

mwcraig:
With this, the `_testing_path` argument can be removed from saved settings code. The alternative to going this route was to add `_testing_path` to another class, the `ReviewSettings` widget, which will be in #371.

### DIFF
--- a/stellarphot/settings/settings_files.py
+++ b/stellarphot/settings/settings_files.py
@@ -129,12 +129,9 @@ class SavedSettings:
             not exist. If False, the directory is not created, which is useful for
             testing.
         """
-        if _testing_path is not None:
-            data_dir = _testing_path
-        else:
-            data_dir = PlatformDirs(
-                "stellarphot", version=SETTINGS_FILE_VERSION
-            ).user_data_dir
+        data_dir = PlatformDirs(
+            "stellarphot", version=SETTINGS_FILE_VERSION
+        ).user_data_dir
         self._settings_path = Path(data_dir)
         if _create_path:
             if not self.settings_path.exists():
@@ -441,6 +438,11 @@ class PhotometryWorkingDirSettings:
     def load(self):
         """
         Load full or partial settings.
+
+        Returns
+        -------
+        PhotometrySettings | PartialPhotometrySettings | None
+            The settings loaded from disk, or None if there are no settings files.
         """
         # Assume we have nothing to begin....
         self._partial_settings = None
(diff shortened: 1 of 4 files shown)


stellarphot/settings/settings_files.py
JuanCab, line 1: Looks good (and I like the added bit of documentation).

stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -7,6 +7,7 @@
     Observatory,
     PassbandMap,
     SavedSettings,
+    settings_files,
```
JuanCab, line 10: What is this change doing?  There is no `settings_files` class, just the entire module (which contains your other settings classes you are importing here).

OK, on reviewing the code, it looks like you import `settings_files` directly to allow using of mocked versions of the attributes it contains.  That said, in the code it looks like `settings_file` is just an instantiation of the `SavedSettings` class so I am still troubled by this import and the matching import in `test_settings_file.py`.

stellarphot/settings/tests/test_settings_file.py
```diff
@@ -13,6 +13,7 @@
     PhotometrySettings,
     PhotometryWorkingDirSettings,
     SavedSettings,
+    settings_files,
```
JuanCab, line 16: Again, why?

JuanCab (changes requested):
OK, I followed this for the most part.  I did mark a point I am confused on regarding the importing of the entire `settings_file` module.

stellarphot/settings/tests/test_custom_widgets.py
mwcraig: @JuanCab  -- that is indeed confusing (I tried a bunch of variants of the import before getting something that worked). I added some more explanation [here](https://github.com/feder-observatory/stellarphot/pull/372/files#diff-b03c6eb9430bafc1d0960b6c6f312ba8afd1d7734bc60e7bd2b3b4e72bc03d58R85-R91). Can you please take a look and let me know if that helps clear it up?

JuanCab (approved):

stellarphot/settings/tests/test_custom_widgets.py
JuanCab: Thanks for the clarification, I have a better idea why although still seems a little like voodoo. :)

mwcraig: Yeah, it is definitely voodoo!

PR #373, opened 2024-06-13
Refactor title decoration

mwcraig:
This will make it easy to reuse for decorating accordion or tab titles. The original version could only decorate an HTML widget that it had created.

#372 should be merged before this one is.

### DIFF
--- a/stellarphot/settings/custom_widgets.py
+++ b/stellarphot/settings/custom_widgets.py
@@ -41,9 +41,7 @@ class ChooseOrMakeNew(ipw.VBox):
         PassbandMap.__name__,
     ]
 
-    def __init__(
-        self, item_type_name, *arg, details_hideable=False, _testing_path=None, **kwargs
-    ):
+    def __init__(self, item_type_name, *arg, details_hideable=False, **kwargs):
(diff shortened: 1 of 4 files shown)


stellarphot/settings/custom_widgets.py
```diff
@@ -1,6 +1,18 @@
 # Some settings require custom widgets to be displayed in the GUI. These are defined in
 # this module.
 
+# This workaround is for Python < 3.11. It is not needed in Python 3.11 and later.
```
JuanCab, line 4: Couldn't we just require 3.11?  I think we opted to require 3.10 last summer, but at this point...  In any case, if you are wedded to supporting 3.10, keep this.

JuanCab (approved):
Looks pretty straightforward except for the code supporting Python 3.10 (which is fine to keep).

stellarphot/settings/custom_widgets.py
mwcraig: We could require 3.11. I wasn't aware until I opened this PR that `StrEnum` was new in 3.11.

PR #375, opened 2024-06-20
Fix test that breaks in numpy 2.0

mwcraig:
@JuanCab -- this should be a pretty quick one to review for a change. I'm not quite sure how this test every worked, but it definitely fails with numpy 2 (See #374 for example). The change here fixes it.

### DIFF
--- a/stellarphot/tests/test_core.py
+++ b/stellarphot/tests/test_core.py
@@ -708,7 +708,7 @@ def test_catalog_colname_map():
     )
 
     assert catalog_dat["id"][0] == "ASASSN-V J000052.03+002216.6"
-    assert np.abs(catalog_dat["mag"][0].value - 12.660) < 1e-7
+    assert np.abs(catalog_dat["mag"][0].value - 12.660) < 2e-7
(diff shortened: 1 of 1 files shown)


stellarphot/tests/test_core.py
```diff
@@ -708,7 +708,7 @@ def test_catalog_colname_map():
     )
 
     assert catalog_dat["id"][0] == "ASASSN-V J000052.03+002216.6"
-    assert np.abs(catalog_dat["mag"][0].value - 12.660)
+    assert np.abs(catalog_dat["mag"][0].value - 12.660) < 1e-7
```
JuanCab, line 711: How the heck did this work before?

JuanCab (approved):
Approved, although I don't see how this worked before numpy 2.0

stellarphot/tests/test_core.py
mwcraig: 🤷

mwcraig:
The only change was to increase the tolerance a touch on windows. If tests pass I'll go ahead and merge.

PR #385, opened 2024-07-18
Fix #384, in which the comparison viewer misbehaves

### LINKED ISSUE #384: Comparison viewer does not properly display image if a second image is selected
The first time you select an image in the `comparison_viewer` all is well. If you select second image of a different object, though, the image isn't shown (even though it has properly loaded) because the display is still centered on the object in the previous image.

Incoming are a test for this and a fix.

mwcraig:
Most of the effort here was in constructing a test case that actually checks for #384. The fix was pretty trivial.


mwcraig:
(github is currently having issues with actions....)

JuanCab:
I'll look this over on Sunday.

mwcraig:
Closing/reopening to trigger tests

stellarphot/gui_tools/tests/test_comparison_functions.py
```diff
@@ -17,17 +20,7 @@
...
-    comparison_widget = cf.ComparisonViewer()
-    assert isinstance(comparison_widget.box, ipw.Box)
-
-
-@pytest.mark.parametrize("source_file_name", [None, "sources.ecsv"])
-@pytest.mark.parametrize("has_object", [True, False])
-@pytest.mark.remote_data
-def test_comparison_properties(tmp_path, has_object, source_file_name):
-    # Test that we can load a file...
+def make_ey_uma_image(with_object=False):
```
JuanCab, line 23: nice to load the file separately I suppose.

JuanCab (approved):
Assuming the tests clear, I think this is OK.  So the entire bug was just due to `_set_object` being called late?

mwcraig:
There were a couple of pieces to it. One was -set-object being called late another was that value, not selected, should be used with the file chooser because of a now fixed) bug in ipyautoui,

PR #392, opened 2024-07-31
Use new way to make source images in photutils

### LINKED ISSUE #378: Fix call to a function in photutils that has been renamed
Got this warning suddenly:

```
E   astropy.utils.exceptions.AstropyDeprecationWarning: The make_gaussian_sources_image function is deprecated and may be removed in a future version.
E           Use make_model_image instead.
```

mwcraig:
This fixes #378 by defining `make_gaussian_image_sources` in stellarphot based on its definition in stellarphot. stellarphot will at some point remove this function because it has been deprecated.

### DIFF
--- a/stellarphot/gui_tools/tests/test_seeing_profile.py
+++ b/stellarphot/gui_tools/tests/test_seeing_profile.py
@@ -7,7 +7,7 @@ import matplotlib
 import pytest
 from astropy.nddata import CCDData
 from astrowidgets import ImageWidget
-from photutils.datasets import make_gaussian_sources_image, make_noise_image
+from photutils.datasets import make_noise_image
 
 from stellarphot.gui_tools import (
     seeing_profile_functions as spf,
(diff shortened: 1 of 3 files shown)


mwcraig:
@JuanCab -- this should be another quick one....

JuanCab:
> This fixes #378 by defining `make_gaussian_image_sources` in stellarphot based on its definition in stellarphot. stellarphot will at some point remove this function because it has been deprecated.

Huh?  Do you mean since the original function is deprecated, we will be using an implementation stored as a part of the stellarphot code base?

JuanCab (approved):
Looks like it should work.  As long as the tests pass, should be good.

mwcraig:
> Do you mean since the original function is deprecated, we will be using an implementation stored as a part of the stellarphot code base?

Essentially, yes. Right now in `photutils` the old function `make_gaussian_image_sources` is a thin (maybe 10 lines of code) wrapper around the new `make_model_images`.  

We use `make_gaussian_image_sources` extensively enough that it seemed useful to still have that function around so I just copied the code from `photutils`.

PR #393, opened 2024-07-31
Update AAVSO filter list to match latest AAVSO update

### LINKED ISSUE #382: Repeated entries in AAVSO filters for `PassbandMap`
See this for description and comments: https://github.com/feder-observatory/stellarphot/pull/371#issuecomment-2229555224

mwcraig:
This also fixes #382 because the duplicate R and I entries have been removed.

I *think* the definitive source of filter names is https://www.aavso.org/filters -- I've asked in the AAVSO slack to make sure.

Another quickie @JuanCab 😬 -- good news is that there are not very many open issues milestoned to 2.0.0 left 😀

### DIFF
--- a/stellarphot/settings/aavso_models.py
+++ b/stellarphot/settings/aavso_models.py
@@ -2,13 +2,17 @@ from enum import Enum
 
 
 class AAVSOFilters(str, Enum):
+    """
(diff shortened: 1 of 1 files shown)


JuanCab (approved):
Looks good!

PR #394, opened 2024-07-31
Use different marker shapes and colors in comparison viewer

### LINKED ISSUE #368: Improve markers in comparison viewer, including addressing accessibility issue
As noted in [this comment](https://github.com/feder-observatory/stellarphot/pull/364#discussion_r1633423876), the comparison notebook uses red and green to distinguish between target and comparison stars. That is pretty rough for someone who is red/green colorblind and should be fixed.

In addition, it is in general hard to see these markers, especially the "x", without zooming way, way in.

mwcraig:
Fixes #368 by using markers of different shape to indicate whether a star is a target from Gaia, a comparison from APASS or a variable from VSX. The colors are also still different, but this should make the markers distinguishable for people with color blindness.

The legend has been updated to match the markers.

Last one of the day, @JuanCab, I promise!

### DIFF
--- a/stellarphot/gui_tools/comparison_functions.py
+++ b/stellarphot/gui_tools/comparison_functions.py
@@ -85,11 +85,25 @@ def make_markers(iw, RD, vsx, ent, name_or_coord=None):
             iw.center_on(name_or_coord)
 
     if vsx:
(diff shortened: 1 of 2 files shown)


JuanCab (approved):
Looks good, although I didn't test it in the actual Jupyter notebook.

PR #397, opened 2024-08-01
Testing to see if tweaking codecov.yml avoids testing coverage of test files

### LINKED ISSUE #396: Codecov report is way off
The codecov report is way off because it is including test and __init__ files in the count of source lines. When I run coverage locally it gets the correct coverage number (about 83%). 

@JuanCab -- do you have any time to look into this today?

JuanCab:
Here is an attempt to see if changing `codecov.yml` to include `ignore` entries matching those in the `pyproject.toml` will fix our problem with codecov wanting to test coverage or our tests (Issue #396).  Based on the documentation [here](https://coverage.readthedocs.io/en/latest/source.html#source) the `pyproject.toml` file which controls how the coverage command line is called SHOULD work, but maybe the `codecov.yml` file is being read and contradicting it for settings?

fixes #396


JuanCab:
This may have fixed the issue, although it is annoying because `codecov.yml` would need to be updated to match `pyproject.toml`.  In any case, the code coverage report seems more in line with what I recall before AND it is not listing any test code in the coverage report.  Please take a look @mwcraig !

mwcraig (approved):
Excellent, thanks!

PR #401, opened 2024-08-02
Add testing on python 3.12

mwcraig:


JuanCab:
Probably a good idea.  We should also put a reminder in to test on 3.13, since I think it is coming out in October.

JuanCab (approved):
Looks good, assuming this makes the necessary changes for 3.12 to work in the CI.

PR #407, opened 2024-08-06
Include proper files in sdist so they land in the wheel

### LINKED ISSUE #406: Launcher entries don't appear in jupyterlab
As reported [here](https://github.com/feder-observatory/stellarphot/issues/399#issuecomment-2269996252) and [here](https://github.com/feder-observatory/stellarphot/issues/399#issuecomment-2270114633) the launcher entries for stellarphot were not showing up in JupyterLab when installing the `alpha0` release.

mwcraig:
This fixes #406. The fundamental issue was that the wheel is built from the sdist, and the sdist did not include the logo or jupyter-app-launcher, so the wheel didn't include them. Unfortunate that wheel building didn't raise an error when it didn't find the files, but the missing files are in the sdist now, and a duplicate entry in the wheel section was removed.

~~Also removes a couple of `__init__.py` that are really unnecessary and were ending up in the environment's jupyter data folder.~~ Turns out at least one of them is necessary because we use `get_pkg_data_filename` to get the location of the photometry runner notebook.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,16 +59,24 @@ source = "vcs"
 [tool.hatch.build.hooks.vcs]
 version-file = "stellarphot/_version.py"
 
+[tool.hatch.build.targets.sdist]
+# This section determines what files are included in the source distribution
+include = [
+    # The source code
+    "/stellarphot",
+    # Logo for JupyterLab launcher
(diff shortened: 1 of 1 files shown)


mwcraig:
I'm going to go ahead and merge this to get alpha1 out....

PR #415, opened 2024-08-08
Remove code from tests executed on import in `test_models.py`

### LINKED ISSUE #412: Remove top-level executable code in `test_models`

mwcraig:
This PR removes from `settings/tests/test_models.py` any code outside of a function or class that did any meaningful execution and fixes #412. That means that we no longer call any stellarphot code outside of a class or function and that we no longer create astropy objects (e.g. a `Time` or `Quantity`) outside of a function or class.

A few different modules used the testing values from `test_models` so a few files were affected. 

While I was doing this I also decided to make the names uniform. Some had started `TEST_...` and others `DEFAULT_...` and now they all start out `TEST_...`

@JuanCab - a fairly quick review on this one would be nice, because a few more coming in soon to remove other executable code in tests depend on this rearrangement, I think.

### DIFF
--- a/stellarphot/settings/tests/test_models.py
+++ b/stellarphot/settings/tests/test_models.py
@@ -27,13 +27,13 @@ from stellarphot.settings.models import (
 TEST_APERTURE_SETTINGS = dict(radius=5, gap=10, annulus_width=15, fwhm=3.2)
 
 TEST_CAMERA_VALUES = dict(
-    data_unit=u.adu,
-    gain=2.0 * u.electron / u.adu,
+    data_unit="adu",
+    gain="2.0 electron / adu",
     name="test camera",
-    read_noise=10 * u.electron,
-    dark_current=0.01 * u.electron / u.second,
-    pixel_scale=0.563 * u.arcsec / u.pix,
-    max_data_value=50000 * u.adu,
+    read_noise="10.0 electron",
+    dark_current="0.01 electron / s",
+    pixel_scale="0.563 arcsec / pix",
+    max_data_value="50000.0 adu",
 )
 
 TEST_EXOPLANET_SETTINGS = dict(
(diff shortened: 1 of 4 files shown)


stellarphot/settings/tests/test_models.py
```diff
@@ -24,102 +23,111 @@
...
+    dark_current="0.01 electron / s",
+    pixel_scale="0.563 arcsec / pix",
+    max_data_value="50000.0 adu",
 )
 
-DEFAULT_EXOPLANET_SETTINGS = dict(
-    epoch=Time(0, format="jd"),
-    period=0 * u.min,
+TEST_EXOPLANET_SETTINGS = dict(
+    epoch={
```
JuanCab, line 39: So that is everything hidden in a `Time` object?  Wow...

JuanCab (approved):
I reviewed this (commit by commit) and it all seemed to make sense.  

Just curious why you switched to checking strings instead of direct Astropy Quantity comparison, was it more reliable?  I'm just worried the tests could fail if Astropy makes changes later to Quantity `__str__` representations.  I suppose that is unlikely.

stellarphot/settings/tests/test_models.py
mwcraig: yeah, I was a little surprised...

mwcraig:
> Just curious why you switched to checking strings instead of direct Astropy Quantity comparison, was it more reliable? 

Mostly to avoid having executable code at the top level and to be consistent about what the contents of the `TEST_....` things are. Sometimes they were objects from stellarphot, sometimes not. 

> I'm just worried the tests could fail if Astropy makes changes later to Quantity __str__ representations. I suppose that is unlikely.

Now we will at least catch it :). If that does happen it will break any already-written settings files....

PR #417, opened 2024-08-09
Remove executable code from `test_profile`

### LINKED ISSUE #414: Remove top-level executable code in `test_profiles`

mwcraig:
This closes #414 by removing the only executable code in `test_profiles` that was outside a function. Since the bit of code was used in another test in `gui_tools`, I made the table a [pytest fixture](https://docs.pytest.org/en/stable/how-to/fixtures.html) in `conftest.py`.

### DIFF
--- a/stellarphot/conftest.py
+++ b/stellarphot/conftest.py
@@ -2,7 +2,11 @@
 # test infrastructure.
 import os
 
+import pytest
+from astropy.table import Table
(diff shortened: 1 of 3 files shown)


JuanCab (approved):
Looks good to me.  I just started investigating pytest fixtures a few weeks ago, they really are convenient.

PR #418, opened 2024-08-09
Remove executable top-level code in `test_photometry`

### LINKED ISSUE #413: Remove top-level executable code in `test_photometry`

mwcraig:
The constants in this module that are now fixtures are used only in this test file so I left the fixtures in the file.

Fixes #413

### DIFF
--- a/stellarphot/photometry/tests/test_photometry.py
+++ b/stellarphot/photometry/tests/test_photometry.py
@@ -74,10 +74,17 @@ ZERO_CAMERA = Camera(
     max_data_value=40000 * u.adu,
 )
 
-# Fake observatory location
-FAKE_OBS = Observatory(
-    name="test observatory", latitude=0 * u.deg, longitude=0 * u.deg, elevation=0 * u.m
-)
+
+@pytest.fixture
+def fake_obs():
+    # Fake observatory location
+    return Observatory(
+        name="test observatory",
+        latitude=0 * u.deg,
+        longitude=0 * u.deg,
+        elevation=0 * u.m,
+    )
+
 
 # The fake image used for testing
 FAKE_CCD_IMAGE = FakeCCDImage(seed=SEED)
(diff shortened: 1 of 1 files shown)


stellarphot/photometry/tests/test_photometry.py
```diff
@@ -39,77 +39,107 @@
...
+        longitude=0 * u.deg,
+        elevation=0 * u.m,
+    )
 
-# Fake observatory location
-FAKE_OBS = Observatory(
-    name="test observatory", latitude=0 * u.deg, longitude=0 * u.deg, elevation=0 * u.m
-)
 
 # The fake image used for testing
```
JuanCab, line 98: Did you forget to migrate `FAKE_CCD_IMAGE` to be within a fixture or does it not work that way?

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -39,77 +39,107 @@
...
-    camera=FAKE_CAMERA,
-    observatory=FAKE_OBS,
-    photometry_apertures=DEFAULT_PHOTOMETRY_APERTURES,
-    source_location_settings=DEFAULT_SOURCE_LOCATIONS,
-    photometry_optional_settings=PHOTOMETRY_OPTIONS,
-    passband_map=PASSBAND_MAP,
-    logging_settings=DEFAULT_LOGGING_SETTINGS,
-)
+@pytest.fixture
+def photometry_settings(
```
JuanCab, line 126: I know its just a pytest fixture, but I thought it might make sense to call this `test_photometry_settings` since there are other places in the code where we have photometry settings getting set.

JuanCab (dismissed):
NO showstoppers.  I had a question and a suggested variable/fixture rename for clarity, BUT approved.

stellarphot/photometry/tests/test_photometry.py
mwcraig: How about `photometry_settings_for_test`? I'm not sure how well `pytest` will react if the name starts with `test`...

stellarphot/photometry/tests/test_photometry.py
mwcraig: That seemed OK for now since we have it in a function, but I can fixture-ize it if you want.

stellarphot/photometry/tests/test_photometry.py
JuanCab: Sounds good, I forgot pytest is going to look for particular function names.

stellarphot/photometry/tests/test_photometry.py
JuanCab: No need, just checking.

PR #423, opened 2024-08-10
Fix aggregation issues in calculation of AIJ relative fluxes

### LINKED ISSUE #422: Calculation of AIJ relative fluxes fails
A consequence of #421 is the calculation of `aij_relative_flux` fails because it relies on aggregation to calculate the relative flux.

mwcraig:
This fixes #422 by using the workaround suggested in https://github.com/astropy/astropy/issues/10944, converting a `QTable` to a `Table`, aggregating, and then converting back to `QTable`.

The first commit modifies the relevant test so that it fails to ensure that we are testing for this failure mode.


JuanCab (approved):
Man, what an annoyance, glad you figured it out.  Looks good as long as the tests pass.

Issue #188, opened 2023-12-02
Notebook 06 needs settings for stuff that is currently in the code


mwcraig:
These should be settings:

```python
# These affect the fitting that is done

#bin size in minutes
bin_size = 5 * u.min

# Keep the time of transit fixed?
keep_fixed_transit_time = True
transit_time_range = 60 * u.min

# Keep radius of planet fixed?

keep_fixed_radius_planet = False

# Keep radius of orbit fixed?

keep_fixed_radius_orbit = False

# Remove effects of airmas?
fit_airmass = False

# Remove effects of sky background?
fit_spp = False

# Remove effects of change in focus?
fit_width = False
```

PR #425, opened 2024-08-13
This PR rewrites to `io.tess.TOI` object using pydantic 

mwcraig:
Doing so has a couple of advantages:

+ The code is more compact
+ It is easy to serlialize to disk

Helps address, but does not complete fix the issue #188


pyproject.toml
```diff
@@ -101,6 +101,8 @@ include = [
         # Don't complain if tests don't hit assertions
         "raise AssertionError",
         "raise NotImplementedError",
+        # Don't complain about connection errors
+        "raise requests.ConnectionError",
         # Don't complain about script hooks
```
JuanCab, line 106: Is this to avoid errors during testing if the TESS servers are not available?

stellarphot/io/tess.py
JuanCab, line 1: Nice!  I had to read through it a bit, but pydantic classes do make some things a lot easier, like having properties.  That said, I wonder, can you now set the values of those properties by calling them or is pydantic preventing that.  Not a huge concern, but one nice feature of `@property` is allowing you to make read-only attributes... at the pydantic ones read-only once set?

JuanCab (approved):
I approve this, but would appreciate feedback on the questions.

pyproject.toml
mwcraig: Yes, and I wasn't sure how to test for that case...

stellarphot/io/tess.py
mwcraig: You can now set them, but pydantic has a way of making them non-settable: https://docs.pydantic.dev/latest/concepts/models/#faux-immutability

Lmk if you want me to do that

mwcraig: (merging for now, though, so I can open the next PR)

PR #428, opened 2024-08-14
Add test for #427 and fix it

mwcraig:
Fixes #427 by selecting just the first row on the table retrieved from MAST.

@JuanCab if this could get a quick review that would be great.

PR #429, opened 2024-08-14
Add the ability to generate the data needed to do photometry on a TESS object of interest

mwcraig:
There are a couple of specialized things a TESS observation needs. One is a an input source list of targets that must have photometry performed on them -- it is the set to check for eclipsing binaries if no event is seen in the primary target. The other is some information about the TOI

The new code here provides a way to generate both of those things, and there is a small notebook to generate the files via JupyerLab.

This also helps address #188 

PR #428 should get merged before this.


mwcraig:
Please ignore the first commit in this PR -- it is the commit in #428

stellarphot/settings/custom_widgets.py
```diff
@@ -1075,3 +1077,141 @@ def _confirmation(self, change=None):
...
+        self.drop_label = ipw.HTML("How would you like to specify the TIC ID?")
+        self.drop = ipw.Dropdown(
+            options=(("Type in TIC ID", 0), ("Choose an image", 1))
+        )
+        self.tic_id_entry = ipw.IntText(description="TIC ID")
+        self.fits_opener = FitsOpener()
+        self.confirm = Confirm()
+        self.spinner = Spinner(message="<h3>Downloading TIC info...</h3>")
+        self.all_done = ipw.HTML(
+            "<h2>Files have been written, you can close this notebook.</h2>"
```
JuanCab, line 1101: *Suggestion, not demand.* I would remove "you can close this notebook" since I could envision a notebook with this TESS selection at the top and other code below.  Seemed like "hard coding" the solution.  You could instead provide links to the files created to be viewed before proceeding.

stellarphot/settings/tests/test_tess_photometry_setup.py
```diff
@@ -0,0 +1,136 @@
...
+        # Confirmation box is displayed, click yes
+        widget.confirm._yes.click()
+
+        # The confirmation dialog should be hidden now
+        assert widget.confirm.layout.display == "none"
+
+        # The all done display should be visible
+        assert widget.all_done.layout.display == "flex"
+
+        # There should be exactly one json file from which we can create a TOI object
```
JuanCab, line 115: For testing this is fine, since you are in a temp directory, but we can't assume this normally, right?

JuanCab (approved):
I confirmed functionality of the notebook with a hand entered TIC number, didn't have a TIC FITS file to test.

Looks good overall, see by comments, but I approve this.  Feel free to consider my mild suggestion(s).

stellarphot/settings/tests/test_tess_photometry_setup.py
mwcraig: Correct -- I just didn't want to hard code the file name.

stellarphot/settings/custom_widgets.py
mwcraig: Fixed that, thanks.

PR #430, opened 2024-08-15
Exclude notebook files from ruff

mwcraig:
It turns out that ruff 0.6.0, released today now [examines notebooks by default](https://astral.sh/blog/ruff-v0.6.0#jupyter-notebooks-are-now-linted-and-formatted-by-default).

We may want to do that at some point but I would really prefer that today not be that day....

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,10 @@ include = '\.pyi?$|\.ipynb$'
 # )
 # '''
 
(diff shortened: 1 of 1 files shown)


JuanCab (approved):
I don't know... this was just so much code to review.  I'm just gonna approve it. ;)

Looks fine.

PR #431, opened 2024-08-17
Add sourcelist input option to comparison

mwcraig:
This is the last step needed to make exoplanet photometry as functional as it was in stellarphot 1.4.X.

Also it adds the spinner to a notebook.


JuanCab (approved):
Looks good as far as I can tell.  I had to dig into an astropy command I was less familiar with and understand a bit of why you split up `_set_file`, but it seems sound.

PR #440, opened 2024-09-08
Fix APASS passband names

### LINKED ISSUE #439: Current `apass_dr9` implementation does not change passbands to the AAVSO standard
The `apass_dr9` implementation does not change the passbands present in the Vizier catalog to the AAVSO standard names. It should.

mwcraig:
This PR ensures that the APASS passband names are the AAVSO standard names.

Fixes #439

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -9,7 +9,7 @@ from astropy.time import Time
 from astropy.wcs import WCS
 from astroquery.vizier import Vizier
 
-from .settings import Camera, Observatory
+from .settings import Camera, Observatory, PassbandMap
 
 __all__ = [
     "BaseEnhancedTable",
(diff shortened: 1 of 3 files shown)


mwcraig:
Made this a draft until I figure out what caused the test failure....

mwcraig:
...and I found the issue. I made the smallest possible change to get the tests to pass and open #441 to note the underlying problem.

mwcraig:
@JuanCab -- this is ready for review now

mwcraig:
Just added a test that hits the line that was not covered.

stellarphot/core.py
```diff
@@ -1017,6 +1032,21 @@ def apass_dr9(field_center, radius=1 * u.degree, clip_by_frame=False, padding=10
 
     # IAU says there is a space between the acronym and the coordinates.
     raw_catalog["id"] = [f"{designation_acronym} {coord}" for coord in coord_string]
+
+    # Translate the passbands to AAVSO standard names.
+    # No need to change B and V since those are already correct.
+    # Do this *after* initialization so that the original APASS band names
+    # are used for the tidy-ification operation.
+    raw_catalog.passband_map = PassbandMap(
```
JuanCab, line 1040: I assume Vizier only tracks these bandpasses, so you can hardcode this, right?

JuanCab (approved):
Looks good to me, and no typos as far as I can see.

stellarphot/core.py
mwcraig: Each catalog seems to be difference 🎉 🙄 but for APASS these are the names in Vizier. A new data release of APASS could conceivably change that, but this is explicitly DR9, so a new data release would mean adding a new class (which is fine I think).

PR #442, opened 2024-09-08
Add method to transform catalog to band-named columns

mwcraig:
This adds a method for "un-tidying" the data in a catalog to get table in which separate passbands are columns. This briefly made me wonder if going with a tidy format for the catalog was a good decision, but I think it was. This ensures some uniformity across catalogs, hopefully.

This PR is part of the prep for bringing back magnitude transforms.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -983,9 +983,9 @@ class CatalogData(BaseEnhancedTable):
         Notes
         -----
 
-        The column names in the output will be the passband names with "mag_" as a
+        The column names in the output will be the passband names with ``mag_`` as a
         prefix. An error column for each passband will be generated, with the prefix
-        "mag_error_". If the catalog already has columns with these names, they will
+        ``mag_error_``. If the catalog already has columns with these names, they will
         be overwritten. The input catalog will not be changed.
         """
         catalog_passbands = set(self["passband"])
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -963,6 +963,65 @@ class CatalogData(BaseEnhancedTable):
 
         return cat
 
+    def passband_columns(self, passbands=None):
+        """
+        Return a copy of the catalog with passbands as column names instead
+        of the default format, which has s single column for passbands.
+
+        Parameters
+        ----------
+        passbands : list, optional
+            List of passbands to include in the output. If not provided, all
+            passbands in the catalog will be included.
+
+        Returns
+        -------
+        `stellarphot.CatalogData`
+            Table of catalog information with passbands as column names. See Notes below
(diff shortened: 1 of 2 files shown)


stellarphot/core.py
```diff
@@ -963,6 +963,65 @@ def from_vizier(
...
+
+        Notes
+        -----
+
+        The column names in the output will be the passband names with ``mag_`` as a
+        prefix. An error column for each passband will be generated, with the prefix
+        ``mag_error_``. If the catalog already has columns with these names, they will
+        be overwritten. The input catalog will not be changed.
+        """
+        catalog_passbands = set(self["passband"])
```
JuanCab, line 991: Finally, a use for `set` that isn't students trying to de-duplicate items in a list by looking up answer in Stack Overflow.

stellarphot/core.py
```diff
@@ -963,6 +963,65 @@ def from_vizier(
...
+        # This makes a MultiIndex for the columns -- "mag" and "mag_error" are the
+        # top level, and the passbands are the second level.
+        df = df.pivot(
+            columns="passband", index=["id", "ra", "dec"], values=["mag", "mag_error"]
+        )
+
+        # The column names are a MultiIndex, so we flatten them to either "mag_band"
+        # or "mag_error_band", where "band" is the passband name.
+        df.columns = df.columns.to_series().str.join("_")
+
```
JuanCab, line 1018: I had to stare at this a bit to see what it is doing.  I *think* I understand it now.  Your creating a pivoted table that lists magnitude and magnitude error with passband a a column, then using that to build mag_band and magerr_band columns... Still a bit of head scratcher for me, but I understand it has to be done this way.

stellarphot/tests/test_core.py
```diff
@@ -839,6 +839,54 @@ def test_tidy_vizier_catalog_several_mags():
...
+    for pb in passbands:
+        assert f"mag_{pb}" in new_cat.colnames
+        assert f"mag_error_{pb}" in new_cat.colnames
+
+    # Check the data
+    assert set(new_cat["mag_V"]) == set(apass_input["Vmag"])
+    assert set(new_cat["mag_error_V"]) == set(apass_input["e_Vmag"])
+    assert set(new_cat["mag_SI"]) == set(apass_input["i_mag"])
+    assert set(new_cat["mag_error_SI"]) == set(apass_input["e_i_mag"])
+
```
JuanCab, line 881: Seems like a reasonable way to check that the new catalog format holds the same data as the original.

stellarphot/core.py
mwcraig: 🤣

PR #446, opened 2024-09-13
Rename workflow file

### LINKED ISSUE #403: Rename github workflow for test
Right now the workflow that does the testing is named `build.yml`. A name like `test.yml` would be more descriptive of what the workflow actually does.

mwcraig:
This fixes #403 by renaming the test file.


mwcraig:
Self-merging, since this is trivial...

PR #457, opened 2024-10-06
Update minimum required version of jupyter-app-launcher

mwcraig:
This PR requires a recent version of `jupyter-app-launcher` that fixes a bug in which two notebooks opened from the launcher in succession connected to the same kernel instead of different kernels.


mwcraig:
(planning to self-merge once the lint/pre-commit checks pass since the changes no code)

PR #459, opened 2024-10-06
Add an error column for catalogs, even those without error

### LINKED ISSUE #438: `CatalogData` should explicitly allow for an error column
I've been working a bit on magnitude transforms and realized that the current definition of CatalogData does not allo for a magnitude error column.

While an error column probably should not be required, it should exist where possible.

mwcraig:
Adds a mandatory magnitude error column for `CatalogData`. Since some catalogs (e.g. VSX) do not supply an error, a new option was added to `__init__` that sets the error column to all `NaNs`.

Fixes #438 

@JuanCab I'm not sure this is the best solution. I coded up a version where to indicate a catalog had no error available by adding an entry `None: "mag_error"` to `colname_map`, but think I prefer this more explicit version now.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -613,6 +613,10 @@ class CatalogData(BaseEnhancedTable):
         the documentation for `stellarphot.settings.PassbandMap` for more
         information. The object behaves like a dictionary when accessing it.
 
+    no_catalog_error: bool, optional (Default: False)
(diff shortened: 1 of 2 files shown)


JuanCab (approved):
Looks quite reasonable as an approach, all I caught was a typo.  Looks like the tests just updated the old tests, which is fine.

PR #461, opened 2024-10-07
Add ability to calculate BJD from a single coordinate

### LINKED ISSUE #444: BJD calculation is per star, not per frame
I noticed yesterday that the BJD calculation, which is done in [`PhotometryData`](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/core.py#L260), has an issue.

In the past, `BJD_TDB` was calculated using the center of the frame as the RA/Dec. Now it is being calculated for each star individually in [`add_bjd_column`](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/core.py#L260), which results with each star in a single image having a slightly different BJD. 

While at some level that is maybe the correct behavior (or at least not incorrect behavior), since BJD does depend on RA and Dec, it is not usually what one wants. Instead, having a single BJD for each image is more useful.

I have not yet looked into how large a field of view one would need to have for there to be meaningful difference in BJD across the image, but I'm guessing it is tens of degrees.

@JuanCab -- what do you think about this? Maybe meet up to hash out how to handle it?

mwcraig:
This PR fixes #444 by doing two things:

1. No longer calculating BJD on creation
2. Adding an argument that allows the coordinates for the correction to be set.

I started out by not doing 1 -- that is, by adding a new parameter to `PhotometryData.__init__` but realized that meant, eventually, changing or adding a photometry setting, which I wasn't too enthusiastic about, so I went this route instead.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -353,11 +353,10 @@ class PhotometryData(BaseEnhancedTable):
     In addition to these required columns, the following columns are created based
     on the input data during creation.
 
-    bjd    (only if ra and dec are all real numbers, otherwise set to np.nan)
     night
 
     If these computed columns already exist in `data` class the class
-    will throw an error a ValueError UNLESS`ignore_computed=True`
+    will throw an error a `ValueError` UNLESS ``ignore_computed=True``
     is passed to the initializer, in which case the columns will be
     retained and not replaced with the computed values.
     """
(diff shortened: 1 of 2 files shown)


stellarphot/core.py
```diff
@@ -353,11 +353,10 @@ class PhotometryData(BaseEnhancedTable):
     In addition to these required columns, the following columns are created based
     on the input data during creation.
 
-    bjd    (only if ra and dec are all real numbers, otherwise set to np.nan)
     night
 
     If these computed columns already exist in `data` class the class
-    will throw an error a ValueError UNLESS`ignore_computed=True`
+    will throw an error a `ValueError` UNLESS ``ignore_computed=True``
```
JuanCab, line 359: Why the double backticks?

mwcraig: I think single backticks is supposed to be a reference and double means format it as code. Not sure why that wasn't an error before, but maybe it was pointing to the attribute?

PR #464, opened 2024-10-07
Add method for creating lightcurve for object

### LINKED ISSUE #462: Add a conveniences method for `PhotometryData`
Adding a `to_lightcurve` method that makes a [LightCurve](https://lightkurve.github.io/lightkurve/reference/api/lightkurve.LightCurve.html#lightkurve.LightCurve) object from the [`lightkurve` package](https://lightkurve.github.io/lightkurve/quickstart.html) would make `PhotometryData` much more useful. Those objects have methods for performing a periodogram (Lomb-Scargle or BLS), phase folding, plotting, etc. So lots of extra functionality without much effort on our part. A `LightKurve` object is a subclass of astropy's `TimeSeries`, so we get both for free.

mwcraig:
This fixes #462 by adding a `lightcurve_for` method to `PhotometryData`.

A couple items:

1. A better API might be to accept a single required argument, maybe called "target", that can be either a `star_id`, a coordinate or a name. If that is better, then we might need to drop the name option, because star_id is a string so I'm not sure how to distinguish between star_id and name.
2. The term "flux" is a little unfortunate, but we can't change that...it is a lightkiurve thing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -194,5 +194,7 @@ filterwarnings = [
     # papermill is using deprecated jupyter paths
     'ignore:Jupyter is migrating its paths to use standard platformdirs:DeprecationWarning',
     # papermill is also using a deprecated method of getting current time
-    'ignore:.*is deprecated and scheduled for removal.*:DeprecationWarning'
+    'ignore:.*is deprecated and scheduled for removal.*:DeprecationWarning',
+    # lightkurve is using a deprecated numpy interface
+    'ignore:.*numpy.core.einsumfunc is deprecated and has been renamed.*:DeprecationWarning'
 ]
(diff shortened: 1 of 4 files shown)


JuanCab (approved):
Seems reasonable overall.  The tests looked reasonable as well (I mean, you are just checking the right number of rows are in the final light curve).

mwcraig:
> A better API might be to accept a single required argument, maybe called "target", that can be either a star_id, a coordinate or a name. 

@JuanCab  what do you think about this 👆

JuanCab:
> > A better API might be to accept a single required argument, maybe called "target", that can be either a star_id, a coordinate or a name.
> 
> @JuanCab what do you think about this 👆

Yeah, I don't know if the technical term for this is API (we aren't connecting different programs together via some server/client interface), BUT I do think it make sense to have a single required argument, especially if the only difference is the data type.

mwcraig:
Merging, since the new signature matches the proposal above (darn you for learning what things like API actually mean 😉 ) and test coverage is 100% and tests are passing.

PR #466, opened 2024-10-07
Move dictionaries needed for testing to new file

### LINKED ISSUE #465: Running stellarphot now requires that pytest be installed 🤦‍♂️
The problem was introduced in https://github.com/feder-observatory/stellarphot/pull/454 when the code was modified to import some sample photometry settings from the test files, which require pytest.

It was not caught in testing because all of our tests have pytest installed.

Fix incoming.

mwcraig:
This ensures that we can get at those constants without having to have test dependencies (e.g. pytest) installed.

Fixes #465

### DIFF
--- a/stellarphot/gui_tools/tests/test_seeing_profile.py
+++ b/stellarphot/gui_tools/tests/test_seeing_profile.py
@@ -25,7 +25,7 @@ from stellarphot.settings import (
     PhotometryWorkingDirSettings,
     settings_files,  # This import is needed for mocking
 )
-from stellarphot.settings.tests.test_models import (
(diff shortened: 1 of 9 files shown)


JuanCab (approved):
Looks simple and good.  Just curious, why the move to settings.constants if (I assume) these dicts are only used for testing.  I see the stated reason about wanting to access without pytest, but I assume that only has to do with the fact there are pytest commands in the tests.test_models file... why change the directory as well?

mwcraig:
I think, though I haven't checked, that putting it in test might have required the test imports? Not actually positive about that. I was inspired by a similar thing in ipyautoui...

PR #468, opened 2024-10-10
Revive transforms

mwcraig:
This PR gets magnitude transforms working again with the same level of functionality they had in stellarphot 1. There are still several issues with how this is done, but the intent here is to simply have some working transform code, not to improve it yet.


mwcraig:
Merging even though the test coverage is technically down...the modified code was not covered by tests before and still isn't....

PR #470, opened 2024-10-12
Add test and fix for PhotometryData initialization

### LINKED ISSUE #469: `PhotometryData` does not use column name mapping
It turns out that `PhotometryData` does not used the remapped column names in its `__init__`. I'll open a PR with a test that reproduces the issue and fixes it in a moment.

mwcraig:
Fixes #469 by using `self` to get at the data instead of `input_data` after `BaseEnhancedTable` has fixed up column names.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -438,6 +438,9 @@ class PhotometryData(BaseEnhancedTable):
                 **kwargs,
             )
 
+            # From this point forwarsd we should be using self to get at any data
+            # columns, because that is where BaseEnhancedTable has put the data.
+
             # Perform input validation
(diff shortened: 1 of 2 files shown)


stellarphot/core.py
```diff
@@ -438,6 +438,9 @@ def __init__(
                 **kwargs,
             )
 
+            # From this point forwarsd we should be using self to get at any data
```
JuanCab, line 441:
```suggestion
            # From this point forward we should be using self to get at any data
```
Type

JuanCab (approved):
Looks good, I think this may have been a design error of my origin, sorry!  Glad you fixed it.

PR #471, opened 2024-10-12
Add tool for migrating stellarphot v1 data to stellarphot v2

mwcraig:
This adds a way to update v1 data so that it can be used in v2. This will help remove the need to redo a bunch of photometry while allowing users to take advantage of the new stuff in v2.

*NOTE:* #470 should be merged first, because that fix was necessary for this PR to work.


stellarphot/core.py
```diff
@@ -438,6 +438,9 @@ def __init__(
                 **kwargs,
             )
 
+            # From this point forwarsd we should be using self to get at any data
```
JuanCab, line 441:
```suggestion
            # From this point forward we should be using self to get at any data
```
Same typo noted in #470

stellarphot/utils/tests/test_version_migrator.py
```diff
@@ -0,0 +1,104 @@
...
+        with pytest.raises(ValueError, match="Can only migrate from a lower version"):
+            VersionMigrator(from_version="2", to_version="1")
+
+        # Nothing to do if the versions are the same
+        with pytest.raises(ValueError, match="Can only migrate from a lower version"):
+            VersionMigrator(from_version="2", to_version="2")
+
+        # An invalid version should raise an error
+        with pytest.raises(InvalidVersion, match="notta parsable version"):
+            VersionMigrator(from_version="notta parsable version")
```
JuanCab, line 55:
```suggestion
        with pytest.raises(InvalidVersion, match="not a parsable version"):
            VersionMigrator(from_version="not a parsable version")
```

stellarphot/conftest.py
JuanCab, line 1: A little confused why you created a fixture for this rather than just creating a new and old version of the testing file... it it just to allow iteration of test cases via the fixture?

stellarphot/utils/tests/test_version_migrator.py
JuanCab, line 1: Looks like a through test suite as far as I can tell, kind of explains the weird fixture I noted earlier.

stellarphot/utils/version_migrator.py
```diff
@@ -0,0 +1,174 @@
...
+            "xcenter": u.pixel,
+            "ycenter": u.pixel,
+            "exposure": u.second,
+        }
+        for un in unitify:
+            # Note that this deliberately discards any unit that might have been
+            # present in the original data. This is a stellarphot-specific migration
+            # tool, not something more general, and the choices above are the correct
+            # units for the data.
+            new_data[un] = new_data[un].value * unitify[un]
```
JuanCab, line 136: Could we potentially have unit conversion issues here?  Or were these units also used in v1 but just not attached to the columns?  I don't remember.

stellarphot/utils/version_migrator.py
```diff
@@ -0,0 +1,174 @@
...
+            "aperture_net_flux": "aperture_net_cnts",
+            "BJD": "bjd",
+            "noise": "noise_cnts",
+            "noise-aij": "noise_electrons",
+            "filter": "passband",
+            "RA": "ra",
+            "Dec": "dec",
+        }
+
+        # THe new night calculation is better, so drop the old one
```
JuanCab, line 150:
```suggestion
        # The new night calculation is better, so drop the old one
```

JuanCab (dismissed):
Looks good... At least the column renaming and so forth.  I am curious about units (since no explicit unit conversion done).  See my notes.  Approve, but would appreciate feedback and typo corrections.

stellarphot/utils/version_migrator.py
mwcraig: The old almost never included units, but it depended on whether the photometry was saved as CSV (no units saved) or FITS (units sometimes present for some things). This is maybe a little risky, but for feder data the image unit was always ADU, times, in seconds, etc.

No one else ever used v1 as far as I know, so this should be ok, I think

stellarphot/utils/tests/test_version_migrator.py
mwcraig: Yeah, was using it a few different times

stellarphot/core.py
mwcraig: Was fixed there

PR #472, opened 2024-10-12
Add passband option for lightcurve

mwcraig:
I realized that `lightcurve_for` really needs a passband argument in case there are multiple passbands in the data.


JuanCab (approved):
Looks good, good catch.

PR #476, opened 2024-10-27
Change ID field in apass_dr9 catalog

mwcraig:
Looks like Vizier dropped the Vizier-generated ID for APASS entries. This uses the field number in stead, which is not unique to each star, but we generate our own unique ID anyway.

Going to YOLO this since it is blocking progress for some students working on AAVSO posters....


mwcraig:
🤬 no idea why profile code broke. Tracking that down....

mwcraig:
> 🤬 no idea why profile code broke. Tracking that down....

The release of photutils 2 broke our tests. See #477 opened to track this issue. Also #478 (still a draft) that would add a test against dev versions of photutils and add a cron job that runs weekly to do our tests.

PR #478, opened 2024-10-27
Add a github cron job

mwcraig:
We just got bit by a bug caused by a new release of photutils. Adding a cron job, which will include at least one test with the latest version of dependencies, should help catch this kind of thing in the future.

This does need to have one job added in tox, I think, that unpins photutils, or maybe runs against a dev version of photutils, before this can be merged


mwcraig:
Yes, that would help catch things like the change in photutils...

JuanCab (approved):
OK, makes sense.

PR #494, opened 2025-04-07
Change default value for reject_unmatched and fix description

mwcraig:
The description fix here is pretty straightforward. We are preparing to do some notebook-based photometry on several nights and want to keep all the stars we can so I changed the default for that case.


JuanCab (approved):
Looks good, although the "Blah blah" description originally there was compelling.

PR #497, opened 2025-04-11
Correctly calculate FWHM

### LINKED ISSUE #490: FWHM computation is incorrect in all cases
There are two options for computing FWHM, either by fitting or by image moments. The image moment one always returns a goofy number, and the fit always fails. 

So that needs to be fixed.

For the fit, the issue is that the compound model fit fails. A simple gaussian-only works fine.

mwcraig:
This PR fixes #490 by dropping the "moments" method of calculating FWHM which is just plain wrong.

I ended up changing the name of the settings `fwhm_by_fit` to `fwhm_method`.

Along the way I realized that several tests were inadvertantly changing some of the settings "constants" we use so I'm mostly `deepcopy`ing them now.


pyproject.toml
```diff
@@ -196,5 +196,7 @@ filterwarnings = [
     # papermill is also using a deprecated method of getting current time
     'ignore:.*is deprecated and scheduled for removal.*:DeprecationWarning',
     # lightkurve is using a deprecated numpy interface
-    'ignore:.*numpy.core.einsumfunc is deprecated and has been renamed.*:DeprecationWarning'
+    'ignore:.*numpy.core.einsumfunc is deprecated and has been renamed.*:DeprecationWarning',
+    # Sometimes gaussian fits are not great
```
JuanCab, line 200: A little hesitant to ignore the warning about the bad fit without feedback to the user of some sort.

stellarphot/photometry/source_detection.py
```diff
@@ -111,13 +70,31 @@ def compute_fwhm(
         The FWHM of each source in the x and y directions.
 
     """
+    if sky_per_pix_avg is not None and sky_per_pix_column is not None:
```
JuanCab, line 73: I'm getting a codecov error that L74 is not covered.  Can we trigger this error intentionally in the tests?

stellarphot/photometry/source_detection.py
```diff
@@ -111,13 +70,31 @@ def compute_fwhm(
...
+        raise ValueError(
+            "Cannot specify both `sky_per_pix_avg` and `sky_per_pix_column`."
+        )
+
+    if sky_per_pix_column is not None:
```
JuanCab, line 78: These next two if conditionals apparently are also not covered in tests.

stellarphot/photometry/source_detection.py
```diff
@@ -111,13 +70,31 @@ def compute_fwhm(
...
+    elif sky_per_pix_column is None:
+        # User didn't give a value for the sky background to subtract
+        # so try an of estimate it from the image
+
+        sky_values = [np.nanmedian(ccd.data)] * len(sources)
 
     fwhm_x = []
     fwhm_y = []
-    for source in sources:
+    for source, sky in zip(sources, sky_values, strict=True):
```
JuanCab, line 94: If you have strict=True, you might want to wrap this for loop in a try/except block to handle ValueError crashes, or are you handling that when calling the function?

stellarphot/photometry/source_detection.py
```diff
@@ -111,13 +70,31 @@ def compute_fwhm(
...
         x = source[x_column]
```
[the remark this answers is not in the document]
JuanCab: This suggestion from Copilot is irrelevant since we are requiring a relatively recent Python.

stellarphot/photometry/source_detection.py
```diff
@@ -127,29 +104,56 @@ def compute_fwhm(
         except AttributeError:
             pass
 
-        cutout = Cutout2D(ccd.data, (x, y), 5 * fwhm_estimate)
-        if fit:
-            fit = _fit_2dgaussian(cutout.data)
-            fwhm_x.append(gaussian_sigma_to_fwhm * fit.x_stddev_1)
-            fwhm_y.append(gaussian_sigma_to_fwhm * fit.y_stddev_1)
-            # print('Still fitting!!')
+        # SKY SUBTRACT SHIT!!
```
JuanCab, line 107: While one of the milder swears, doesn't seem to particularly be necessary, was this really that hard. :)

stellarphot/photometry/source_detection.py
```diff
@@ -127,29 +104,56 @@ def compute_fwhm(
...
+        match fit_method:
+            case FwhmMethods.FIT:
+                # Make sure we get an odd fits shape
+                fit_shape = int(2 * ((5 * fwhm_estimate) // 2) + 1)
+
+                # fit_fwhm is supposed to handle NaNs automatically but it doesn't
+                # as of photutils 2.2.0
+                # see https://github.com/astropy/photutils/issues/2029
+                # For now replace any NaN with zero and hope for the best.
+                cutout.data[nan_mask] = 0
```
JuanCab, line 129: Given these are masked, isn't there a way to replace with the mean of the surrounding pixels?  Kinda surprised, as that would seem that a safer approach.

I did a quick search of the astropy documentation and found 
```
from astropy.convolution import interpolate_replace_nans
result = interpolate_replace_nans(image, kernel)
```
is supposed to use convolution to replace NaNs, that might be a safer approach.

stellarphot/settings/models.py
```diff
@@ -718,14 +731,28 @@ class PhotometryOptionalSettings(BaseModelWithTableRep):
...
     model_config = ConfigDict(
         json_schema_extra=dict(
             description=_extract_short_description(__doc__),
         )
     )
 
+    # Made this a static method to define inside this class, since this is the only
+    # class that needs to handle this.
+    @staticmethod
+    def _validate_fwh_method(value: Any) -> Any:
```
JuanCab, line 745: Maybe rename method 
```suggestion
    def _validate_fwhm_method(value: Any) -> Any:
```

stellarphot/settings/tests/test_custom_widgets.py
```diff
@@ -663,7 +670,7 @@ def test_details_hideable_plays_nicely_with_new_item(self):
 
         # Making a new camera and saving it should bring back, and respect,
         # the hideable details
-        new_camera = TEST_CAMERA_VALUES.copy()
+        new_camera = TEST_CAMERA_VALUES
```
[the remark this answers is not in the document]
JuanCab: This seems to make sense unless you make sure not to alter `new_camera`

stellarphot/settings/tests/test_models.py
```diff
@@ -372,7 +425,7 @@ def test_camera_unitsless_gain():
 
 
 def test_camera_incompatible_max_val_units():
-    camera_for_test = TEST_CAMERA_VALUES.copy()
+    camera_for_test = TEST_CAMERA_VALUES
     # data unit is adu, not count
```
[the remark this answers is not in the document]
JuanCab: Again, looks like a reasonable suggestion.

JuanCab (approved):
A lot of complaints about code not being covered by tests here.  I flagged the first few, but really might need to expand the tests a bit.  Other comments in the code.  Biggest issues is I think replacing NaN with zero makes less sense than replacing with the mean of the surrounding pixels.

JuanCab (changes requested):
A lot of complaints about code not being covered by tests here.  I flagged the first few, but really might need to expand the tests a bit.  Other comments in the code.  Biggest issues is I think replacing NaN with zero makes less sense than replacing with the mean of the surrounding pixels.

JuanCab (changes requested):
A lot of complaints about code not being covered by tests here.  I flagged the first few, but really might need to expand the tests a bit.  Other comments in the code.  Biggest issues is I think replacing NaN with zero makes less sense than replacing with the mean of the surrounding pixels.

pyproject.toml
mwcraig: These warnings are only ignored during tests; when the code is running as normal this warning would be raised.

stellarphot/photometry/source_detection.py
mwcraig: Yeah, I should be get all of the uncovered lines taken care of

stellarphot/photometry/source_detection.py
mwcraig: Right now I don't think it is getting handled at all; I'm not sure special handling is needed since this is likely an error in the user's data, not an error in what we do with the data. 

I do need to bump the minimum version of Python since `strict` was introduced in 3.11....

stellarphot/photometry/source_detection.py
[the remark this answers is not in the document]
mwcraig: I thought so too, but it turns out we claim to support Python 3.10 I'll fix that by bumping the minimum to 3.11

stellarphot/photometry/source_detection.py
mwcraig: You have no idea....but I'll take it out.

stellarphot/photometry/source_detection.py
mwcraig: At a minimum it can't hurt to do that unless it winds up being slow. The bug has been fixe in photutils and there it is simply ignored.

stellarphot/settings/tests/test_custom_widgets.py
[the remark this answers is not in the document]
mwcraig: Yeah, I don't think I can claim that to be the case. Turns out tests were modifying this input dictionary all over the place.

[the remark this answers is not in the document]
JuanCab: Would it make sense to write a function that always returns copies of all the needed constants?  Maybe not, since ".copy()" seems just as clear... (sorry, just thinking out loud or at least with loud clacking on my keyboard).

stellarphot/photometry/source_detection.py
mwcraig: These are covered now, hopefully the codecov report updates///

stellarphot/photometry/source_detection.py
mwcraig: Bumped the minimum python version...

stellarphot/photometry/source_detection.py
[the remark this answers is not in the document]
mwcraig: Fixed by updating the minimum python version

stellarphot/settings/tests/test_custom_widgets.py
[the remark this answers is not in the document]
mwcraig: Yeah, a pytest fixture to make sure each test gets its own copy of these would be the way to go eventually.

stellarphot/photometry/source_detection.py
mwcraig: I'm leaning against this for now because photutils will, once the next version is out, ignore this value anyway.

mwcraig:
Once again, I do not know what is up with codecov. THe new lines are now almost all covered....

stellarphot/settings/tests/test_models.py
```diff
@@ -34,6 +36,17 @@
...
+# NOTE: ALWAYS USE DEEPCOPY ON THESE DICTS SO TESTS DO NOT MODIFY THEM
+TEST_CAMERA_VALUES = deepcopy(TEST_CAMERA_VALUES)
+TEST_APERTURE_SETTINGS = deepcopy(TEST_APERTURE_SETTINGS)
+TEST_EXOPLANET_SETTINGS = deepcopy(TEST_EXOPLANET_SETTINGS)
+TEST_OBSERVATORY_SETTINGS = deepcopy(TEST_OBSERVATORY_SETTINGS)
+TEST_PHOTOMETRY_OPTIONS = deepcopy(TEST_PHOTOMETRY_OPTIONS)
+TEST_PASSBAND_MAP = deepcopy(TEST_PASSBAND_MAP)
+TEST_PHOTOMETRY_SETTINGS = deepcopy(TEST_PHOTOMETRY_SETTINGS)
+TEST_LOGGING_SETTINGS = deepcopy(TEST_LOGGING_SETTINGS)
+TEST_SOURCE_LOCATION_SETTINGS = deepcopy(TEST_SOURCE_LOCATION_SETTINGS)
```
JuanCab, line 48: I have to admit, looking at this, it looks like you are deep copying the originals but to the same variables.  I hadn't thought about this, but it makes sense that deep copy() will generate a new object that you assign to the old variable name... nice way to keep the old code working and still not change the constants.

stellarphot/settings/tests/test_models.py
```diff
@@ -213,8 +226,8 @@ def test_name_unicode_is_ok(self, model, settings):
 @pytest.mark.parametrize(
     "model,settings",
     [
-        [Camera, TEST_CAMERA_VALUES.copy()],
-        [Observatory, TEST_OBSERVATORY_SETTINGS.copy()],
+        [Camera, TEST_CAMERA_VALUES],
+        [Observatory, TEST_OBSERVATORY_SETTINGS],
```
JuanCab, line 230: I'm assuming you are using the deep copies you created in lines 39-48 here so you are not copying them.  Maybe?

stellarphot/settings/tests/test_models.py
```diff
@@ -284,6 +297,46 @@ def test_example(self, model, settings):
...
+        (
+            [True, FwhmMethods.FIT],
+            [False, FwhmMethods.PROFILE],
+        ),
+    )
+    def test_2_0_0a11(self, old_setting, new_setting):
+        """
+        This version had fwhm_by_fit, which has now changed to fwhm_method.
+        """
+        assert "fwhm_method" in TEST_PHOTOMETRY_OPTIONS
```
JuanCab, line 323: Does line 323 need to be repeated in line 325?

JuanCab (approved):
I had some comments, but it seems to look good.

stellarphot/settings/tests/test_models.py
mwcraig: yep

stellarphot/settings/tests/test_models.py
mwcraig: Technically neither is needed...they are leftover from when I was trying to track down where `fwhm_method` was being changed. I'm going to merge this but will drop those lines in the next PR.

PR #500, opened 2025-04-14
Add hack for catalog filter names

mwcraig:
I'll admit this is awful but the range of filter names in our data is also awful....


mwcraig:
Sorry @JuanCab -- needed to get this working on jupyter3....

PR #503, opened 2025-04-16
Remove tests that are duplicated in test class

mwcraig:
No idea why these were not removed when the tests were mostly rearranged into classes....


JuanCab (approved):
I looked this over, I have to agree, no clue why this wasn't caught before, glad it is resolved.

PR #504, opened 2025-04-19
Refactor `source_detection` to allow it to run faster

mwcraig:
This makes some relatively minor changes to `source_detection` that allow it to run significantly faster and makes the `print` stuff optional. The changes to the functionality are listed below and occur only if the user sets `stddev` to a value.

+ One change is to allow the user to input a `stddev` that is used to calculate the threshold for source detection (the `threshold` parameter is multiplied by `stddev`). 
+ Another significant change is to use the median image value as the sky background if i) no input sky value has been provided and ii) sigma clipped stats have not been calculated.
+ Finally, if sigma clipped stats have been done, and the user has not input a sky value, then the sigma clipped median is used instead of the mean.
 
It turns out that on Feder images, sigma clipped stats take roughly a second, even though you get a good estimate of the standard deviation just by using the noise level of the camera. I think the standard deviation of the background is really the quadrature sum of the camera noise and the Poisson noise of the background, but the noise alone isn't terrible.

A median estimate of the background is also not terrible.

### DIFF
--- a/stellarphot/photometry/source_detection.py
+++ b/stellarphot/photometry/source_detection.py
@@ -164,9 +164,11 @@ def source_detection(
     sigma=3.0,
     iters=5,
     threshold=10.0,
+    stddev=None,
     find_fwhm=True,
     sky_per_pix_avg=0,
     padding=0,
+    verbose=False,
 ):
     """
     Returns an SourceListData object containing the position of sources
@@ -195,6 +197,11 @@ def source_detection(
     threshold : float, optional. (default=10.0)
         The number of standard deviations above which to select sources.
 
+    stddev : float, optional
+        If provided, this value will be used as the standard deviation
+        of the image.  If not provided, the standard deviation will be
+        calculated using `sigma_clipped_stats` on the image.
+
     find_fwhm : bool, optional (default=True)
         If ``True``, estimate the FWHM of each source by fitting a 2D Gaussian
         to it. If ``False``, the 'fwhm_x', 'fwhm_y', and 'width' columns of
@@ -208,6 +215,10 @@ def source_detection(
     padding : int, optional (default=0)
         Distance from the edge of the image to ignore when searching for
         sources.
+
+    verbose : bool, optional
+        If ``True``, print additional information about the source
+        detection process.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


mwcraig:
Copilot makes an interesting point here:

> The default value of sky_per_pix_avg is set to 0 in the function signature, but this condition checks for None. This branch may never execute as intended; consider updating the default value to None if you plan to auto-calculate the sky background when no value is provided.

I think the current behavior is deliberate -- if you don't do anything then 0 is used as a sky value, making the auto-background estimation opt-in.

My leaning at the moment is to change the default to know and then the savvy user can opt in to using 0, so the foot-gun would still be available

mwcraig:
@JuanCab -- this should be at 100% test coverage in `source_detection.py`.

stellarphot/photometry/source_detection.py
```diff
@@ -228,34 +241,47 @@ def source_detection(
...
+        median = None
+
     if sky_per_pix_avg is None:
-        sky_per_pix_avg = mean
-        print(f"source_detection: sky_per_pix_avg set to {sky_per_pix_avg:.4f}")
+        if median is not None:
+            # We calculated sigma clipped stats, so use them
+            sky_per_pix_avg = median
+        else:
+            # A median is a pretty good estimate of the sky background, so use it
```
JuanCab, line 266: Good to spell this out!

stellarphot/photometry/tests/test_detection.py
```diff
@@ -71,11 +71,27 @@ def test_compute_fwhm_with_missing_data(mask_by_nan):
...
+def test_detect_source_number_location(
+    provide_spp_input, spp_unit, stddev_input, find_fwhm, image_is_ccd
+):
     """
     Make sure we detect the sources in the input table....
     """
-    fake_image = FakeImage(seed=SEED)
+    # Skip some cmobinations we really don't need to test
+    if not provide_spp_input and spp_unit is not None:
+        pytest.skip("No point in testing this combination")
```
JuanCab, line 87: Cute.

JuanCab (approved):
Other than a bunch of code coverage warnings, looks good.

PR #505, opened 2025-04-20
Allow the photometry aperture to be a multiple of the FWHM in each image

mwcraig:
The title says most of it, but a couple of points that will not be obvious from the code itself that should be considered:

+ The "gap" and "width" of the annulus are currently still in pixels even though the "radius" is a multiplier of the FWHM iv `variable_aperture` is True. So, for example, if `variable_aperture` is `True`, and the fwhm is 5 and the aperture radius is 1.5, the radius in pixels is 7.5. **The key question here is whether gap and annulus should be pixels or something else.**
+ I'm testing all of the new code, but the variable aperture test only checks that the FWHM that was put into the three generated test images is correct, and that the correct aperture radius is used. I think that is fine, since aperture photometry itself is checked elsewhere.


mwcraig:
Not sure why codecov is so far off on coverage (it is 100% in source detection), but rebased to see what happens.

stellarphot/gui_tools/seeing_profile_functions.py
```diff
@@ -574,6 +574,8 @@ def _update_plots(self):
         if self.rad_prof is None:
             return
 
+        fwhm = 2 * self.rad_prof.HWHM
```
JuanCab, line 577: So here you are setting the FWHM to 2*HWHM, what did you use previously?  Or I guess, based on line 608 that you are replacing putting the cutoff radius as a vertical line in the plot and making it FWHM?   Might want to clarify this in comments.

stellarphot/photometry/photometry.py
```diff
@@ -380,7 +393,11 @@ def single_image_photometry(
     if use_coordinates == "sky":
         try:
             xcen, ycen = centroid_sources(
-                ccd_image.data, xs, ys, box_size=2 * photometry_apertures.radius + 1
+                # TODO: FIX THIS
```
JuanCab, line 396: Is this a reminder that you need finish something here?

stellarphot/photometry/photometry.py
```diff
@@ -380,7 +393,11 @@ def single_image_photometry(
...
+                ccd_image.data,
+                xs,
+                ys,
```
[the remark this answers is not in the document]
JuanCab: Funny, the AI caught this as well...

stellarphot/photometry/source_detection.py
```diff
@@ -361,3 +364,178 @@ def source_detection(
...
+    if isinstance(ccd, CCDData):
+        data = ccd.data
+        mask = ccd.mask
+    else:
+        data = ccd
+        mask = None
+
+    with warnings.catch_warnings():
+        # block_reduce generates some warnings about things like unit, wcs, etc
+        # that are set on ccd buut not preserved in the reduced image.
```
JuanCab, line 440: Typo
```suggestion
        # that are set on ccd but not preserved in the reduced image.
```

stellarphot/photometry/source_detection.py
```diff
@@ -361,3 +364,178 @@ def source_detection(
...
+    # flux by the area of the block.  If this is greater than max_adu,
+    # then the sources are too bright.
+    too_bright = block_sources["peak"] / block_size**2 > max_adu
+
+    # Drop the sources that are too bright
+    block_sources = block_sources[~too_bright]
+
+    # Only keep the n_brightest sources (eventually)
+    n_brightest = min(len(block_sources), int(n_brightest_sources))
+    # Pad n_bnrightest a bit in case some of these end up being too bright
```
JuanCab, line 474:
```suggestion
    # Pad n_brightest a bit in case some of these end up being too bright

File: stellarphot/photometry/source_detection.py
Comment (JuanCab) [inline stellarphot/photometry/source_detection.py:1]: I think I understood most of what you were doing here, looks reasonable.

File: stellarphot/photometry/tests/fake_image.py [hunk 6]
```diff
@@ -24,10 +25,46 @@ class FakeImage:
         the seed will be randomly generated.
```
JuanCab, line 25: For consistency, please describe ALL the parameters including the new fwhm and n_repeats in the docstring.

stellarphot/photometry/tests/fake_image.py
```diff
@@ -24,10 +25,46 @@ class FakeImage:
...
+            self._sources["x_stddev"] = gaussian_fwhm_to_sigma * fwhm
+            self._sources["y_stddev"] = gaussian_fwhm_to_sigma * fwhm
+        if n_repeats > 1:
+            amplitude_scales = np.logspace(
+                start=np.log10(0.1), stop=np.log10(60.0), num=n_repeats**2
+            )
+            # Reshape to n_repeats x n_repeats
+            amplitude_scales = amplitude_scales.reshape((n_repeats, n_repeats))
+            extra_sources = []
+            for i in range(n_repeats):
```
JuanCab, line 44: This nested pair of loops seems to suggest you apprend `n_repeats` SQUARED sources to `extra_sources`.  That doesn't seem to make sense given the parameter  name.  Can you clarify what this nested loop is supposed to do.

JuanCab (commented):
See the comments, looks mostly good, but could use some additional commenting.

JuanCab:
> * The "gap" and "width" of the annulus are currently still in pixels even though the "radius" is a multiplier of the FWHM iv `variable_aperture` is True. So, for example, if `variable_aperture` is `True`, and the fwhm is 5 and the aperture radius is 1.5, the radius in pixels is 7.5. **The key question here is whether gap and annulus should be pixels or something else.**

What is the "something else" you are curious about, as in should they be in multiples of FWHM?

If I am recalling correctly, the outer annulus is used for background estimation, so in that regard, we just need to get far enough out that very little of the star's light is out there contributing to the background estimate.  Is 1.5*FWHM far enough out to avoid enough "starlight" contributing to the counts in the outer annulus?

If I am misinterpreting the concern, please clarify.

stellarphot/photometry/tests/fake_image.py
mwcraig: It is supposed to produced `n__repeats**2` copies. I'll come up with a better name when I document the arguments.

stellarphot/photometry/photometry.py
mwcraig: I think it was just a reminder to change the aperture radius to be in pixels, i.e. to use `radius_pixel` instead of `radius`.

mwcraig: Removed it.

stellarphot/photometry/photometry.py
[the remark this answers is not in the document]
mwcraig: I find its summaries _really_ nice.

stellarphot/photometry/tests/fake_image.py
mwcraig: Fixed,

stellarphot/photometry/tests/fake_image.py
mwcraig: Renamed the variable to `n_repeats_per_side`

stellarphot/gui_tools/seeing_profile_functions.py
mwcraig: There are two places below that need the FWHM. It turns out that `rad_prof` has a `FWHM` attribute, so now I just use that in the two places that the FWHM is needed.

mwcraig:
@JuanCab -- I think I've addressed all of the review comments. Thanks for the quick review!

> What is the "something else" you are curious about, as in should they be in multiples of FWHM?

I didn't have anything super concrete in mind. I could see leaving `gap` and `width` in pixels. I could also see setting gap to some somewhat bigger multiple of the FWHM (so if radius is 1.5x FWHM, then gap is 2x FWHM, say) and setting the width so that the number of annulus pixels is much larger than the number of aperture pixels.

Leaving it as is is fine for now I think.

> Is 1.5*FWHM far enough out to avoid enough "starlight" contributing to the counts in the outer annulus?

Not really -- I'd expect it is better to use, e.g. 2x FWHM, so that the gap is 0.5x FWHM.

JuanCab (approved):
Looks good.

PR #509, opened 2025-04-26
Add ATLAS All Sky Survey catalog (refcat2) as a catalog option

mwcraig:
This pull request adds another catalog to stellarphot. It is the catalog used by the ATLAS project and has much better photometric precision that APASS. Unfortunately, magnitudes are  only available in the Pan-STARRS griz passbands, which are close to, but not quite the same as, the Sloan griz passbands. This PR treats them as being the same.

I'm about to open an issue about how to properly handle the passbands.


stellarphot/core.py
```diff
@@ -1326,9 +1334,128 @@ def prepare_cat(cat):
...
+    field_center : `astropy.coordinates.SkyCoord`, `astropy.wcs.WCS`, or FITS header
+        Either a `~astropy.coordinates.SkyCoord` object, a `~astropy.wcs.WCS` object
+        or a FITS header with WCS information. The input coordinate should be the
+        center of the frame; if a header or WCS is the input then the center of the
+        frame will be determined from the WCS.
+
+    radius : `astropy.units.Quantity`, optional
+        Radius around which to search.
+
+    clip_by_frame : bool, optional
```
Tanner728, line 1357: Is this needed due to the catalog search being a circular region and the frame being square?

Tanner728 (commented):
I don't see anything that will break it, but that doesn't mean too much.

stellarphot/core.py
mwcraig: Yeah; the search is always a cone, but it can be trimmed down to a rectangle if desired.

mwcraig:
I am once again thoroughly confused by codecov's claim of uncovered lines. Most of them are in `refcat2`, which I added a test for, so I'm pretty cofused about how it ends up thinking stuff isn't covered. I could maybe understand it not thinking that the nested function def is covered but 🤷‍♂️

.pre-commit-config.yaml
JuanCab, line 1: I'm assuming this is tweaking the pre-commit testing to support the latest version of ruff and to assume we need python 3.13 support

stellarphot/core.py
```diff
@@ -1030,6 +1032,11 @@ def from_vizier(
             If True, the catalog data does not contain error information, so a
             column of NaNs will be added for the error values.
 
+        tidy_catalog : bool, optional
+            If True, the catalog will be tidied into a long format with one
```
JuanCab, line 1036: I must admit I am not sure what you mean by 'tidied'.... can you possibly document this a bit more clearly, explaining that it reformats the table keeping magnitudes and colors more cleanly?

stellarphot/core.py
```diff
@@ -1200,7 +1207,7 @@ def apass_dr9(field_center, radius=1 * u.degree, clip_by_frame=False, padding=10
 
     clip_by_frame : bool, optional
         If ``True``, only return items that are within the field of view
-        of the frame. Default is ``True``.
+        of the frame.
```
JuanCab, line 1210: You removed the statement that the default was ``True``.  Do you want to instead say it is ``False``.

stellarphot/core.py
```diff
@@ -1292,7 +1299,7 @@ def vsx_vizier(field_center, radius=1 * u.degree, clip_by_frame=False, padding=1
...
```
JuanCab, line 1302: Again, do you want to indicate the default is ``False``.

stellarphot/core.py
```diff
@@ -1326,9 +1333,128 @@ def prepare_cat(cat):
...
+        """
+        This function does a few things:
+
+        1. Filter out galaxies from the catalog.
+        2. Only keep stars that are in the Gaia DR2 catalog.
+        3. Add the Gaia DR2 ID number to the catalog as the ID column.
+        """
+        # 1.
+        # The refcat2 paper says that "Virtually all galaxies can be rejected by
+        # selecting objects forwhich Gaia provides a nonzero proper-motion uncertainty,"
```
JuanCab, line 1399:
```suggestion
        # selecting objects for which Gaia provides a nonzero proper-motion uncertainty,"
```

JuanCab (approved):
Looks good, I made some comments, but as far as I can tell, looks good (it took a while for me to get the logic of your filtering of the ATLAS entries using GAIA, but it seems sound.

.pre-commit-config.yaml
mwcraig: Yeah...

stellarphot/core.py
mwcraig: Done

PR #515, opened 2025-05-12
Add ability for catalogs to perform transforms on themselves

mwcraig:
This also replaces the old transform notebook with something that is simpler.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -1186,7 +1186,7 @@ class CatalogData(BaseEnhancedTable):
 
         reduced_input = self[passband_mask]
 
-        # Grab a copy of the metadata to make sure it is preserved. to_pands will
+        # Grab a copy of the metadata to make sure it is preserved. to_pandas will
         # strip the metadata from the table.
         metadata = reduced_input.meta.copy()
 
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -1134,7 +1134,7 @@ class CatalogData(BaseEnhancedTable):
 
         return cat
 
-    def passband_columns(self, passbands=None):
+    def passband_columns(self, passbands=None, transformer=None):
         """
         Return an `astropy.table.Table` with passbands as column names instead
         of the default format, which has a single column for passbands.
@@ -1144,6 +1144,12 @@ class CatalogData(BaseEnhancedTable):
         passbands : list, optional
             List of passbands to include in the output. If not provided, all
             passbands in the catalog will be included.
+
+        transformer : callable, optional
+            Function to transform the data in the passband columns. The function
+            should take a single argument, which is the data in the passband
+            column, and return the transformed data. If not provided, no
+            transformation will be applied.
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


Tanner728 (approved):

stellarphot/core.py
```diff
@@ -1163,7 +1170,12 @@ def passband_columns(self, passbands=None):
             passbands = catalog_passbands
         input_passbands = set(passbands)
         missing_passbands = input_passbands - catalog_passbands
-        if missing_passbands:
+        if missing_passbands == input_passbands:
+            # The user only request transformed passbands, give them all of them
```
JuanCab, line 1174: If I interpret the code correctly, if `missing_passbands == input_passbands` then `catalog_passbands` and `input_passbands` have no passbands in common, so what is the logic in replacing `input_passbands` with `catalog_passbands`?  Presumably the user has requested passbands NOT in the catalog in this case, are we not honoring that?

The comment is grammatically unclear which is part of the problem.

stellarphot/core.py
```diff
@@ -1173,6 +1185,10 @@ def passband_columns(self, passbands=None):
 
         reduced_input = self[passband_mask]
 
+        # Grab a copy of the metadata to make sure it is preserved. to_pands will
```
JuanCab, line 1188:
```suggestion
        # Grab a copy of the metadata to make sure it is preserved. to_pandas will
```

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -214,3 +214,85 @@ def load(cls) -> "PanStarrs1ToJohnsonCousins":
...
+    table["mag_IC"] = filter_transform(
+        table,
+        output_filter="I",
+        g="mag_SG",
+        r="mag_SR",
+        i="mag_SI",
+        transform="jester",
+    )
+
+    # Yes, this is dumb. You fix it if you want it less dumb.
```
JuanCab, line 249: Why is this necessary? It might be better to explain the problem here in the comments if you want to give a future 'fixer' of the bug motivation.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -603,39 +605,42 @@ def fake_it(one_image):
...
-        our_coords = SkyCoord(one_image["ra"], one_image["dec"], unit="degree")
-        if cat is None or cat_coords is None:
+    one_coord = SkyCoord(
+        observed_mags_grouped["ra"][0], observed_mags_grouped["dec"][0], unit="degree"
+    )
+    if cat is None or cat_coords is None:
+        if cat_name == "apass_dr9":
             cat = apass_dr9(
-                our_coords[0], radius=1 * u.degree, clip_by_frame=False, padding=0
+                one_coord, radius=1 * u.degree, clip_by_frame=False, padding=0
```
JuanCab, line 614: Not a complaint, but isn't 1 degree a pretty large search radius for APASS DR9 or is it really so few stars?

stellarphot/utils/magnitude_transforms.py
```diff
@@ -644,7 +649,10 @@ def fake_it(one_image):
         color = cat["color"][cat_idx]
 
         # Impose some constraints on what is included in the fit
-        good_cat = ~(color.mask | cat_mag.mask) & (d2d.arcsecond < 1)
+        good_cat = ~(
+            getattr(color, "mask", False)
```
JuanCab, line 653: Just curious why the switch to `getattr`, was this to allow default values?

JuanCab (approved):
Looks mostly good, most of my comments were about comments....

stellarphot/utils/magnitude_transforms.py
mwcraig: This was because the color, etc sometimes has a mask and some times doesn't, which is sort of a hassle.

stellarphot/utils/magnitude_transforms.py
mwcraig: our fov is about 0.5 degree, and the 1 degree search executes pretty quickly.

PR #522, opened 2025-06-10
Change name of the optional setting "method" to "partial_pixel_method"

### LINKED ISSUE #498: Rename `PhotometryOptionalSettings.method` to something more explicit
I noticed the name when I changed `fwhm_by_fit` to `fwhm_method`. The current `method` is really the method for handling partial pixels, and a more explicit name would be better.

mwcraig:
This pull request fixes #498 by renaming the generic sounding "method" to "partial_pixel_method".

I also added a couple of tests to make sure that the "alpha" series of photometry files are still readable.

### DIFF
--- a/stellarphot/tests/test_core.py
+++ b/stellarphot/tests/test_core.py
@@ -1,4 +1,5 @@
 import warnings
+from pathlib import Path
 
 import numpy as np
 import pytest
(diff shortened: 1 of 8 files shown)


stellarphot/tests/test_core.py
```diff
@@ -1369,3 +1370,14 @@ def test_to_lightcurve_multiple_passbands(simple_photometry_data):
         match=r"Passband SI not found for this star",
     ):
         two_filters.lightcurve_for(1, passband="SI")
+
+
+def test_reading_2_0_0_alpha_photometry_file():
+    """
+    Just make sure that old photometry files are still readable.
```
JuanCab, line 1377: Is this the old or the new format when you refer to 2.0.0alpha, I am guessing old, but if it is unchanged from the last version 1 settings, I think it makes more sense to call this `test_reading_v1_photometry_file`.

mwcraig: "old" here refers to `2.0.0alpoha`, which I've been running on jupyter3 since last August. That has been an adventure....I wanted to make sure that updating to a 2.0 release won't stop research students in their tracks.

PR #525, opened 2025-06-12
Add ability to set limiting magnitude in catalog searches

### LINKED ISSUE #521: Add option for limiting magnitude for catalog retrieval
Tagging as a new feature, but borders on bug if you want a wide shallow field.

mwcraig:
This fixes #521 by adding the ability to specify a limiting magnitude when getting a catalog.


JuanCab (dismissed):
Looks generally good, but see comments.

stellarphot/core.py
```diff
@@ -1140,7 +1152,17 @@ def from_vizier(
             center = SkyCoord(*wcs.wcs.crval, unit="deg")
 
         # Get catalog via cone search -- all columns, no limit on rows
-        vizier = Vizier(columns=["all"], row_limit=-1)
+        column_filter = (
+            {f"{magnitude_limit_passband}": f"<={magnitude_limit}"}
+            if magnitude_limit is not None
```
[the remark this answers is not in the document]
mwcraig: No, because I've checked for that combination earlier in the code

stellarphot/core.py
```diff
@@ -1396,6 +1462,14 @@ def vsx_vizier(field_center, radius=1 * u.degree, clip_by_frame=False, padding=1
...
         is 100.
 
+    magnitude_limit : float, optional
+        If provided, only return items with a brightest magnitudes less than or equal
+        to this value.
+
+    magnitude_limit_passband : str, optional
+        There is no straightforward way to limit the VSX catalog by passband. The
+        magnitude limit will be applied to the variable star's magnitude at maximum.
+
```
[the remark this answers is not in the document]
mwcraig: It is for consistency with others -- I'll update the docstring in a moment. I guess it is included for two reasons, really -- one is for consistency but also to be able to tell the user what to do if they try to force in a passband (rather than just not allowing the keyword argument).

mwcraig:
Tests pass locally, so if you are fine with merging this, I am too....

stellarphot/core.py
```diff
@@ -1347,7 +1347,7 @@ def apass_dr9(
         If provided, only return items with magnitudes less than or equal
         to this value.
 
-    magnitude_limit_passband : str, optional
+    magnitude_limit_passband : str, optional, default is "V"
```
JuanCab, line 1350: You changed the default to `None` from what I can tell...

stellarphot/core.py
```diff
@@ -1476,9 +1489,10 @@ def vsx_vizier(
         If provided, only return items with a brightest magnitudes less than or equal
         to this value.
 
-    magnitude_limit_passband : str, optional
+    magnitude_limit_passband : str, optional, default is "max"
```
JuanCab, line 1492: Default has NOT been set to "max" in the function declaration.

stellarphot/core.py
```diff
@@ -1529,7 +1543,7 @@ def refcat2(
     clip_by_frame=False,
     padding=100,
     magnitude_limit=None,
-    magnitude_limit_passband="SR",
+    magnitude_limit_passband=None,
```
JuanCab, line 1546: it appears to be defaulting to `None` instead of `SR` as the documentation below implies (and as it formerly did).

JuanCab (changes requested):
Default bandpasses appear to have been accidentally set to `None` (they appear to have been defined and the change was reverted).  Please review them and confirm.

stellarphot/core.py
mwcraig: Later in the code the default is set to SR if `magnitude_limit` is not None...

mwcraig:
> Default bandpasses appear to have been accidentally set to None (they appear to have been defined and the change was reverted). Please review them and confirm.

To clarify, I was aiming to have default values but also be able to check, in `from_vizier` for a `magnitude_limit` without a `magnitude_limit_passband`. 

Setting a default  as a string in the individual catalogs meant that the when `from_vizier` there was a `magnitude_limit_passband` set for even if the user hadn't set one.

I suppose I could, if you would prefer, add a check in the catalogs for `magnitude_limit is None` and unset the `magnitude_limit_passband` if it has the default value.

LMK if you want me to switch it to that...

PR #532, opened 2025-06-20
Add transform from USNO to SDSS system

### LINKED ISSUE #516: Fix Jester et al transforms
Buried somewhere in the code is an implementation of the transform equations given by jester et al for transforming to/from SDSSS passbands to Johnson/Cousins.

Those transforms are to the *ugriz* filters that are the main SDSS scope, not the *u'g'r'i'z'* filters everyone else has. That should be fixed. See https://github.com/feder-observatory/stellarphot/issues/510#issuecomment-2832541152.

mwcraig:
This fixes #516 by adding the option to include the transform from USNO u'g'r'i'z' to SDSS ugriz.


stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -251,6 +382,10 @@ def transform_apass_bands(table):
     table["mag_I"] = table["mag_IC"]
     # The dumbness has ended for now
 
+    # Remove the temporary columns if they were created
+    if apply_sdssdr7_transform:
+        del table["mag_SG_tmp", "mag_SR_tmp", "mag_SI_tmp"]
+
```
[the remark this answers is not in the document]
mwcraig: You are still wrong about this, @Copilot...

stellarphot/utils/data/USNO_SDSS_to_SDSS_DR7.json
```diff
@@ -0,0 +1,40 @@
...
+            "z"
+        ]
+    },
+
+
+    "transform_information": {
+        "name": "USNO_SDSS_to_SDSS_DR7",
+        "from_passbands": ["u'","g'","r'","i'","z'"],
+        "to_passbands": ["u","g","r","i","z"],
+        "transformation_matrix":
```
Tanner728, line 30: I am not seeing where some of the numbers listed in the transformation matrix are coming from. I probably misunderstand the transformation matrix but I assume it is just these equations from the referenced 2005 website.
image

stellarphot/utils/data/PS1_to_JC.json
```diff
@@ -21,7 +21,7 @@
             "Ic"
         ]
     },
-    "transform_coefficients": {
+    "transform_information": {
```
Tanner728, line 24: I think this better conveys what is actually contained in the dictionary.

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -82,9 +84,42 @@ def polynomial(self) -> np.polynomial.Polynomial:
         return np.polynomial.Polynomial(self.polynomial_coefficients)
 
 
+class MagnitudeTransformMatrix(BaseModel):
+    """
```
Tanner728, line 88: Do we want the same definition for both "MagnitudeTransformPolynomial" and this class?

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -229,20 +322,58 @@ def transform_apass_bands(table):
...
         g="mag_SG",
         r="mag_SR",
         i="mag_SI",
+    )
+
+    # The Jester transformation really for transformation from the ugriz on the
+    # 2.5m SDSS telescope to the Johnson-Cousins system.
+    #
+    # APASS is using u'g'r'i'z' (well, only g'r'i' in APASS) and so I *think*
+    # that the right thing to do is to transform the APASS g'r'i' to the SDSS DR7
```
Tanner728, line 336: Will this add a significant amount of uncertainty to the transformed magnitudes if we are doing a double transform? If I am understanding this correctly.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: I wish -- in the long run we can do this because we can write the polynomial thing as a matrix if we decide we need to. For now it was easier to just make two classes to represent the two broad ways I've seen of transforming.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Not sure what the residual is in each of the fits and no one seems to express the coefficients as value ± error so we can't really propagate any transform errors.

Right now I *think* that using Jester only to change APASS r'i' to Rc and Ic is wrong. Jester's paper clearly states that he redid an earlier set of transforms to be *on the SDSS system*, not the USNO system, (aka. the "primed" system).

stellarphot/core.py
```diff
@@ -1217,7 +1217,9 @@ def from_vizier(
 
         return cat
 
-    def passband_columns(self, passbands=None, transformer=None):
+    def passband_columns(
+        self, passbands=None, transformer=None, transformer_kwargs=None
+    ):
         """
         Return an `astropy.table.Table` with passbands as column names instead
```
JuanCab, line 1224: You might want to document `transformer_kwargs` as a parameter in the docstring (I can guess from context, but just for completeness).

stellarphot/core.py
```diff
@@ -1295,9 +1297,10 @@ def passband_columns(self, passbands=None, transformer=None):
         # Add the metadata back to the table
         return_table.meta.update(metadata)
 
+        transformer_kwargs = transformer_kwargs or {}
```
JuanCab, line 1300: Clever, had to stare at this a bit to figure it out.  I might use it in my class when introducing booleans.

stellarphot/utils/data/USNO_SDSS_to_SDSS_DR7.json
JuanCab: I'll admit I am lost here as well as to how you get this transformation matrix from the referenced documentation.  (time passes) I looked over your gist and have a better idea of what you are doing.  You might want to distill that gist down into some comments or documentation to include with the code.

stellarphot/utils/tests/test_magnitude_system_transforms.py
```diff
@@ -216,3 +233,21 @@ def test_refcats_adds_BVRI(self):
...
+        # Check that the zero point for ug is correct
+        inp_mag = np.asarray([1.39, 0.0, 0.0, 0.0, 0.0, 1.0])
+        # This should produce this output:
+        # u: same as u' so 1.39
+        # g: product of -b_g and -zp_gr, so -0.0318
+        # r: product of -b_r and -zp_ri, so -0.00735
+        # i: product of -b_i and -zp_ri, so -0.00861
+        # z: product of -b_z and -zp_iz, so 0.0027
+
+        # This effectively checks the whole matrix
```
JuanCab, line 249: Are there any test values we can use other than ones based on knowing the matrix based on observations (e.g. known ugriz to u'g'r'i'z' mapping)?

JuanCab (dismissed):
Looks good for the most part, I was curious if another test was possible BUT don't think that needs to hold things up.  Might be worth it to see if we can reproduce known transforms of ugriz to u'g'r'i'z' if possible.

PR #533, opened 2025-06-22
Fix representation of table attributes

mwcraig:
It turns out there was a bug in #519: stellarphot objects stored as `dict`s in the metadata were not properly turned back into onjects on reading if they were table attributes instead of being stored directly in the table metadata dictionary.

This PR adds a test for that and fixes the issue.


mwcraig:
YOLO!

PR #535, opened 2025-06-22
Add properties for several of the more useful table columns

### LINKED ISSUE #452: Etiher the documentation or `PhotometryData` needs updating...
In [the documentation](https://github.com/feder-observatory/stellarphot/blob/aca116777bf6fb351590906279f252fc81bffcc0/docs/stellarphot/index.rst?plain=1#L7) we say this about photometry data in stellarphot:

```
Photometry objects always have these attributes, also available dictionary-style:

+ ``mag`` -- the calibrated magnitude; may be missing (e.g. your data prior to calibration)
+ ``mag_err`` -- error in ``mag``
+ ``inst_mag`` -- the instrumental magnitude; may be missing (e.g. catalog data)
+ ``inst_mag_err`` -- error int the ``inst_mag``
+ ``band`` -- the filter of the magnitude; required
+ ``BJD`` -- Barycentric Julian Date of the midpoint of the observation; may be missing/masked
+ ``RA2000`` -- right ascension in degrees, in the ICRS frame at epoch 2000
+ ``DEC2000`` -- declination in degrees in the ICRS frame at epoch 2000
```

We should either update `PhotometryData` to do this (which seems like a fine and simple route) or update the documentation. I'd lean towards having a coordinate object provided instead of separate RA and Dec, though. The coordinate object is much more useful.

@JuanCab -- accumulating issues to talk through 😬

mwcraig:
Fixes #452

@JuanCab -- the thing I have the question about is renaming the column "mag_error"....

### DIFF
--- a/docs/stellarphot/user_guide/using_photometry_data.rst
+++ b/docs/stellarphot/user_guide/using_photometry_data.rst
@@ -1,15 +1,28 @@
 Working with  Photometry Data
 #############################
 
-Photometry objects always have these attributes, also available dictionary-style:
 
-+ ``mag`` -- the calibrated magnitude; may be missing (e.g. your data prior to calibration)
-+ ``mag_err`` -- error in ``mag``
-+ ``inst_mag`` -- the instrumental magnitude; may be missing (e.g. catalog data)
-+ ``inst_mag_err`` -- error int the ``inst_mag``
-+ ``band`` -- the filter of the magnitude; required
-+ ``BJD`` -- Barycentric Julian Date of the midpoint of the observation; may be missing/masked
-+ ``RA2000`` -- right ascension in degrees, in the ICRS frame at epoch 2000
-+ ``DEC2000`` -- declination in degrees in the ICRS frame at epoch 2000
+Photometry data attributes
+--------------------------
 
-There may be additional fields.
+`~stellarphot.PhotometryData` objects always have these attributes, also available dictionary-style:
+
++ ``coord`` -- an astropy `~astropy.coordinates.SkyCoord` object with the coordinates for each star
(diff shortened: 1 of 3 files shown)


docs/stellarphot/user_guide/using_photometry_data.rst
```diff
@@ -1,15 +1,28 @@
...
-There may be additional fields.
+
+Photometry data attributes
+--------------------------
+
+`~stellarphot.PhotometryData` objects always have these attributes, also available dictionary-style:
+
++ ``coord`` -- an astropy `~astropy.coordinates.SkyCoord` object with the coordinates for each star
++ ``mag_inst`` -- the instrumental magnitude
++ ``mag_inst_error`` -- error in the instrument magnitude
```
JuanCab, line 12: shouldn't it be "instrumental magnitude" instead of "instrument magnitude"?  Also, for each star, right?

docs/stellarphot/user_guide/using_photometry_data.rst
```diff
@@ -1,15 +1,28 @@
...
-
-There may be additional fields.
+
+Photometry data attributes
+--------------------------
+
+`~stellarphot.PhotometryData` objects always have these attributes, also available dictionary-style:
+
++ ``coord`` -- an astropy `~astropy.coordinates.SkyCoord` object with the coordinates for each star
++ ``mag_inst`` -- the instrumental magnitude
```
JuanCab, line 11: I may have forgotten the details of Photometry Data objects, but isn't this instrumental magnitudes for each star?

JuanCab (approved):
So far looks good...

PR #567, opened 2026-06-16
Decouple core data structures from the ipywidgets/jupyter stack

mwcraig:
## What & why

Importing `stellarphot` eagerly pulled in the entire GUI stack
(`ipywidgets`, `ipyautoui`, `bqplot`, `ipyfilechooser`, `traitlets`), so the
core data structures (`PhotometryData`, `CatalogData`, `SourceListData`,
`Camera`, `Observatory`, `PassbandMap`, `PhotometrySettings`, …) could not be
used in a lightweight/headless context and every import paid the widget cost.
This also removes a `core` ↔ `io` circular import previously worked around by an
in-method lazy import.

The change is split into three independently-reviewable commits so each can be
evaluated (and the second dropped) on its own.

## Commits

1. **Decouple ipywidgets from `import stellarphot`.** `settings/__init__.py` no
   longer does `from .views import *`; `ui_generator` is re-exported lazily via
   a PEP 562 module `__getattr__`, so `from stellarphot.settings import
   ui_generator` keeps working without loading the GUI stack at import time.
   Internal GUI consumers import `ui_generator` directly from
   `stellarphot.settings.views`.

2. **Remove the `core` ↔ `io` lazy-import workaround.** Only `io/tess.py` imports
   `core`; `io/aavso.py` and `io/aij.py` are clean. `io/__init__.py` now exposes
   `tess` lazily (PEP 562 `__getattr__`, skipping private/dunder probes), so
   `core.py` imports `write_aavso_extended` from `io.aavso` at the top of the
   file and the in-method lazy import is gone. `io/tess.py` imports
   `SourceListData` from `stellarphot.core` directly.

3. **Update architecture diagrams** to reflect the lazy widget layer and the
   one-way `core` → `io.aavso` (eager) / `io.tess` → `core` (lazy) relationship.

## Verification

- `import stellarphot` and importing any core data structure load **no**
  `ipywidgets`/`ipyautoui`/`bqplot`/`ipyfilechooser`.
- `from stellarphot.settings import ui_generator` still works (loads widgets on
  demand); `from stellarphot.io import write_aavso_extended, tess_photometry_setup`
  still resolve; `import stellarphot.io` works in both import orders.
- Full test suite: **671 passed, 49 skipped** (skips are batman/remote-data/
  detection-combo, unrelated).

## Note for reviewers

Commit 1 stands on its own. Commit 2 is the more speculative one: it makes
`io.tess` lazy, a subtle behavior change (e.g. `dir(stellarphot.io)` will not list
tess names until first access). It can be dropped without affecting commit 1.

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -19,6 +19,9 @@ from astropy.wcs import WCS
 from astroquery.vizier import Vizier
 from astroquery.xmatch import XMatch
 
+# Imported under a private alias so the module-level name does not collide with
+# the PhotometryData.write_aavso_extended method that delegates to it, and so it
+# is not exposed as a public stellarphot.core attribute.
 from .io.aavso import write_aavso_extended as _write_aavso_extended
 from .settings import Camera, Observatory, PassbandMap
 from .table_representations import (
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -19,6 +19,7 @@ from astropy.wcs import WCS
 from astroquery.vizier import Vizier
 from astroquery.xmatch import XMatch
 
+from .io.aavso import write_aavso_extended as _write_aavso_extended
 from .settings import Camera, Observatory, PassbandMap
 from .table_representations import (
     _generate_old_table_representers,
@@ -788,11 +789,7 @@ class PhotometryData(BaseEnhancedTable):
         See `stellarphot.io.write_aavso_extended` for the full signature and
         behavior. All keyword arguments are forwarded to that function.
         """
-        # Lazy import to avoid a circular dependency: stellarphot.io imports
-        # from stellarphot.settings, which is loaded before core.
-        from .io.aavso import write_aavso_extended
-
-        return write_aavso_extended(self, path, **kwargs)
+        return _write_aavso_extended(self, path, **kwargs)
 
 
 class CatalogData(BaseEnhancedTable):
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -786,7 +786,7 @@ class PhotometryData(BaseEnhancedTable):
     def write_aavso_extended(self, path, **kwargs):
         """Write this photometry table in the AAVSO Extended File Format.
 
-        See `stellarphot.io.write_aavso_extended` for the full signature and
+        See `stellarphot.io.aavso.write_aavso_extended` for the full signature and
         behavior. All keyword arguments are forwarded to that function.
         """
         return _write_aavso_extended(self, path, **kwargs)
--- a/stellarphot/io/__init__.py
+++ b/stellarphot/io/__init__.py
@@ -1,5 +1,22 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import importlib
+
 from .aavso import *
 from .aij import *
-from .tess import *
+
+
+# tess imports stellarphot.core; expose its public names lazily so that importing
+# io (e.g. from core.py) does not create a core <-> io import cycle. We import the
+# submodule via importlib rather than ``from . import tess`` to avoid re-entering
+# this __getattr__ through the from-import machinery.
+def __getattr__(name):
+    # Short-circuit private/dunder probes (e.g. doctest's ``hasattr(mod,
+    # "__test__")``) so they do not import tess and mutate this package's
+    # namespace mid-iteration in tools that walk ``__dict__``.
+    if name.startswith("_"):
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+    tess = importlib.import_module(".tess", __name__)
+    if name in tess.__all__:
+        return getattr(tess, name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/stellarphot/io/__init__.py
+++ b/stellarphot/io/__init__.py
@@ -1,22 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import importlib
-
-from .aavso import *
-from .aij import *
-
-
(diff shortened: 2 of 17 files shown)


stellarphot/io/__init__.py
```diff
@@ -1,5 +1,22 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import importlib
+
 from .aavso import *
 from .aij import *
-from .tess import *
+
+
+# tess imports stellarphot.core; expose its public names lazily so that importing
```
mwcraig: Could we change the way the imports work so that consumers have to import from stellarphot.io.aavso, .aij, .tess explicitly instead of populating the stellarphot.io namespace?

stellarphot/settings/__init__.py
```diff
@@ -14,3 +13,14 @@
 
 # Clean up the namespace
 del TEST_PHOTOMETRY_SETTINGS
+
+
+def __getattr__(name):
+    # Lazily expose the widget builder so importing the data models does not pull
+    # in ipywidgets/ipyautoui at package import time (see views.py). This keeps
+    # ``from stellarphot.settings import ui_generator`` working for notebooks.
```
mwcraig: Clarify: is this a compatibility shim or something that needs to stay around long term?

stellarphot/core.py
```diff
@@ -19,6 +19,7 @@
 from astroquery.vizier import Vizier
 from astroquery.xmatch import XMatch
 
+from .io.aavso import write_aavso_extended as _write_aavso_extended
```
mwcraig, line 25: Why add the underscore at the front of the name?

stellarphot/io/__init__.py
mwcraig: Yes — and I think that's the cleaner design. There are only four internal sites that pull names out of the `stellarphot.io` namespace:

- `settings/custom_widgets.py` → `tess_photometry_setup` (tess)
- `gui_tools/seeing_profile_functions.py` → `TessSubmission` (tess)
- `settings/tests/test_tess_photometry_setup.py` → `TOI` (tess)
- `io/tests/test_aij_io.py` → `ApertureFileAIJ` (aij)

If those move to explicit submodule imports (`from stellarphot.io.tess import ...`, `from stellarphot.io.aij import ...`), then `io/__init__.py` can drop the `from .aavso import *` / `from .aij import *` re-exports **and** the `__getattr__` entirely — it becomes essentially empty. That:

- stops populating the `stellarphot.io` namespace (this ask),
- removes the lazy shim and the `importlib` import Copilot flagged below,
- still keeps the cycle broken: `core.py`'s `from .io.aavso import write_aavso_extended` runs an (empty) `io/__init__` and imports the clean `aavso` submodule, never touching `tess`.

The only cost is a breaking change for any external/notebook code doing `from stellarphot.io import X`, which I'd call out in the changelog. Want me to go this route instead of the lazy `__getattr__`?

stellarphot/settings/__init__.py
mwcraig: It's a compatibility shim. Its only job is to keep `from stellarphot.settings import ui_generator` working for existing notebooks/users while making `import stellarphot.settings` GUI-free. All in-tree consumers already import `ui_generator` from `stellarphot.settings.views` directly (this PR updates them), so nothing internal depends on it.

Long-term I see two options:

1. Keep it indefinitely as a thin back-compat alias.
2. Deprecate it — have the shim emit a `DeprecationWarning` pointing at `stellarphot.settings.views.ui_generator`, then remove it in a future release.

I'd lean toward (2) if we want callers to migrate off it, otherwise (1) is harmless. Happy to do whichever you prefer.

stellarphot/core.py
mwcraig: Two reasons, both cosmetic:

1. It avoids a module-level global named `write_aavso_extended` sitting right next to the `PhotometryData.write_aavso_extended` method. The leading underscore makes clear the module name is the imported `io` helper the method delegates to, not the method itself.
2. It signals "internal" for plain attribute access / tab-completion on `stellarphot.core`. (`core.py` already has an `__all__`, so the name wouldn't be re-exported via `*` either way — this is just about direct access.)

It's not required for correctness. If you'd rather not rename, `from .io.aavso import write_aavso_extended` works equally well and the method body just calls `write_aavso_extended(self, ...)`. Let me know your preference.

stellarphot/io/__init__.py
```diff
@@ -1,5 +1,22 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import importlib
+
```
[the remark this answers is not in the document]
mwcraig: Good catch. If we keep this structure, `import importlib as _importlib` (or `from importlib import import_module as _import_module`) avoids leaking the name.

That said, this becomes moot if we adopt @mwcraig's suggestion above to require explicit submodule imports (`from stellarphot.io.tess import ...`) — that lets `io/__init__.py` drop the `__getattr__` and the `importlib` import altogether.

stellarphot/settings/__init__.py
mwcraig: Just delete it. As far as I know the only code using this is the code in this repo. If we don't need it then no one does....

stellarphot/io/__init__.py
mwcraig: Yes, go this route.

stellarphot/core.py
mwcraig: Keep as is but add a comment explaining it.

stellarphot/io/__init__.py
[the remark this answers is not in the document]
mwcraig: This will end up getting dropped on the next commit.

PR #574, opened 2026-06-18
Fix multi_image_photometry crash on a single missing source (#474)

### LINKED ISSUE #474: Add tests for uncovered block
This [block of code](https://github.com/feder-observatory/stellarphot/blob/main/stellarphot/photometry/photometry.py#L790C1-L815C1) in `photometry.py` is not covered by any tests. There is at least one bug in here on line 794: whether or not `missing_sources` has length one or larger we want `set(missing_courses)` not `set([missing_sources])`. There may (or may not) be others, but increasing the coverage of this block would be good.

mwcraig:
Closes #474.

## Problem
In `multi_image_photometry`, when exactly one source was missing from at least one image and `reject_unmatched=True`, the dedup code took a buggy branch:

```python
if len(missing_sources) > 1:
    uniques = set(missing_sources)
else:
    uniques = set([missing_sources])   # wraps the list -> unhashable
```

`set([missing_sources])` tries to build a set containing the list itself, raising `TypeError: unhashable type: 'list'`.

## Fix
The `len > 1` / `len == 1` split is unnecessary — `set(missing_sources)` is correct in all cases. Replaced the whole `if/else` with a single `uniques = set(missing_sources)`.

## Test
Added `test_reject_unmatched_single_missing_source`, which drives `multi_image_photometry` on a directory and wraps `single_image_photometry` to report exactly one missing source on a single image. It fails with the original `TypeError` and passes with the fix; it also asserts the missing source is rejected from the combined table.

### DIFF
--- a/stellarphot/photometry/tests/test_photometry.py
+++ b/stellarphot/photometry/tests/test_photometry.py
@@ -666,6 +666,77 @@ class TestAperturePhotometry:
             root_logger.removeHandler(foreign_root_handler)
             multilogger.removeHandler(foreign_named_handler)
 
+    def test_reject_unmatched_single_missing_source(
+        self, monkeypatch, tmp_path, photometry_settings_for_test
+    ):
+        # Regression test for #474. When exactly one source is missing from at
+        # least one image, reject_unmatched must not raise. The old code did
+        # ``set([missing_sources])`` in the length-one case, which wraps the
+        # list and raises ``TypeError: unhashable type: 'list'``.
+        from stellarphot.photometry import photometry as photometry_mod
+
+        # Two images are enough: one source is forced missing on the first image
+        # only (see ``fake_single`` below), so the accumulated missing_sources
+        # list has length exactly one -- the case that triggered #474 -- while
+        # the second image lets us check the source is rejected from every image.
+        num_files = 2
+        fake_images = self.list_of_fakes(num_files)
+
+        noise_unit = photometry_settings_for_test.camera.read_noise.unit
+        photometry_settings_for_test.camera.read_noise = (
+            fake_images[0].noise_dev * noise_unit
+        )
+
+        temp_file_names = [
+            tmp_path / f"tempfile_{i:02d}.fit" for i in range(1, num_files + 1)
+        ]
+        for i, image in enumerate(fake_images):
+            image.write(temp_file_names[i])
+        object_name = fake_images[0].header["OBJECT"]
+
+        found_sources = source_detection(
+            fake_images[0],
+            fwhm=fake_images[0].sources["x_stddev"].mean(),
+            threshold=10,
+        )
+        source_list_file = tmp_path / "source_list.ecsv"
+        found_sources.write(source_list_file, format="ascii.ecsv", overwrite=True)
+
+        photometry_settings_for_test.source_location_settings.source_list_file = str(
+            source_list_file
+        )
+
+        # Force exactly one source to be reported missing, and only on the
+        # first image processed, so the accumulated missing_sources list has
+        # length exactly one -- the case that triggered #474. With the default
+        # settings no sources are dropped naturally (see
+        # test_photometry_on_directory), so this mock is the only source missing.
+        missing_id = int(found_sources["star_id"][0])
+        real_single = photometry_mod.single_image_photometry
+        state = {"calls": 0}
+
+        def fake_single(*args, **kwargs):
+            phot, missing = real_single(*args, **kwargs)
+            state["calls"] += 1
+            if state["calls"] == 1:
+                missing = list(missing) + [missing_id]
+            return phot, missing
+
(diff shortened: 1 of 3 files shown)


stellarphot/photometry/tests/test_photometry.py
```diff
@@ -568,6 +568,86 @@ def test_photometry_on_directory(self, coords, photometry_settings_for_test):
...
+    def test_reject_unmatched_single_missing_source(
+        self, monkeypatch, photometry_settings_for_test
+    ):
+        # Regression test for #474. When exactly one source is missing from at
+        # least one image, reject_unmatched must not raise. The old code did
+        # ``set([missing_sources])`` in the length-one case, which wraps the
+        # list and raises ``TypeError: unhashable type: 'list'``.
+        from stellarphot.photometry import photometry as photometry_mod
+
+        num_files = 5
```
mwcraig: Why 5 files? Seems like 2 would be enough....

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -568,6 +568,86 @@ def test_photometry_on_directory(self, coords, photometry_settings_for_test):
...
+        fake_images = self.list_of_fakes(num_files)
+
+        noise_unit = photometry_settings_for_test.camera.read_noise.unit
+        photometry_settings_for_test.camera.read_noise = (
+            fake_images[0].noise_dev * noise_unit
+        )
+
+        with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as temp_dir:
```
mwcraig: Why not use pytest's tmp_path fixture?

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -568,6 +568,86 @@ def test_photometry_on_directory(self, coords, photometry_settings_for_test):
...
+                fwhm=fake_images[0].sources["x_stddev"].mean(),
+                threshold=10,
+            )
+            source_list_file = Path(temp_dir) / "source_list.ecsv"
+            found_sources.write(source_list_file, format="ascii.ecsv", overwrite=True)
+
+            phot_options = (
+                photometry_settings_for_test.photometry_optional_settings.model_copy()
+            )
+            phot_options.include_dig_noise = True
```
mwcraig: Do all of htese need to be set explicitly?

stellarphot/photometry/tests/test_photometry.py
mwcraig: You're right, 2 is enough. The bug only fires when the *accumulated* `missing_sources` list has length exactly 1, and that length is controlled entirely by the `fake_single` mock (it appends `missing_id` on the first call only) — it's independent of how many files there are. The only reason to have more than one image is to make the "rejected from *every* image" assertion meaningful, so 2 is the natural minimum.

I copied `num_files = 5` from the sibling `test_photometry_on_directory`. I verified locally that `num_files = 2` still passes on this branch **and** still fails with `TypeError: unhashable type: 'list'` on the pre-fix code, so dropping to 2 keeps it a valid regression test (and a bit faster). Happy to make that change.

stellarphot/photometry/tests/test_photometry.py
mwcraig: No reason not to — `tmp_path` would be cleaner. I used `tempfile.TemporaryDirectory(ignore_cleanup_errors=True)` only because that's what the sibling directory tests (`test_photometry_on_directory`, etc.) do. Switching to `tmp_path` lets us drop the `with` block, the manual `Path(temp_dir)` plumbing, and the `ignore_cleanup_errors=True` cleanup workaround. The directory still gets passed to `ap_phot(...)` exactly the same way. Happy to switch it.

stellarphot/photometry/tests/test_photometry.py
mwcraig: No — most of these are unnecessary here. Because the missing source is *forced* by the `fake_single` mock, the actual photometry values never matter to the assertion (`missing_id not in phot_data["star_id"]`); only that processing runs and the rejection branch fires. I copied this whole block from `test_photometry_on_directory`.

I verified locally which lines are load-bearing by stripping them and re-running (test still passes on this branch, still fails on the pre-fix code):

- `include_dig_noise`, `reject_too_close`, `reject_background_outliers`, `fwhm_method` → **removable**, no effect on the outcome.
- `use_coordinates = "sky"` → **removable**, it's already the default.
- `source_list_file` → **required** (provides the source list).

Worth noting `reject_unmatched` is never set in the test and isn't needed explicitly — `multi_image_photometry` defaults it to `True`, which is what actually drives the buggy branch. (Side note unrelated to this PR: the `PhotometryOptionalSettings.reject_unmatched` field defaults to `False`, while the function arg defaults to `True`, and `__call__` doesn't pass the setting through — possible inconsistency for another day.)

So this could shrink to just setting `source_list_file`. Happy to trim it.

Issue #153, opened 2023-08-08
Removing root handlers


mwcraig:
In #150 it was necessary to remove several handlers to make logging work. It would be worth thinking through whether we can rework that to avoid removing handlers we may not have added.

JuanCab:
As noted in my comments in #150, the big issue is that anything applied to the root appears to propagate down to the other handlers (which are children of it, if I understand `logger`).  I tried:
- Using `.disable` method on the root handler, this disabled ALL logging.
- Using `.setLevel(logging.CRITICAL+10)` to attempt to get the root handler to not report anything, this sets ALL handlers to never log (since nothing is reported at that level).

This is why I have kept the removal of the root handlers... which is not necessarily friendly, but the `logging` package doesn't seem to offer a way to change the behavior of root handlers WITHOUT it propagating to other loggers' handlers.

JuanCab:
Also, to be fair, the only package causing a problem is `ccdproc` which upon creating a `ImageFileCollection` creates a root logger.  I really believe this is effectively a bug, or at least equally unfriendly behavior of `ccdproc` to force a root logger on all packages importing it.

PR #575, opened 2026-06-18
Factor duplicated photometry logging setup into a helper (#152)

### LINKED ISSUE #152: See if some of the repeated logging code in photometry can be factored out
Some of the nice logging code added in #150 has some repeated stuff (removing handlers, setting up some file-related things) that could perhaps be factored out.

mwcraig:
Closes #152.

The logging code added in #150 duplicated the same handler setup in both `single_image_photometry` and `multi_image_photometry`: build a `FileHandler` (with the `"%(levelname)s - %(message)s"` format) when a logfile is configured, plus a `StreamHandler`/`NullHandler` console handler with the `"%(message)s"` format.

This PR extracts that shared construction into a private `_add_log_handlers(logger, logfile, console_log)` helper that returns the created file handler (or `None`). The genuine differences stay at the call sites:

- `single_image_photometry` only sets up handlers when the logger has none (so a call nested inside `multi_image_photometry` reuses the existing setup);
- `multi_image_photometry` first rewrites the log-file path into the image directory and updates the settings before adding handlers.

Behavior is unchanged. The handler-*removal* cleanup is left to a follow-up for #153 (stacked on this branch).

The existing logging tests (`test_logging_single_image` / `test_logging_multiple_image`) pass unchanged and guard this refactor.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,8 @@ Other Changes and Additions
   back-compat shim that raises an ``AstropyDeprecationWarning``. [#194]
 + The ``change_to_tmp_dir`` test fixture, previously duplicated in four test
   modules, is now defined once in the top-level ``conftest.py``. [#426]
++ The duplicated logging setup in ``single_image_photometry`` and
+  ``multi_image_photometry`` has been factored into a shared helper. [#152]
 
 Bug Fixes
 ^^^^^^^^^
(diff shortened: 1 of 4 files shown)


JuanCab (dismissed):
Assuming this is passing the pre-existing tests (which is should have to reach PR stage), this is an excellent refactoring to avoid duplication.

PR #576, opened 2026-06-18
Stop clearing foreign log handlers in photometry (#153)

### LINKED ISSUE #153: Removing root handlers

mwcraig:
Closes #153.

> **Stacked on #575** (factor logging setup). Until #575 merges, this PR's diff also shows that commit; review the second commit here, or merge #575 first.

## Problem
The logging added in #150 needed to remove handlers to get clean output, and did so bluntly:

- `multi_image_photometry` ran `logging.root.handlers.clear()` — wiping handlers stellarphot never installed (the caller's / other libraries' logging config);
- both functions did `logger.handlers.clear()` / a remove-all loop on their own named loggers.

## Fix
- Handlers created by `_add_log_handlers` are tagged with a `_stellarphot_handler` attribute, and a new `_remove_our_handlers()` removes **only** tagged handlers.
- `_add_log_handlers` sets `logger.propagate = False`, so stellarphot's messages are fully handled by its own handlers and are not duplicated through the root logger. This removes the reason the root handlers were being cleared, so the `logging.root.handlers.clear()` block is deleted entirely.

## Test
`test_multi_image_photometry_preserves_foreign_log_handlers` installs a handler on the root logger and on the `multi_image_photometry` logger, runs a multi-image photometry pass, and asserts both survive. It fails against the old code (handlers cleared) and passes now. Existing logging output tests still pass, confirming logging itself is unaffected.

### DIFF
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -45,6 +45,11 @@ __all__ = [
 # Allowed FITS header keywords for exposure values
 EXPOSURE_KEYWORDS = ["EXPOSURE", "EXPTIME", "TELAPSE", "ELAPTIME", "ONTIME", "LIVETIME"]
 
+# Attribute used to tag logging handlers that stellarphot itself created, so
+# that we can later remove only those and leave any caller-supplied handlers
+# (and the root logger's handlers) untouched. See issue #153.
+_STELLARPHOT_HANDLER = "_stellarphot_handler"
+
 
 class AperturePhotometry(BaseModel):
     """Class to perform aperture photometry on one or more images"""
@@ -134,8 +139,19 @@ def _add_log_handlers(logger, logfile, console_log):
     file_handler : `logging.FileHandler` or None
         The file handler that was created, or ``None`` if ``logfile`` was
         ``None``. Callers use this to flush/close the file when finished.
+
+    Notes
+    -----
+    Handlers created here are tagged with ``_STELLARPHOT_HANDLER`` so that
+    `_remove_our_handlers` can later remove only the handlers stellarphot
+    added. Propagation is also disabled on ``logger`` so its messages are not
+    duplicated by handlers attached to ancestor loggers (such as the root
+    logger), which means we never need to touch handlers we did not create.
     """
     logger.setLevel(logging.INFO)
+    # Do not propagate to ancestor loggers; the handlers added below fully
+    # handle our output, so there is no need to clear the root logger's handlers.
+    logger.propagate = False
 
     file_handler = None
     if logfile is not None:
@@ -143,15 +159,35 @@ def _add_log_handlers(logger, logfile, console_log):
         file_handler = logging.FileHandler(logfile)
         file_handler.setFormatter(logging.Formatter("%(levelname)s - %(message)s"))
         file_handler.setLevel(logging.INFO)
+        setattr(file_handler, _STELLARPHOT_HANDLER, True)
         logger.addHandler(file_handler)
 
     # Set up logging to console if requested, otherwise effectively suppress output
     console_handler = logging.StreamHandler() if console_log else logging.NullHandler()
     console_handler.setFormatter(logging.Formatter("%(message)s"))
     console_handler.setLevel(logging.INFO)
+    setattr(console_handler, _STELLARPHOT_HANDLER, True)
     logger.addHandler(console_handler)
 
     return file_handler
+
+
+def _remove_our_handlers(logger):
+    """
+    Remove only the handlers stellarphot added to ``logger``.
+
+    Handlers created by `_add_log_handlers` are tagged with
+    ``_STELLARPHOT_HANDLER``; any other handlers (added by the caller or by
+    other libraries) are left in place.
+
+    Parameters
+    ----------
+    logger : `logging.Logger`
+        Logger to clean up.
+    """
+    for handler in logger.handlers[:]:
+        if getattr(handler, _STELLARPHOT_HANDLER, False):
+            logger.removeHandler(handler)
 
 
 def single_image_photometry(
@@ -653,8 +689,8 @@ def single_image_photometry(
         fh.flush()
         fh.close()
 
-    # Remove logger handler
-    logger.handlers.clear()
+    # Remove only the handlers we added
+    _remove_our_handlers(logger)
 
     # Create PhotometryData object to return
     photom_data = PhotometryData(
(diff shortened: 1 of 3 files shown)


stellarphot/photometry/tests/test_photometry.py
```diff
@@ -568,6 +568,94 @@ def test_photometry_on_directory(self, coords, photometry_settings_for_test):
...
+                # multi_image_photometry writes the log file into the image dir.
+                full_logfile = Path(temp_dir) / "test.log"
+
+                with warnings.catch_warnings():
+                    warnings.filterwarnings(
+                        "ignore",
+                        message="Cannot merge meta key",
+                        category=MergeConflictWarning,
+                    )
+                    ap_phot = AperturePhotometry(settings=photometry_settings_for_test)
```
JuanCab, line 649: It is very unclear where `test_multi_image_photometry_preserves_foreign_log_handlers` actually calls `multi_image_photometry` to make the test happen.  Took me a second to remember `AperturePhotometry` automatically handles this.  Might be worth some comments to clarify this.

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -568,6 +568,94 @@ def test_photometry_on_directory(self, coords, photometry_settings_for_test):
...
+        # present.
+        num_files = 3
+        fake_images = self.list_of_fakes(num_files)
+
+        noise_unit = photometry_settings_for_test.camera.read_noise.unit
+        photometry_settings_for_test.camera.read_noise = (
+            fake_images[0].noise_dev * noise_unit
+        )
+
+        root_logger = logging.getLogger()
```
JuanCab, line 599: Add comments to explain why you are creating these loggers when we want to test the ones in `multi_image_photometry` with this test.  As I gather it, you are creating these loggers to make sure they are not eliminated by `multi_image_photometry`, but it is not clear.

stellarphot/photometry/tests/test_photometry.py
JuanCab, line 1: Given you have to create all these fake images for multi-photometry testing, almost seems like a fake image generator should be a built-in for testing to avoid duplication in the testing code.  It shouldn't be part of this PR, but thinking it might make sense if we have to do this (probably a needless optimization for now).

stellarphot/photometry/tests/test_photometry.py
```diff
@@ -568,6 +568,94 @@ def test_photometry_on_directory(self, coords, photometry_settings_for_test):
...
+                with warnings.catch_warnings():
+                    warnings.filterwarnings(
+                        "ignore",
+                        message="Cannot merge meta key",
+                        category=MergeConflictWarning,
+                    )
+                    ap_phot = AperturePhotometry(settings=photometry_settings_for_test)
+                    ap_phot(temp_dir, object_of_interest=object_name)
+
+                # single_image_photometry must still write to the log file even
```
JuanCab, line 656: Are you calling the `AperturePhotometry` with a single image?  A little unclear on this.

stellarphot/photometry/photometry.py
JuanCab, line 1: So happy you found a way to tag the logging handler we use.  I was so new to logging when I implemented this (still am) that I just wasn't sure how to approach it.

JuanCab (changes requested):
Oh sweet Flying Spaghetti Monster, thank you for fixing this... It bugged the hell out of me that I couldn't use the logging without nuking all the logging handlers when I closed the logging.

It does look this this incorporates the changes from #152 but is currently conflicting with it.  It looks like it is from the same git branch but further down.

stellarphot/photometry/tests/test_photometry.py
mwcraig: Good call — the dispatch is implicit. `AperturePhotometry.__call__` (photometry.py:99-101) does `path = Path(file_or_directory); if path.is_dir(): multi_image_photometry(...)`. The test passes `temp_dir` (a directory of 3 fake images), so `ap_phot(temp_dir, ...)` routes into `multi_image_photometry`, which then calls `single_image_photometry` per image. I'll add a comment at the call site spelling this out. (Not applied locally yet — happy to push it to the branch.)


stellarphot/photometry/tests/test_photometry.py
mwcraig: Exactly right. The two `NullHandler`s — one on the root logger, one on the `multi_image_photometry` named logger — are "foreign" handlers added by something other than stellarphot, standing in for handlers a downstream app/user might have installed. The test asserts they're still present after the run (lines 654-655), i.e. that we no longer nuke handlers we didn't create. I'll add a comment making that intent explicit. (Not applied locally yet — can push to the branch.)


stellarphot/photometry/tests/test_photometry.py
mwcraig: Agreed, and good to keep it out of this PR. Worth noting a shared generator already exists: `FakeImage`/`FakeCCDImage` in `stellarphot/photometry/tests/fake_image.py`, used by both `test_photometry.py` and `test_detection.py`. The only photometry-test-local duplication is the `list_of_fakes` helper (test_photometry.py:168); promoting that into `fake_image.py` or a fixture would be a small self-contained follow-up. Leaving it out of this PR per your note.


stellarphot/photometry/tests/test_photometry.py
mwcraig: No — `temp_dir` is a directory containing 3 fake images (`num_files = 3`), so `__call__` takes the `path.is_dir()` branch and dispatches to `multi_image_photometry` (which calls `single_image_photometry` once per image internally). I'll fold this into the clarifying comment from the note above.


stellarphot/photometry/photometry.py
mwcraig: Thanks! For reference, the mechanism tags every handler stellarphot adds with a sentinel attribute (`_STELLARPHOT_HANDLER`) so `_remove_our_handlers` removes only our own and leaves foreign handlers untouched; we also set `logger.propagate = False` so we never need to touch the root logger's handlers at all. Glad it scratches the #153 itch.


mwcraig: Sure wish I had actually been the one to figure it out 😂

stellarphot/photometry/tests/test_photometry.py
JuanCab: Looks good

stellarphot/photometry/tests/test_photometry.py
JuanCab: OK, much clearer.

JuanCab (approved):
Looks good

PR #581, opened 2026-07-04
Replace batman-package with pytransit for transit fitting

mwcraig:
Switches the transit-fitting light-curve engine from `batman-package` to [`pytransit`](https://github.com/hpparvi/PyTransit). `pytransit`'s `RoadRunnerModel('quadratic')` maps directly onto the existing `a/Rs` + inclination fit parameters, so this is a narrow change confined to the transit-fitting module and its test file. Only a circular-orbit, quadratic-limb-darkening light curve is used.

## What changed

- **`stellarphot/transit_fitting/core.py`** — `TransitModelFit` now uses a `RoadRunnerModel("quadratic")` instead of a `batman.TransitParams` container + rebuilt `TransitModel`. The `times` setter calls `set_data`; the fit closure calls `evaluate(k, ldc, t0, p, a, i=radians(inclination))`; `model_light_curve`'s `at_times` branch swaps/restores via `set_data`. pytransit stays an optional dependency: the module imports without it and `TransitModelFit()` raises a clear error only if it is missing.
- **Regression testing** — `generate_batman_reference.py` writes a committed reference fixture (`data/batman_reference_lightcurves.ecsv`) once; the new `test_pytransit_matches_batman_reference` asserts pytransit reproduces it to `atol=1e-4` (measured agreement ~1e-5). `batman` is not a test dependency — pytransit is installed via the tox test/coverage envs.
- The synthetic-data test helper now generates ground truth through the pytransit-backed model; `importorskip` switched to pytransit.
- Removed the Windows-only test skip.
- **Deps/docs** — `pyproject.toml` `exoplanet` extra, CI workflow, `tox.ini`, `README.md`, and the architecture diagrams.

## Notes for reviewers

- **pytransit's dependency footprint is larger than batman's** — it pulls in numba, pyopencl, celerite, tables, blosc2. No compiler needed and it is conda-forge-installable.
- **The Windows CI skip was removed** — please keep an eye on the Windows job; easy to restore if numba misbehaves there.


stellarphot/transit_fitting/tests/generate_batman_reference.py
```diff
@@ -0,0 +1,114 @@
+"""
```
mwcraig: No need to include this in the pr

stellarphot/transit_fitting/core.py
```diff
@@ -636,15 +602,11 @@ def model_light_curve(self, at_times=None, detrend_by=None):
...
-            self._batman_mod_for_fit = batman.TransitModel(
-                self._batman_params, at_times
-            )
+            # temporarily point the transit model at the new times,
+            # then restore the original times.
+            self._transit_model.set_data(np.asarray(at_times, dtype=float))
             model = self.model(at_times, airmass, width, spp)
-            self._batman_mod_for_fit = original_model
+            self._transit_model.set_data(np.asarray(self.times, dtype=float))
         else:
```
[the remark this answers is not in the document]
mwcraig: Confirmed — both issues are real, and I reproduced them against this branch.

**1. Length mismatch.** The trend covariate arrays are sized to `self.times`, but the model flux comes out sized to `at_times`, so the `flux += airmass_trend * airmass` lines in `transit_model_with_trends` broadcast-error whenever `len(at_times) != len(self.times)`. This fails even in the no-trend case (the `zeros` fallback is still `np.zeros_like(self.times)`), and even the docstring's own example (`at_times=np.linspace(..., num=1000)`) triggers it:

```
ValueError: operands could not be broadcast together with shapes (1000,) (100,) (1000,)
```

The existing test only passes `at_times` of length `len(tmod.times)`, so it never exercised this.

**2. Exception safety.** Worse than "left pointing at the wrong time array" — the instance is left *corrupted*. After a failed different-length `at_times` call, `_transit_model` still points at the new times, so a subsequent plain `model_light_curve()` (which calls `evaluate()` internally) also raises. Confirmed by running a failing `at_times=...num=1000` call and then a bare `model_light_curve()`, which then blew up on line 611.

A `try/finally` handles the restore. For the length issue, the honest fix is that we have no airmass/width/spp covariates at arbitrary new times, so the model at those times should be transit-only (trends zeroed to `len(at_times)`):

```python
if at_times is not None:
    eval_times = np.asarray(at_times, dtype=float)
    z = np.zeros_like(eval_times)
    self._transit_model.set_data(eval_times)
    try:
        # no covariates exist at arbitrary new times -> transit model only
        model = self.model(eval_times, z, z, z)
    finally:
        self._transit_model.set_data(np.asarray(self.times, dtype=float))
else:
    zeros = np.zeros_like(self.times)
    airmass = self.airmass if self.airmass is not None else zeros
    width = self.width if self.width is not None else zeros
    spp = self.spp if self.spp is not None else zeros
    model = self.model(self.times, airmass, width, spp)
```

One open design question for @mwcraig: the docstring example combines `at_times=...` with `detrend_by=['airmass']`. `_detrend` also reads `getattr(self, trend)` (length `self.times`), so detrending at new times has the same mismatch and no well-defined covariates. Options are (a) forbid `detrend_by` together with a differently-sized `at_times` with a clear error, or (b) require the caller to supply matching covariates. Happy to implement whichever you prefer, plus a regression test with a genuinely different `at_times` length.

stellarphot/transit_fitting/tests/generate_batman_reference.py
mwcraig: Agreed. The test `test_pytransit_matches_batman_reference` only reads the committed fixture (`data/batman_reference_lightcurves.ecsv`) and never imports batman, so this generator isn't needed for the suite to run.

Two dangling references to clean up if we drop it:
- `test_transit_model_fit.py:45` — the comment "generated once with batman (see generate_batman_reference.py)"
- `data/batman_reference_lightcurves.ecsv:9` — metadata "Regenerate with generate_batman_reference.py."

Rather than lose the provenance entirely, one option is to move the how-it-was-generated note (and the batman snippet) into the fixture's `meta`/header comment and delete the standalone file. Want me to remove `generate_batman_reference.py` and fix those two references on the branch?

stellarphot/transit_fitting/core.py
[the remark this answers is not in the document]
mwcraig: How about b, with an error raised if the lengths don't match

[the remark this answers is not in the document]
mwcraig: Sounds good — going with (b): `at_times` reuses the stored `airmass/width/spp` covariates, so it must match the model's time length, and a mismatch raises a clear `ValueError` (checked *before* any `set_data`, so no corruption) rather than a broadcast crash. Wrapping the evaluate in `try/finally` too, and adding a regression test with a genuinely different length. Implementing on the branch now.

stellarphot/transit_fitting/tests/generate_batman_reference.py
mwcraig: 👍 taken as a yes — removing `generate_batman_reference.py` and fixing the two references (folding the provenance note into the fixture header) on the branch now.

stellarphot/transit_fitting/core.py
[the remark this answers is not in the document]
mwcraig: Implemented and pushed (68893a4): `at_times` now requires `len(at_times) == len(self.times)` and raises a clear `ValueError` before repointing the model, with a `try/finally` restore and a regression test for the mismatch case.

The trade-off is that you can no longer evaluate on a denser grid than the data for a smooth overlay when trends are set. Tracked as a follow-up in feder-observatory/stellarphot#582 (likely resolved by interpolating the covariates onto `at_times`).

stellarphot/transit_fitting/core.py
```diff
@@ -137,11 +143,9 @@ class TransitModelFit:
     """
     Transit model fits to observed light curves.
 
-    Parameters
-    ----------
-    batman_params : batman.TransitParams
-        Parameters for the batman transit model. If not provided, the
-        default parameters will be used.
+    The underlying transit light curve is computed with pytransit's
```
mwcraig, line 146: Make this a sphinx-style link and add an inter sphinx entry to pytransit if needed

mwcraig: Oops, I meant this for RoadRunnerModel below

stellarphot/transit_fitting/tests/test_transit_model_fit.py
```diff
@@ -1,11 +1,15 @@
...
 import numpy as np
 import pytest
+from astropy.table import Table
 
 from stellarphot.transit_fitting import TransitModelFit
 
-batman = pytest.importorskip("batman")
+pytest.importorskip("pytransit")
+
+DATA_DIR = Path(__file__).parent / "data"
```
mwcraig: Any reason not to use astropy's get package data for this? It may be that there is no need to switch to that either...

stellarphot/transit_fitting/core.py
mwcraig: Looked into this for `RoadRunnerModel`. Short version: an intersphinx `:class:` cross-reference **won't work**, so the honest fix is a plain external link.

Details I checked:

- Intersphinx is already enabled (we inherit `sphinx.ext.intersphinx` + the default `intersphinx_mapping` from `from sphinx_astropy.conf.v2 import *` in `docs/conf.py`); pytransit is just not in the mapping.
- pytransit *does* publish an inventory — `https://pytransit.readthedocs.io/en/latest/objects.inv` returns 200. But decompressing it shows it contains **only** `std:label` (123) and `std:doc` (26) entries and **zero `py:` domain entries** — i.e. pytransit's docs are hand-written notebooks with no autodoc/API pages, so there is no `pytransit.RoadRunnerModel` object to point a `:class:` role at. Adding it to `intersphinx_mapping` wouldn't give us a resolvable target.

So the best we can do is a regular external hyperlink from the docstring. Proposed edit at `core.py:146-148`:

```rst
The underlying transit light curve is computed with pytransit's
`RoadRunnerModel
<https://pytransit.readthedocs.io/en/latest/notebooks/models/roadrunner/roadrunner_model_example_1.html>`_
using a quadratic limb-darkening law and a circular orbit.
```

That target (the "RoadRunner transit model example I — basics" page) *is* in the inventory, so it's a stable link. Alternatives if you'd rather: the pytransit landing page, or the GitHub repo.

Want me to make it an external link like the above (and skip the intersphinx entry, since it can't resolve the class)? Happy to point it at whichever URL you prefer.

stellarphot/transit_fitting/core.py
```diff
@@ -636,15 +605,24 @@ def model_light_curve(self, at_times=None, detrend_by=None):
         spp = self.spp if self.spp is not None else zeros
```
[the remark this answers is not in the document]
mwcraig: Valid — confirmed against the branch. `model_light_curve()` has no guards, while `fit()` checks both preconditions (`core.py:481-484`):

```python
if self.times is None:
    raise ValueError("The times must be set before trying to fit.")
if self._model is None:
    raise ValueError("Run setup_model() before trying to fit.")
```

Called cold, `model_light_curve()` hits `zeros = np.zeros_like(self.times)` with `self.times is None` (that quietly yields `array(None, dtype=object)` rather than erroring), then `self.model(...)` dereferences `self._model`, which is `None`, giving the cryptic `TypeError: 'NoneType' object is not callable`. `BIC` calls `model_light_curve()` too, so it inherits the same failure.

Proposed fix — mirror `fit()`'s guards at the top of `model_light_curve()`:

```python
if self.times is None:
    raise ValueError("The times must be set before computing a model light curve.")
if self._model is None:
    raise ValueError("Run setup_model() before computing a model light curve.")
```

Want me to add these on the branch (with a small test that each raises `ValueError` when called before `times`/`setup_model()`)?

stellarphot/transit_fitting/tests/test_transit_model_fit.py
mwcraig: Good call — and it actually matches the established convention. Almost every other test that loads a fixture uses astropy's helper rather than `Path(__file__)`: `conftest.py`, `tests/test_catalogs.py`, `io/tests/test_aavso.py`, `io/tests/test_aij_io.py`, `differential_photometry/tests/test_vsx_mags.py`, `settings/tests/test_models.py` (via `get_pkg_data_path`), etc. This test file is the odd one out.

There's no functional reason to keep the `Path` version, so switching is a consistency win:

```python
from astropy.utils.data import get_pkg_data_filename

REFERENCE_FILE = get_pkg_data_filename("data/batman_reference_lightcurves.ecsv")
```

`get_pkg_data_filename` resolves paths relative to the calling package, so from `transit_fitting/tests/` it maps to `transit_fitting/tests/data/…` as intended, and it returns a `str` that `Table.read` accepts — the `DATA_DIR`/`Path` lines can then go away.

Want me to make this swap on the branch?

stellarphot/transit_fitting/core.py
mwcraig: The link you propose is fine,

mwcraig: Done in 53895c2 — the docstring now links `RoadRunnerModel` to the RoadRunner example page:

```rst
The underlying transit light curve is computed with pytransit's
`RoadRunnerModel
<https://pytransit.readthedocs.io/en/latest/notebooks/models/roadrunner/roadrunner_model_example_1.html>`_
using a quadratic limb-darkening law and a circular orbit.
```

No intersphinx entry added, since pytransit's inventory has no `py:` objects for a `:class:` role to resolve.

stellarphot/transit_fitting/core.py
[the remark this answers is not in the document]
mwcraig: Done in 53895c2. `model_light_curve()` now fails fast, mirroring `fit()`:

```python
if self.times is None:
    raise ValueError("The times must be set before computing a model light curve.")
if self._model is None:
    raise ValueError("Run setup_model() before computing a model light curve.")
```

Added `test_model_light_curve_requires_times_and_model`, which checks both the times-unset and model-unset paths raise `ValueError` (it was confirmed red against the old `TypeError` behaviour before the guards went in).

stellarphot/transit_fitting/tests/test_transit_model_fit.py
mwcraig: Done in 53895c2:

```python
from astropy.utils.data import get_pkg_data_filename

REFERENCE_FILE = get_pkg_data_filename("data/batman_reference_lightcurves.ecsv")
```

The `Path`/`DATA_DIR` lines (and the now-unused `pathlib` import) are gone, and the full transit-fitting suite still passes (16/16).

Issue #222, opened 2023-12-19
Fix linting issues in photometry


mwcraig:
For all of [these issues](https://github.com/feder-observatory/stellarphot/issues?q=is%3Aopen+is%3Aissue+label%3Alinting), first delete the appropriate line in `pyproject.toml` that excludes the subpackage from checks. for `photometry` you would delete [this line](https://github.com/feder-observatory/stellarphot/blob/main/pyproject.toml#L123):

```toml
"stellarphot/photometry/*" = ["F"]
```

After that, run `ruff` either with `tox -e lint` or `ruff .`. The errors you get should be the ones -- fix them! If it is an unused function argument make sure to search the project for calls to that function to update the arguments there too.

- [ ] stellarphot/photometry/tests/test_photometry.py:526:9: F841 Local variable `expected_deviation` is assigned to but never used
- [ ] stellarphot/photometry/tests/test_photometry.py:578:13: F841 Local variable `phot_data` is assigned to but never used
- [ ] stellarphot/photometry/tests/test_photometry.py:637:13: F841 Local variable `phot_data` is assigned to but never used

Tanner728:
I would also like to work on this one.

mwcraig:
My best guess is that in these cases the lines with the unused variables can just be deleted...don't know for sure.

mwcraig:
This needs to wait until #236 is merged, since that PR adds some more photometry functions

PR #584, opened 2026-07-05
Migrate to the astrowidgets bqplot viewer, drop ginga

mwcraig:
## Summary

astrowidgets 0.5.0 replaced the ginga backend stellarphot used with a new bqplot-backed viewer implementing the [astro-image-display-api](https://github.com/astropy/astro-image-display-api) protocol, and its legacy `astrowidgets.ginga` module no longer imports against astro-image-display-api 0.2.1. This PR migrates the seeing-profile and comparison-star viewers to `astrowidgets.bqplot.ImageWidget` throughout and removes the direct ginga dependency.

**Update 2026-07-21:** the branch now targets the slimmed bqplot backend that lands after astropy/astrowidgets#215/#221/#222 (to be released as astrowidgets 0.6), rather than 0.5.1. That backend removed the `ImageWidget.viewer` property (stellarphot now reaches the figure through `_astro_im`), draws catalog marks with the SVG `Scatter` (all documented shapes render, style size honored), and made clicks safe — so the stellarphot-side workaround module written for 0.5.x is deleted again in full.

Commits are in strict TDD order: the first commit rewrites the tests (red at that commit — the old code fails at collection against astrowidgets 0.5.0 because the pre-0.4 import no longer exists), and the production migration follows.

## API mapping

| Old (ginga backend) | New (bqplot backend) |
|---|---|
| `load_fits` | `load_image` |
| `add_markers`/`get_markers`/`remove_markers` | `load_catalog`/`get_catalog`/`remove_catalog` with `catalog_style` |
| `zoom_level = "fit"` / zoom math | `set_viewport(fov=...)` (accepts angular `Quantity` directly) |
| `center_on` | `set_viewport(center=...)` |
| ginga canvas `cursor-down` callback | `iw._astro_im.interaction.on_msg` dispatcher filtering `"click"` events |
| `set_keybindings` | deleted — bqplot's PanZoom pans/zooms natively; no keybinding API |
| ginga `Text` marks for star labels | native `bqplot.Label` marks, tracked in `ComparisonViewer._label_marks` |

## User-visible changes

- **Plain click** (not shift-click) toggles star exclusion in the comparison viewer. The bqplot click payload carries no modifier-key info; drag-pan does not emit a click, so there is no conflict. Help text and legend updated.
- APASS comparison stars are **red diamonds**; exclusion markers are red plus signs.
- Catalog markers render at the size the catalog style asks for (the 0.5.x backend hardcoded one large size).
- The "Click closer to a star" message is now actually visible — it goes to a status line under the image. The old code sent it to an `ipw.Output` that was never added to the layout (disclosed bugfix).
- The target circle no longer leaks into the aperture file written by the comparison viewer (disclosed bugfix — previously the circle marker could end up in `generate_table()` output).

## Relationship to upstream

The 0.5.x workarounds this PR originally carried (`click_center`/`is_marking` neutralization for astropy/astrowidgets#206, a `load_catalog` wrapper suppressing a spurious traittypes warning and re-applying the ignored style size, and a ScatterGL→Scatter swap so diamond/plus shapes draw) are all fixed in the post-#222 backend and have been removed here.

Still handled on the stellarphot side:

- Catalog tables are loaded with their SkyCoord column renamed to `coord`, astro-image-display-api's default, so catalogs read back from the viewer combine cleanly.
- Catalogs built from astroquery results carry Masked coordinate arrays that the high-level WCS API rejects; coordinates are unmasked before loading.
- Star labels and the 2.5-arcmin target circle are bqplot `Label`/`Lines` marks mirrored into the private `_astro_im._scatter_marks` dict (under reserved `__label__*`/`__circle__*` keys) so they survive the backend's `_update_marks` calls. This — plus the `interaction` hookup — is a deliberate dependence on astrowidgets private internals; flagging it here so the next backend refactor doesn't silently break it.

## Packaging

- `[gui]` extra: `astrowidgets >=0.6` + `astro-image-display-api >=0.3.0` + explicit `bqplot >=0.13,<0.14` (stellarphot imports `Label`/`Lines` directly); `ginga` removed.
- **CI will stay red until astrowidgets 0.6.0 is on PyPI** — the pin is intentionally ahead of the latest release (0.5.1). No stellarphot change is needed once 0.6.0 ships.
- `.rtd-environment.yml` installs astrowidgets/astro-image-display-api from PyPI (conda-forge tops out at astrowidgets 0.3.0 and does not package astro-image-display-api).
- `filterwarnings`: dropped the ginga/old-astrowidgets ignores; the traittypes ignore added for 0.5.x is gone again (the backend now converts catalog columns with `np.asarray` before they reach the bqplot traits).

## Testing

Run against astrowidgets at the tip of astropy/astrowidgets#222 (`e97aa8a`, installed from a local clone) with bqplot 0.13.1, bqplot-gl 0.1.1, astro-image-display-api 0.3.0:

- Offline suite: 784 passed, with `filterwarnings = error` guarding the removed ignore.
- `stellarphot/gui --remote-data=any`: all 129 pass, including the end-to-end `ComparisonViewer` tests against VizieR.
- Import isolation tests pass (base install stays headless).
- Notebooks need no changes (`02-seeing-and-comparison.ipynb` only instantiates `ComparisonAndSeeing`).

Manual Jupyter QA of the click/pan/zoom/label/TESS-view flows still needs a human pass (headless tests cannot see front-end rendering).

https://claude.ai/code/session_01UoLkBuZUbs5kJY94pbsrdT
https://claude.ai/code/session_01JcJuqpQTxx3Em7yCCmahXP


stellarphot/gui/comparison_functions.py
```diff
@@ -771,15 +830,16 @@ def show_circle(self, radius=2.5 * u.arcmin, pixel_scale=0.56 * u.arcsec / u.pix
             Circle is shown around the target.
         """
         radius_pixels = np.round((radius / pixel_scale).to(u.pixel).value, decimals=0)
-        orig_marker = self.iw.marker
-        self.iw.marker = {"color": "yellow", "radius": radius_pixels, "type": "circle"}
-        self.iw.add_markers(
-            Table(data=[[self.target_coord]], names=["coords"]),
-            skycoord_colname="coords",
+        self.iw.load_catalog(
```
mwcraig: Do this by drawing a circle in bqplot instead of using a marker

stellarphot/gui/seeing_profile_functions.py
```diff
@@ -192,12 +124,21 @@ def __init__(
...
+        # handler references the attributes below, which are never
+        # initialized, so any click raises AttributeError and prevents
+        # callbacks registered later (like ours) from running. Setting both
+        # to False makes the built-in handler a no-op.
+        self.iw.click_center = False
+        self.iw.is_marking = False
+
+        # Match the behavior of the old ginga-backed widget, in which
+        # scroll-zoom was off by default.
+        self.iw.viewer.set_scroll_zoom(False)
```
mwcraig: Turns out that bqplot always has scroll zoom and you cannot change it.

stellarphot/gui/tests/test_seeing_profile.py
```diff
@@ -220,18 +192,74 @@ def test_seeing_profile_error_messages_no_star(tmp_path):
...
+        profile_widget.iw.viewer.interaction,
+        {"event": "mousemove", "domain": {"x": star_loc_x, "y": star_loc_y}},
+        [],
+    )
+
+    # No profile should have been computed
+    assert profile_widget.rad_prof is None
+
+
+def test_builtin_click_handler_is_neutralized():
```
mwcraig: Is this test repeated somewhere else?

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{3.10,3.11,3.12}-test
+    py{3.12,3.13}-test
```
mwcraig: Add a 3.14 test as long as we are making changes here. Also update the GitHub workflow.

stellarphot/gui/comparison_functions.py
```diff
@@ -37,15 +33,98 @@
...
+        Table with a "coords" column of `astropy.coordinates.SkyCoord`.
+
+    Returns
+    -------
+
+    `astropy.table.Table`
+        Copy of the table with the column renamed.
+    """
+    out = Table(table)
+    out.rename_column("coords", "coord")
```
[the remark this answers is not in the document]
mwcraig: Partially agree. There is no failure in practice today: every call site reaches this helper through `make_markers`, and the tables passed there (`targets_from_file`, `set_up()`'s VSX table, and the APASS output of `crossmatch_APASS2VSX`/`in_field`) all use stellarphot's `coords` column convention — the docstring documents that contract. However, `make_markers` is exported in `__all__`, so an external caller could reasonably pass a table that already uses the API's `coord` name and get an unhelpful `KeyError` from `rename_column`. Hardening the helper is cheap:

```python
if "coord" not in out.colnames:
    if "coords" not in out.colnames:
        raise ValueError(
            "table must have a 'coords' (or 'coord') SkyCoord column"
        )
    out.rename_column("coords", "coord")
```

stellarphot/gui/comparison_functions.py
```diff
@@ -64,117 +143,102 @@ def make_markers(iw, RD, vsx, ent, name_or_coord=None):
...
-                        try:
-                            imagewidget.remove_markers_by_name(marker_name=elim)
-                        except AttributeError:
-                            imagewidget.remove_markers(marker_name=elim)
+        all_table = _all_catalog_entries(imagewidget)
+        if len(all_table) == 0:
+            status_widget.value = "Click closer to a star"
+            return
 
+        index, d2d, d3d = out_skycoord.match_to_catalog_sky(all_table["coord"])
```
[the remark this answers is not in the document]
mwcraig: Confirmed. `status_widget.value` is set on a miss (both the empty-catalog early return and the >10 arcsec branch) and never reset, so a stale "Click closer to a star" stays visible after a later successful toggle. Clearing it at the top of the match branch (or right after the `len(all_table) == 0` check succeeds) fixes it:

```python
if d2d < 10 * u.arcsec:
    status_widget.value = ""
    ...
```

Will apply on the branch.

stellarphot/gui/seeing_profile_functions.py
```diff
@@ -454,117 +395,115 @@ def _make_tess_box(self):
...
+                # tests.
+                self.error_console.append_display_data(
+                    ipw.HTML(
+                        "<strong>No star found at this location. "
+                        "Try clicking closer "
+                        "to a star or on a brighter star</strong>"
+                    )
                 )
-                update_aperture_settings = True
+                print(f"{self.error_console.outputs=}")
```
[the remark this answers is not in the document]
mwcraig: Agreed — this is a leftover debug print. The message already reaches the user through `error_console`, and the regression test asserts on `error_console.outputs`, not stdout, so nothing depends on it. Will remove it on the branch.

stellarphot/gui/comparison_functions.py
mwcraig: Agreed, and there's a second reason to do this: astrowidgets 0.5.0 does not honor `size` in `catalog_style`, so `size=radius_pixels` doesn't render at the intended radius anyway — the circle is currently wrong, not just inelegant.

Plan: build a bqplot `Lines` mark tracing the circle (parametric points in the image's pixel scales, converted from `target_coord` + `radius` via the WCS) and inject it into `viewer._scatter_marks` under a reserved key, the same pattern `show_labels`/`remove_labels` already use for the `__label__*` marks, so it survives the viewer's `_update_marks()` calls. `remove_circle` then drops that key instead of removing a catalog. Will implement on the branch.

stellarphot/gui/seeing_profile_functions.py
mwcraig: Confirmed, and the plumbing explains why: `set_scroll_zoom(False)` only sets `allow_zoom = False` on the `PanZoom` instance, but that `PanZoom` is attached as the `next` interaction behind bqplot-image-gl's `MouseInteraction`, which is the figure's actual interaction — wheel handling happens in the image-gl frontend, so the flag never takes effect. So this line is a no-op and its "match ginga behavior" comment is misleading. I'll remove both here and in `ComparisonViewer`, and note the always-on scroll zoom as a behavior change in CHANGES.rst / the migration doc. Probably also worth an upstream astrowidgets issue alongside the click-handler one.

stellarphot/gui/tests/test_seeing_profile.py
mwcraig: Close but not a duplicate: `test_comparison_functions.py::test_click_center_and_marking_neutralized` (line 60) checks the same workaround, but for `ComparisonViewer.__init__`, while this one guards `SeeingProfileWidget.__init__` — the two constructors apply the fix independently, so each test protects its own class. This one also goes a step further by loading an image and checking that `iw._mouse_click(...)` is actually a no-op rather than raising. If you'd like, I can consolidate them into a single parametrized test over the two widget classes in a shared module, but keeping one per class next to the class's other tests also seems fine.

tox.ini
mwcraig: Will do. Plan: add `py3.14` to the tox envlist and `"3.14"` to the matrix in `.github/workflows/test.yml` — the workflow builds the tox env name from the matrix value (`tox -e ${{ matrix.python-version }}-test`), so those two edits are all that's needed. One thing CI will tell us: whether the GUI stack (bqplot / bqplot-image-gl / pytransit) all ship 3.14 wheels yet; if one doesn't, we can keep 3.14 in the matrix with `continue-on-error` until it does. Will push this to the branch.

stellarphot/gui/seeing_profile_functions.py
mwcraig: Do the fix here, but no need for an upstream issue.

stellarphot/gui/tests/test_seeing_profile.py
mwcraig: Consolidate

stellarphot/gui/comparison_functions.py
```diff
@@ -312,37 +387,49 @@ def _init(self):
...
+                name_or_coord=self.target_coord,
+            )
+        except Exception as err:
+            # An exception raised here is invisible to the user -- this runs in
+            # a widget callback, whose traceback only goes to the log -- so put
+            # the error on screen, then re-raise so the log gets the traceback.
+            error_message = ipw.HTML(
+                value="<p style='color: red'>Loading variable/comparison stars "
+                f"failed: {err}</p>"
+            )
```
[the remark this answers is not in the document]
mwcraig: Confirmed — real network errors frequently carry markup-significant characters (server response snippets, URLs with query strings). Fixed in 3dbdced: the message now uses `html.escape(str(err))`, and the existing failure-path test raises an error containing `<b>…</b> &` and asserts the escaped form appears while the raw markup does not.

stellarphot/core.py
```diff
@@ -9,6 +9,7 @@
     import lightkurve as lk
 import numpy as np
 import pandas as pd
+import requests
```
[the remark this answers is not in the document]
mwcraig: Agreed. `requests` was always present in practice (astroquery hard-depends on it, and `stellarphot.core` needs astroquery anyway), so the import could not actually fail — but a module-level import shouldn't lean on another package's dependency list. Declared it as a direct dependency in 8924558 with a comment noting why it's there (the commit also regenerates pixi.lock, which validates against the manifest).

PR #610, opened 2026-07-07
Join Gaia IDs to refcat2 by row index instead of row order

mwcraig:
Fixes #586

### The problem

`process_refcat2` attached Gaia DR2 IDs by uploading the full filtered refcat2 table (46 columns) to CDS XMatch and assigning `catalog["id"] = result["source_id"]` positionally. XMatch does **not** preserve input row order, so:

- when the result happened to have the same length as the input, most stars silently got the *wrong* Gaia ID (only ~2.8% of rows came back in place for a 1.8° field);
- when any star was unmatched or matched twice, the assignment crashed with `ValueError: Inconsistent data column lengths` — the failure reported in #586;
- the full-table upload itself was slow (~100 s for the 1.8° field) and flaky.

### The fix

- Upload only three columns to XMatch: a row index (`_sp_index`), `RA_ICRS`, and `DE_ICRS`, shrinking the upload from ~137 MB-scale to a few hundred kB.
- Join the result back to the catalog on the echoed index via a new unit-testable helper, `_attach_gaia_ids`: nearest match kept when a star matches more than one Gaia source, unmatched stars dropped with a warning saying how many.
- `process_refcat2` is extracted from its closure to module level (`_process_refcat2`) so the logic is testable without network.
- The XMatch call is timed and logged at info level (elapsed, rows uploaded, matches returned) to feed the decision about whether keeping Gaia IDs is worth the extra query.

### Tests

New network-free unit tests: shuffled XMatch result (every star must get its own ID), unmatched row (dropped + warning), duplicate match (nearest kept), and a monkeypatched `XMatch.query` asserting the slim upload and index round-trip. The remote-data `test_find_refcat2` now also asserts IDs are unique per star and that the star nearest the field center carries EY UMa's known Gaia DR2 source_id.

### Timing (EY UMa field, live runs with the fix)

| Radius | Total | XMatch | Vizier + rest | Stars |
|---|---|---|---|---|
| 10′ | 2.5 s | 0.9 s | 1.6 s | 187 |
| 0.5° | 4.7 s | 1.4 s | 3.3 s | 1,806 |
| 1.824° (the call from #586) | 42.4 s | 7.9 s | 34.5 s | 24,203 |

All uploaded stars matched (0 dropped), and a 200-star sample from the 1.824° run passed the Gaia HEALPix consistency check (`source_id // 2**35 == healpix12(position)`) at 100%.

*This PR was written by Claude Code at Matt's direction.*

### DIFF
--- a/stellarphot/catalogs.py
+++ b/stellarphot/catalogs.py
@@ -8,7 +8,13 @@ because they are data *retrieval* helpers rather than data-structure
 definitions. ``core.py`` holds the table classes only.
 """
 
+import logging
+import time
+import warnings
+
+import numpy as np
 from astropy import units as u
+from astropy.table import Table, unique
 from astroquery.xmatch import XMatch
 
 from .core import CatalogData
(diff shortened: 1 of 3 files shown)


stellarphot/catalogs.py
```diff
@@ -20,6 +26,113 @@
...
+
+    Returns
+    -------
+    `astropy.table.Table`
+        The matched rows of ``catalog``, in their original order, with a new
+        ``id`` column holding the Gaia source_id. Rows with no match are
+        dropped, with a warning saying how many.
+    """
+    # When one input row matches several Gaia sources, keep only the nearest.
+    xmatch_result = xmatch_result.copy()
```
mwcraig: Do we really need a copy here?

stellarphot/catalogs.py
```diff
@@ -20,6 +26,113 @@
...
+    n_unmatched = len(catalog) - len(nearest)
+    if n_unmatched > 0:
+        warnings.warn(
+            f"{n_unmatched} of {len(catalog)} catalog entries had no Gaia "
+            "match and have been dropped.",
+            stacklevel=2,
+        )
+
+    # nearest is sorted by index_col, so this preserves the original row order.
+    catalog = catalog[np.asarray(nearest[index_col])]
```
mwcraig, line 76: Is this seeking the rows that have matches?

stellarphot/catalogs.py
mwcraig: Strictly, no — it's there only because `Table.sort` sorts **in place**, so without the copy the helper would reorder the caller's `xmatch_result` as a side effect. The one production caller (`_process_refcat2`) never touches the result table after this call, so nothing would break today if the copy were dropped; it just keeps the helper from mutating its input, which felt worth the cost of copying a small 3-ish-column table. Happy to remove it if you'd rather.

stellarphot/catalogs.py
mwcraig: Yes, exactly. `nearest[index_col]` holds the catalog row numbers that came back from XMatch — one per matched star, deduped to the nearest match, and sorted ascending by the `sort` above. Indexing `catalog` with it does two things at once:

1. **Drops unmatched rows** — a catalog row with no match never appears in the result, so it isn't selected (that's what the `n_unmatched` warning above counts).
2. **Aligns the rows with `nearest`** — the selected rows come out in the same order as `nearest`, so the `source_id` assignment on the next line lines up row-for-row.

Since the line prompted the question, it probably deserves a fuller comment than the one it has — can expand it to say the above.

stellarphot/tests/test_catalogs.py
```diff
@@ -253,6 +263,140 @@ def test_find_refcat2(mag_limit, mag_limit_band):
     for band in ["GBP", "GRP", "GG", "SG", "SR", "SI", "SZ", "J", "H", "K"]:
         assert band in all_refcat2["passband"]
 
+    # The Gaia IDs must be unique per star. The catalog is tidy (one row per
+    # star per passband), so compare the number of distinct IDs to the number
+    # of distinct positions.
+    n_stars = len(set(ra.value for ra in all_refcat2["ra"]))
+    assert len(set(all_refcat2["id"])) == n_stars
```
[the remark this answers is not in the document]
mwcraig: Technically right, though two things soften it in practice:

- With float64 catalog RAs an exact collision is essentially impossible, and if one ever did occur the assertion would fail loudly (distinct IDs > distinct RAs) rather than silently pass — so this can't hide a bug, only produce an astronomically unlikely false failure.
- RA-only set comparison is the existing idiom in this very test (the `expected_all` check a few lines up does the same, with a comment to that effect).

That said, keying on `(ra, dec)` pairs is a one-line change with no downside, so happy to make it:

```python
n_stars = len(set(zip(all_refcat2["ra"].value, all_refcat2["dec"].value)))
```

stellarphot/tests/test_catalogs.py
```diff
@@ -253,6 +263,140 @@ def test_find_refcat2(mag_limit, mag_limit_band):
...
+
+    # Spot check: the star nearest the field center (EY UMa) must carry EY UMa's
+    # known Gaia DR2 source_id, i.e. the crossmatch attached the right ID to the
+    # right row. EY UMa's refcat2 rmag is 15.24, so it is only present in the
+    # runs without a magnitude limit.
+    if mag_limit is None:
+        ey_uma = SkyCoord.from_name("EY UMa")
+        cat_coords = SkyCoord(ra=all_refcat2["ra"], dec=all_refcat2["dec"])
+        nearest = cat_coords.separation(ey_uma).argmin()
```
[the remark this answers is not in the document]
mwcraig: Fair point — `SkyCoord.from_name` goes through the Sesame name resolver, which is a third external service unrelated to the Vizier/XMatch code under test, and EY UMa's coordinates are a constant. Agreed this should be a fixed coordinate next to the existing `EY_UMA_GAIA_DR2_ID` constant:

```python
# ICRS coordinates of EY UMa, from SIMBAD
EY_UMA_COORD = SkyCoord(ra=135.58650087 * u.deg, dec=49.81921088 * u.deg)
```

stellarphot/tests/test_catalogs.py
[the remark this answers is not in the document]
mwcraig: > only produce an astronomically unlikely false failure.

Nice one! Do not change to ra,dec just use ra

stellarphot/catalogs.py
mwcraig: Remove it

stellarphot/catalogs.py
mwcraig: Expand the comment

stellarphot/catalogs.py
mwcraig: Done in 3750dfc.

stellarphot/catalogs.py
mwcraig: Done in 3750dfc.

stellarphot/tests/test_catalogs.py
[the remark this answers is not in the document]
mwcraig: Done in 3750dfc.

Issue #426, opened 2024-08-14
Factor out the run-in-a-temporar-directory pytest fixture into conftest.py


mwcraig:
Right now there are several places (and one more on the way) where we repeat this block of code:

```python
    # This auto-used fixture changes the working directory to the temporary directory
    # and then changes back to the original directory after the test is done.
    @pytest.fixture(autouse=True)
    def change_to_tmp_dir(self, tmp_path):
        original_dir = os.getcwd()
        os.chdir(tmp_path)
        # Yielding here is important. It means that when the test is done, the remainder
        # of the function will be executed. This is important because the test is run in
        # a temporary directory and we want to change back to the original directory
        # when the test is done.
        yield
        os.chdir(original_dir)
```

We should move this to `conftest.py` (but will need to take out the `autouse=True` maybe). In places we are currently using it with `autouse=True` we will to keep a small fixture in place to specify the `autouse=True` part. 

If we take out the `autouse` thing in conftest then using the fixture would look like this (I think this is untested):

```python
@pytest.fixture(autouse=true)
def run_in_temp_dir(self, change_to_tmp_dir):
    pass
```

PR #611, opened 2026-07-07
Verify the USNO u'g'r'i'z' to SDSS ugriz transform and pin it down with tests

### LINKED ISSUE #536: Double-check the USNO to SDSS transform
The u'g'r'i'z' → ugriz transform in #532 is, I *think*, correct, but it would be good to take another look at it to confirm.

mwcraig:
Fixes #536

### Verification result: the transform is correct

The transformation matrix in `USNO_SDSS_to_SDSS_DR7.json` was re-derived independently from its cited reference, the [SDSS DR7 photometric equations page](https://classic.sdss.org/dr7/algorithms/jeg_photometric_eq_dr1.php#usno2SDSS). With all zpOffsets zero the published equations are

```
u = u'
g = g' + 0.060 * ((g' - r') - 0.53)
r = r' + 0.035 * ((r' - i') - 0.21)
i = i' + 0.041 * ((r' - i') - 0.21)
z = z' - 0.030 * ((i' - z') - 0.09)
```

Expanded into matrix form over `[u', g', r', i', z', 1]`, every entry agrees with the stored matrix to machine precision (max difference 8.7e-19). The usage in `transform_apass_bands` was also checked: the direction is right (APASS Sloan magnitudes are tied to the Smith et al. 2002 u'g'r'i'z' standards and the Jester transform expects 2.5m ugriz, so primed → SDSS DR7 → Jester is the correct chain), and the zero placeholders for u' and z' are harmless because the g/r/i output rows never reference those columns. **No coefficient changes were needed.**

### Why this PR still changes code

The existing unit test only exercised the constant terms — its input is zero in every band except u', so all eight color coefficients multiply zero (despite the test's comment claiming it "effectively checks the whole matrix"). Checked by mutation: a swapped 0.041/-0.035 pair, a sign flip on -0.06, and a 1.06 → 1.006 typo all pass the old test. This PR pins every coefficient down:

- **Fixed-point test** — a star with exactly the zeropoint colors (u'-g' = 1.39, g'-r' = 0.53, r'-i' = 0.21, i'-z' = 0.09) must transform to identical magnitudes.
- **Reference-equation test** — two stars with non-zeropoint colors are checked against the published equations evaluated directly in the test; also covers the 2-D `(n_bands + 1, n_stars)` input path.
- **Shape-error test** — input without the constant-term row raises instead of silently producing wrong magnitudes.

All three mutations above are caught by the new tests. Also corrects the `MatrixTransformMixin` docstring, which described the input shape backwards and omitted the required final row of ones.

### Caveat noted while verifying (not enforced by this PR)

The reference states the transformation is valid for stars in roughly `0.15 <= g'-r' <= 1.20` (and the standard system is poorly determined redder than M0). The code does not warn outside that range — fine for typical comparison-star use, possible follow-up if desired.

*This PR was written by Claude Code at Matt's direction.*

### DIFF
--- a/stellarphot/utils/tests/test_magnitude_system_transforms.py
+++ b/stellarphot/utils/tests/test_magnitude_system_transforms.py
@@ -236,8 +236,20 @@ class TestCatalogTransforms:
 
 
 class TestUSNOPrimeToSDSSDR7:
+    # The reference for the transform is
+    # https://classic.sdss.org/dr7/algorithms/jeg_photometric_eq_dr1.php#usno2SDSS
+    # which gives, with all zpOffsets zero,
+    #
+    #   u = u'
+    #   g = g' + 0.060 * ((g' - r') - 0.53)
+    #   r = r' + 0.035 * ((r' - i') - 0.21)
+    #   i = i' + 0.041 * ((r' - i') - 0.21)
+    #   z = z' - 0.030 * ((i' - z') - 0.09)
+
     def test_transform(self):
-        # Check that the zero point for ug is correct
+        # Check the constant terms: with all input magnitudes zero, each output
+        # is the product of -b and the zeropoint color (u' is checked too since
+        # its coefficient in the u row is one).
         inp_mag = np.asarray([1.39, 0.0, 0.0, 0.0, 0.0, 1.0])
         # This should produce this output:
         # u: same as u' so 1.39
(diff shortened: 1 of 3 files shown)


mwcraig (commented):
**Verdict: approve.** The central claim holds up under independent verification: the stored matrix in `USNO_SDSS_to_SDSS_DR7.json` was checked entry-by-entry against the SDSS DR7 reference page (b25 coefficients, zeropoint colors, all-zero zpOffset25, and the subtle detail that both r and i use the (r′−i′) color), and it matches exactly. The code diff is docstring-only, so behavior is unchanged. The new tests were mutation-tested: with three deliberate corruptions of the JSON (1.06→1.006, a sign flip on −0.06, and swapping 0.041/−0.035), the new value tests fail every time while the old `test_transform` passes them all — confirming the old test was vacuous for the color coefficients, exactly as the PR description says.

Minor findings, none blocking:

1. **nit** — `stellarphot/utils/magnitude_system_transforms.py:283` — the `ValueError` message says the matrix "does not match the number of passbands in the from_system (5)" but the required row count is passbands + 1. Now that a test pins this message, consider rewording to "expected {p} passband rows plus a final row of ones ({p+1} rows total)".
2. **nit** — no `CHANGES.rst` entry. This repo does record test-only/verification work under "Other Changes and Additions" (e.g. the #426 conftest entry), so a one-liner noting the transform was verified against its reference would fit the convention.
3. **nit** — the new value tests use `np.allclose` with default `rtol=1e-5`, which on ~13-mag values tolerates ~1e-4 mag; a fourth-decimal coefficient drift would slip through. `atol=1e-6, rtol=0` would pin the coefficients exactly.

Also confirmed (no action needed): the zero u′/z′ placeholders in `transform_apass_bands` are harmless — the g/r/i matrix rows have zero coefficients in the u′ and z′ columns and only `sdss_mags[1:4]` is consumed; and the reference's color validity cuts quoted in the PR are real (checked on the page), so the deferred-warning caveat is accurate.

*Verification: reference page fetched and equations expanded by hand; full test file run at the PR head in a scratch worktree (11 passed, 3 remote-data skipped); three JSON mutations run to confirm the new tests are red on each.*

stellarphot/utils/tests/test_magnitude_system_transforms.py
```diff
@@ -236,18 +236,77 @@ def test_refcats_adds_BVRI(self):
...
+    # which gives, with all zpOffsets zero,
+    #
+    #   u = u'
+    #   g = g' + 0.060 * ((g' - r') - 0.53)
+    #   r = r' + 0.035 * ((r' - i') - 0.21)
+    #   i = i' + 0.041 * ((r' - i') - 0.21)
+    #   z = z' - 0.030 * ((i' - z') - 0.09)
+
     def test_transform(self):
-        # Check that the zero point for ug is correct
```
mwcraig, line 240: Is this test still needed at all given the new tests added below? I gather the problem with this test is that the colors are zero?

stellarphot/utils/tests/test_magnitude_system_transforms.py
```diff
@@ -236,18 +236,77 @@ def test_refcats_adds_BVRI(self):
...
+        # Check the constant terms: with all input magnitudes zero, each output
+        # is the product of -b and the zeropoint color (u' is checked too since
+        # its coefficient in the u row is one).
```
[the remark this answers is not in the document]
mwcraig: Partially agree. The `u' = 1.39` entry does contradict "with all input magnitudes zero," so the wording should change. The final row of ones is not an input magnitude, though — it is the constant-term row the transform's API requires (see the `__call__` docstring), so the comment does not need to call it out as an exception.

Suggested rewording:

```python
# Check the constant terms: with all input magnitudes except u' zero,
# every color term vanishes, so each output is the product of -b and the
# zeropoint color; u' = 1.39 additionally checks that the u row passes
# u' through unchanged.
```

Note that in the thread below there is a question about whether this test should simply be deleted, which would make this moot.

stellarphot/utils/tests/test_magnitude_system_transforms.py
mwcraig: Yes, that's exactly the problem: with all magnitudes zero, every color (`g'-r'`, `r'-i'`, `i'-z'`) is zero, so all of the color coefficients are multiplied by zero and only the constant column (plus the pass-through of `u'`) is exercised. That is why the old "This effectively checks the whole matrix" claim was removed.

Whether it is still needed: it adds no realistic detection power beyond the new tests. Both stars in `test_transform_matches_reference_equations` have every component nonzero, so any single wrong entry in the stored matrix changes an output there and fails the `allclose` — including the constant terms this test checks. The only thing this test still buys is failure localization: its output reads the constant column directly, so a failure points straight at the constants/zeropoint colors instead of at "some coefficient somewhere."

Recommendation: it is safe to delete, and deleting it also moots Copilot's wording nit above. Happy to remove it on the branch if you want.

mwcraig: Let's drop this

mwcraig: Done — `test_transform` removed in 40401f6.

stellarphot/utils/tests/test_magnitude_system_transforms.py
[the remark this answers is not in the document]
mwcraig: Resolved by removing `test_transform` entirely in 40401f6 (see the thread below), so the comment this refers to no longer exists.

PR #613, opened 2026-07-07
Fix v1→v2 migration writing mag_inst = 0 for passbands with no matching column

mwcraig:
### Bug

When migrating v1 photometry data to v2, the merged `mag_inst` column was initialized with `np.zeros_like(new_data["date-obs"])`. Any observation whose `passband` had no matching `mag_inst_<passband>` column was left at that initial value, so it silently migrated with `mag_inst = 0` — no warning, no error. (As a bonus quirk, `zeros_like` on the `Time` column produced an object-dtype column of Python int zeros.)

### Fix

- Initialize the merged column with `np.full(len(new_data), np.nan)` so unmatched rows come out as NaN in a proper float column.
- Emit an `AstropyUserWarning` listing the passbands present in the data that have no matching `mag_inst_<passband>` column.

Includes a regression test (`test_migrate_v1_v2_missing_passband_column`) that migrates a two-filter v1 table with one magnitude column removed, and checks that the affected rows are NaN, the warning names the missing passband, and the other passband's magnitudes are untouched.

Fixes #596

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,6 +65,9 @@ Bug Fixes
 + ``comparison_utils.in_field`` no longer swaps the x/y image bounds, which
   excluded valid comparison-star candidates near the long edge of non-square
   images (and could include off-frame stars). [#612]
++ The v1 to v2 photometry migration now writes ``NaN`` and emits a warning for
+  observations whose passband has no matching ``mag_inst_<band>`` column,
+  instead of silently writing ``mag_inst = 0``. [#613]
 
 1.4.15 (2024-08-16)
 -------------------
--- a/stellarphot/utils/tests/test_version_migrator.py
+++ b/stellarphot/utils/tests/test_version_migrator.py
@@ -1,8 +1,10 @@
 from copy import deepcopy
 
+import numpy as np
 import pytest
 from astropy import units as u
 from astropy.table import Table
+from astropy.utils.exceptions import AstropyUserWarning
 from packaging.version import InvalidVersion
 
 from stellarphot import PhotometryData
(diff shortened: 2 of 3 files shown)


mwcraig (commented):
**Verdict: approve with minor suggestions.** The diagnosis was reproduced exactly: running the pre-fix migrator on the two-filter fixture with `mag_inst_ip` removed yields `mag_inst = [0, 0, 0]` (object dtype!) for the ip rows with no warning; the fixed version yields NaN in a proper float64 column plus an `AstropyUserWarning` naming the missing passband. The regression test is genuinely red pre-fix on both of its teeth (the `pytest.warns` and the `isnan` assertions). The full test file passes at the PR head (10 tests).

Findings:

1. **low** — `stellarphot/utils/version_migrator.py` (the new `warnings.warn`) — `stacklevel=2` attributes the warning to `VersionMigrator.migrate`'s internals, not the user's call site. The chain is user → `migrate` → `_migrate_v1_v2` → `warn`, so `stacklevel=3` would point at the user's code.
2. **low (worth a line in the PR description or changelog)** — Even in the all-columns-present case, `mag_inst` dtype changes from object (the old `np.zeros_like(Time-column)` artifact) to float64. This is a strict improvement and the existing tests still pass, but it is an unstated behavior change.
3. **low (pre-existing, follow-up candidate)** — The enclosing guard `mag_cols_have_filter_names = any(...)` means that if *every* `mag_inst_<band>` suffix fails to match a filter value, the whole block — including the new NaN/warning logic — is skipped and the band-suffixed columns pass through unrenamed with no `mag_inst` at all. That is the same "silent wrong migration" class this PR targets, just one level up. Out of scope here, but worth a follow-up issue.
4. **nit** — the comment above the new test duplicates the `two_filters` fixture explanation already given above the previous test.

*Verification: pre-fix behavior reproduced by running the old migrator blob directly against the fixture in a scratch worktree; fixed behavior and warning confirmed the same way; full `test_version_migrator.py` run at PR head.*

PR #614, opened 2026-07-07
Raise a clear error if transit eccentricity is un-fixed before fitting

mwcraig:
### The bug (#593)

Since the pytransit migration, `transit_model_with_trends` never forwards `eccentricity` to `self._transit_model.evaluate(...)` — pytransit always uses a circular orbit (`e=0`, `w=0`). The model sets `eccentricity.fixed = True` by default, but nothing stopped a user from setting `model.eccentricity.fixed = False`. The fit then ran anyway: the LM fitter saw a zero Jacobian column for that parameter, reported a meaningless "best-fit" eccentricity, and the extra free parameter inflated `n_fit_parameters` and the BIC. (Verified: un-fixing and setting eccentricity to 0.9 changes the model output by exactly 0.0.)

### The fix

Fail loudly instead of fitting a no-op:

- `fit()` now raises a `ValueError` with an explanatory message if `eccentricity.fixed` is `False`.
- `fit()` emits an `AstropyUserWarning` if the (fixed) eccentricity has a nonzero value, since that value is silently ignored by the pytransit model.
- The limitation is now documented in the `TransitModelFit` class docstring and the `fit()` docstring, not just buried in inline comments.

Erroring on un-fix was chosen over implementing `e`/`w` forwarding to pytransit: actually fitting eccentricity is a new feature (it also requires an argument-of-periastron parameter and sensible bounds/defaults) and is tracked by the broader transit-stack proposal. This PR just stops the silent no-op fit.

### Tests

- `test_fit_raises_if_eccentricity_not_fixed`
- `test_fit_warns_if_fixed_eccentricity_is_nonzero`
- `test_fit_with_default_eccentricity_neither_raises_nor_warns`

Written first (TDD, confirmed failing before the fix); the whole transit-fitting test file passes (19 tests).

Fixes #593

### DIFF
--- a/stellarphot/transit_fitting/core.py
+++ b/stellarphot/transit_fitting/core.py
@@ -148,6 +148,13 @@ class TransitModelFit:
     <https://pytransit.readthedocs.io/en/latest/notebooks/models/roadrunner/roadrunner_model_example_1.html>`_
     using a quadratic limb-darkening law and a circular orbit.
 
+    Because the orbit is always circular, the model's ``eccentricity``
+    parameter has no effect on the computed light curve. It is kept in the
+    model for API stability, but it must remain fixed (the default) and its
+    value is ignored. Calling ``fit`` with ``eccentricity.fixed`` set to
+    ``False`` raises a ``ValueError``, and a fixed nonzero eccentricity
+    triggers a warning.
+
     Attributes
     ----------
 
@@ -335,11 +342,12 @@ class TransitModelFit:
             width_trend=0.0,
             spp_trend=0.0,
         ):
-            # pytransit uses inclination in radians. The orbit is circular
-            # (eccentricity is always fixed at 0 in this model), so ``evaluate``
-            # is left with its default ``e=0``; the ``eccentricity`` argument is
-            # retained in the signature for API stability but is unused, matching
-            # the previous behavior where batman's ``ecc`` was fixed at 0.
+            # pytransit uses inclination in radians. The orbit is circular:
+            # ``evaluate`` is left with its default ``e=0``/``w=0`` and the
+            # ``eccentricity`` argument is NOT forwarded, matching the previous
+            # behavior where batman's ``ecc`` was fixed at 0. The argument is
+            # retained in the signature only for API stability; ``fit()``
+            # raises if a user un-fixes it and warns if it is nonzero.
             flux = self._transit_model.evaluate(
                 k=rp,
                 ldc=[limb_u1, limb_u2],
@@ -357,7 +365,9 @@ class TransitModelFit:
 
         self._model = ModelClass()
 
-        # Set up some defaults for what is fixed
+        # Set up some defaults for what is fixed. The eccentricity must stay
+        # fixed: it is not forwarded to pytransit (the orbit is always
+        # circular) and fit() raises if it has been un-fixed.
         self._model.period.fixed = True
         self._model.eccentricity.fixed = True
         self._model.limb_u1.fixed = True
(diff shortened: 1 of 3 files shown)


stellarphot/transit_fitting/core.py
```diff
@@ -148,6 +148,13 @@ class TransitModelFit:
     <https://pytransit.readthedocs.io/en/latest/notebooks/models/roadrunner/roadrunner_model_example_1.html>`_
     using a quadratic limb-darkening law and a circular orbit.
 
+    Because the orbit is always circular, the model's ``eccentricity``
```
mwcraig: This goes in the notes section

mwcraig: Agreed — this is caveat material, not summary material, and the repo already uses numpydoc ``Notes`` sections for exactly this kind of thing (e.g. ``apass_dr9``/``refcat2`` in ``catalogs.py``, ``PhotometryData``/``CatalogData`` in ``core.py``). The class docstring currently has no ``Notes`` section, so the restructure would be:

```python
class TransitModelFit:
    """
    Transit model fits to observed light curves.

    The underlying transit light curve is computed with pytransit's
    `RoadRunnerModel
    <https://pytransit.readthedocs.io/en/latest/notebooks/models/roadrunner/roadrunner_model_example_1.html>`_
    using a quadratic limb-darkening law and a circular orbit.

    Attributes
    ----------
    ...
    width : array-like
        Width of the star in pixels at each time. Must be set before fitting.

    Notes
    -----
    Because the orbit is always circular, the model's ``eccentricity``
    parameter has no effect on the computed light curve. It is kept in the
    model for API stability, but it must remain fixed (the default) and its
    value is ignored. Calling ``fit`` with ``eccentricity.fixed`` set to
    ``False`` raises a ``ValueError``, and a fixed nonzero eccentricity
    triggers a warning.
    """
```

The one line I'd keep up top is the existing "circular orbit" mention in the extended summary, so the constraint is still visible at a glance. Happy to push this to the branch if that looks right.

mwcraig (commented):
**Verdict: approve with two suggestions.** The diagnosis is independently confirmed: pytransit's `RoadRunnerModel.evaluate` has `e=0.0, w=0.0` defaults (checked via `inspect.signature` on the installed pytransit 2.8.1) and the code never passes them, so eccentricity is a no-op. Red/green verified by reverting `core.py` to main under the new test file: `test_fit_raises_if_eccentricity_not_fixed` and `test_fit_warns_if_fixed_eccentricity_is_nonzero` genuinely fail without the fix; all 19 tests pass at the head. Also checked that the new error/warning cannot fire spuriously today: both TESS notebooks, `TOI`, and `TransitModelOptions` were scanned and none of them touches `model.eccentricity`.

Findings:

1. **low (suggest rewording)** — `stellarphot/transit_fitting/core.py:345–350` — the rewritten comment still says the non-forwarding matches "the previous behavior where batman's ``ecc`` was fixed at 0". That's only true for the *default* case. Verified against the pre-migration code (`9a49fe9^`, line 374): batman-era `transit_model_with_trends` did `self._batman_params.ecc = eccentricity` on every call, so un-fixing eccentricity back then would have actually fit an eccentric orbit. The pytransit migration silently dropped a real capability — which is exactly what #593 reports. Since this PR rewrites that very paragraph, it's the right moment to state that plainly, e.g. "unlike the batman-era code, which did forward ``ecc``, pytransit's ``evaluate`` is never given ``e=``/``w=``, so the parameter is dropped regardless of ``fixed``."
2. **nit** — `core.py:518` — `self.model.eccentricity.value != 0` is an exact float comparison. Nothing in the current codebase computes an eccentricity, so no practical risk today, but a tolerance (`not np.isclose(..., 0)`) would be more robust against a future caller that sets it from a catalog value.

*Verification: pytransit evaluate signature inspected at runtime; red/green by selective revert in a scratch worktree; batman-era forwarding confirmed directly against the pre-migration commit; notebook/TOI/options call sites scanned for eccentricity usage.*

PR #617, opened 2026-07-07
Fix operator precedence that disabled distance cut in transform_to_catalog

mwcraig:
### The bug

In `transform_to_catalog` the constraint on which stars enter the per-image fit was written as

```python
good_cat = ~(
    getattr(color, "mask", False)
    | getattr(cat_mag, "mask", False) & (d2d.arcsecond < 1)
)
```

Since `&` binds tighter than `|`, this parses as `~(color_mask | (cat_mag_mask & dist_cut))`. For ordinary unmasked columns both `getattr(..., "mask", False)` calls return `False`, the distance term collapses, and `good_cat` is `True` for every star -- the cross-match distance cut never applied. Stars whose nearest catalog match was arbitrarily far away were included in the `curve_fit` for the transform coefficients, dragging the fit away from the correct values.

### The fix

Parenthesize as intended, so a star is used only if its color and catalog magnitude are unmasked **and** its match is closer than 1 arcsec:

```python
good_cat = ~(
    np.asarray(getattr(color, "mask", False))
    | np.asarray(getattr(cat_mag, "mask", False))
) & (d2d.arcsecond < 1)
```

The `np.asarray` wrapping is needed because with the corrected grouping, unmasked columns would otherwise hit `~False` on a plain Python bool, which is deprecated in Python 3.12+.

### Test

Adds `test_transform_to_catalog_excludes_distant_matches`, which builds a synthetic catalog (monkeypatched in place of the network fetch) where 20 stars lie exactly on the transform relation and one observed star's nearest match is a degree away, with a magnitude offset small enough to survive the median outlier cut. Before the fix the bad match pulls the fitted zero point off by about 0.08 mag; after the fix the good stars are recovered exactly and the distant star's calibrated magnitude is NaN.

Fixes #588

### DIFF
--- a/stellarphot/utils/tests/test_magnitude_transforms.py
+++ b/stellarphot/utils/tests/test_magnitude_transforms.py
@@ -351,7 +351,7 @@ class FakeCatalogTable(Table):
         return self
 
 
-def test_transform_to_catalog_excludes_distant_matches(monkeypatch):
+def test_transform_to_catalog_excludes_distant_matches(mocker):
     # A star whose nearest catalog match is far away (much more than
     # 1 arcsec) should be excluded from the fit for the transform
     # coefficients. See issue #588 -- an operator precedence error
@@ -402,9 +402,7 @@ def test_transform_to_catalog_excludes_distant_matches(monkeypatch):
 
     # Replace the catalog fetch with our synthetic catalog so no
     # network access is needed.
-    monkeypatch.setattr(
-        magnitude_transforms, "apass_dr9", lambda *_args, **_kwargs: catalog
-    )
+    mocker.patch.object(magnitude_transforms, "apass_dr9", return_value=catalog)
 
     result = transform_to_catalog(
         observed,
--- a/stellarphot/utils/tests/test_magnitude_transforms.py
+++ b/stellarphot/utils/tests/test_magnitude_transforms.py
@@ -8,10 +8,12 @@ from astropy.table import Column, Table
 from astropy.utils.data import get_pkg_data_filename
 from astropy.utils.exceptions import AstropyUserWarning
 
+from .. import magnitude_transforms
 from ..magnitude_transforms import (
     calculate_transform_coefficients,
     filter_transform,
     transform_magnitudes,
+    transform_to_catalog,
 )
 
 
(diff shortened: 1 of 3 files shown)


stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -337,3 +339,91 @@ def test_coordinate_all_mismatches():
...
+            "dec": np.append(dec.to_value("degree"), bad_dec.to_value("degree")),
+            "mag_inst": np.append(good_mags, bad_mag),
+            "mag_error": [0.01] * (n_good + 1),
+        }
+    )
+    observed = observed.group_by("file")
+
+    # Replace the catalog fetch with our synthetic catalog so no
+    # network access is needed.
+    monkeypatch.setattr(
```
mwcraig: Can this be replaced by something in pytest.mock? Also, did you check that this test fails before the fix was added?

mwcraig: **On pytest-mock:** Yes. There's no `pytest.mock` as such — the two options are stdlib `unittest.mock` or the `pytest-mock` plugin's `mocker` fixture. The latter is already available here (it's a dependency of `pytest-astropy` in the `test` extra) and is what the repo uses elsewhere, e.g. `mocker.patch.object(...)` in `test_settings_file.py` and the GUI tests. The swap is:

```python
def test_transform_to_catalog_excludes_distant_matches(mocker):
    ...
    mocker.patch.object(
        magnitude_transforms, "apass_dr9", return_value=catalog
    )
```

which also drops the `lambda *_args, **_kwargs` shim, since `mocker` handles swallowing the call arguments. Happy to push that change to the branch if you'd like it.

**On the test failing before the fix:** Yes, and I just re-verified it in a clean worktree. At the parent commit `aa7f49c` with only the new test file overlaid, the test fails — the distant match drags the fit to `z = 18.8416` instead of 20 (`a = -0.1235` instead of 0), and the good stars' calibrated magnitudes come out wrong by up to 0.077 mag, tripping the `assert_allclose` on the good-star magnitudes. Restoring the fixed `magnitude_transforms.py` makes it pass. One nuance: the first assertion (the distant star's calibrated mag is NaN) passes even *without* the fix, because the NaN comes from the separate per-star match cut when magnitudes are written out — the teeth of the test are in the good-star coefficient checks.

mwcraig (commented):
**Verdict: approve.** The precedence diagnosis is correct and the fix produces the intended semantics (mask cut AND distance cut). Red/green was re-verified independently: reverting just the fixed hunk under the new test file reproduces the ~0.077 mag zero-point bias on all 20 good stars; the full test file (25 tests) passes at the head. An AST/regex scan of every `.py` file at main found exactly one instance of the `~(A | B & C)` mixed-precedence pattern — the one this PR fixes — so no siblings were missed.

Findings (nits only):

1. **nit (documentation accuracy)** — the PR body's justification for the `np.asarray` wrapping ("deprecated in Python 3.12+") undersells the real hazard. Without it, `~False` evaluates to the *int* `-1`, and `-1 & bool_array` produces an **int64** 0/1 array — which, if ever used for fancy indexing, does integer indexing instead of boolean masking and silently selects wrong elements. Verified by micro-experiment (`~False & np.array([True, False, True])` → `array([1, 0, 1])`, dtype int64). The code as written is correct — `np.asarray` keeps the dtype bool in all scalar/array combinations (verified) — so this is only worth a slightly stronger inline comment, e.g. "np.asarray keeps this boolean; `~False` would be the int -1".
2. **nit (optional hardening)** — the new test only covers the both-unmasked case. Adding a variant with an actually masked `cat_mag`/`color` column would pin the array-mask path of the same expression against future refactors. Not required for this fix.

Also checked: `MaskedColumn.mask` is always a full-length ndarray (never a scalar sentinel), validating the `getattr(..., "mask", False)` fallback; the new test passes under `-W error::DeprecationWarning`; `ruff check` is clean on the changed files; and the PR diagnosis matches issue #588's reproduction exactly.

*Verification: red/green by selective revert in a scratch worktree; full test file run at head; numpy/astropy micro-experiments for the `~False` and mask-dtype behavior; codebase-wide pattern scan.*

PR #618, opened 2026-07-07
Fix silent bad output, NameError, and comp-star error consistency in AIJ relative flux code

mwcraig:
This PR fixes three issues in `stellarphot/differential_photometry/aij_rel_fluxes.py`: the two originally targeted bugs (#590, #597) plus the error/consistency cleanups from #605, which overlap the same lines and interact with the zero-comps guard.

### Bug 1: `calc_aij_relative_flux` silently produces garbage when no comparison stars match (#590)

The comparison total/error vectors are initialized to ones and only overwritten for times present in the grouped comparison fluxes. The existing guard (`len(set(comp_num_stars)) > 1`) does not fire when the set is *empty*, so if every comparison star fails the coordinate match (e.g. the comparison table has an RA offset) the function returned normally with `relative_flux == aperture_net_cnts` -- silently wrong output.

**Fix:** two `RuntimeError` guards, each with a message matching the condition it detects: one fires when no comparison star matches the photometry data positions at all (pointing at the coordinates), and one fires when one or more times have no surviving comparison stars (explaining that a comparison star that is missing, has NaN counts, or has a mismatched position at any one time is excluded at every time -- so this guard also covers the all-excluded case). `RuntimeError` matches the existing error convention in this function. The per-time guard compares the set of times with surviving comps against all times in the input rather than counting groups, which also avoids an astropy crash on `len(groups)` of an empty grouped table. (This is item 3 of #605, which suggested fixing it together with the items below.)

### Bug 2: `add_relative_flux_column` raises `NameError` when the table already has `bjd` (#597)

`flux_group` (and `bjds`) were assigned only inside the `if "bjd" not in flux_table.colnames:` block, but `flux_group["bjd"] = ...` and `flux_group.write(...)` ran unconditionally after it, so any photometry file that already contained a `bjd` column crashed with an unbound-variable error.

**Fix:** group the table by `file` unconditionally and move the BJD-column assignment inside the "bjd missing" branch.

### Bug 3: comparison-star error/consistency cleanups (#605)

**Item 1 -- inconsistent comp-star SNR.** For comparison-star rows, `relative_flux` self-excludes the star via `flux_offset`, but `rel_flux_error` used the *unadjusted* denominator and included the star's own error in the quadrature sum, so the flux and its error were computed against different ensembles.

**Fix:** the error for a comparison-star row now uses the same self-excluded ensemble as its relative flux -- the star's own counts are removed from the comparison total and its own error from the quadrature sum. Note that **this changes the reported errors (and SNR) of comparison-star rows.** AIJ-compatibility was considered as the alternative (documenting the old behavior as intentional), but since the flux already self-excludes, making the error follow suit is the internally consistent choice. The `"comparison counts"`/`"comparison error"` output columns still record the full-ensemble totals, AIJ style. The error propagation is also rewritten algebraically (same formula, no division by the star's own counts) so a star with zero counts gets a finite error instead of NaN.

**Item 2 -- `comp_num_stars` used `np.count_nonzero`.** A comparison star with *exactly zero* net counts -- a legitimate measured value, net counts can even be negative after sky subtraction -- was miscounted as a missing star by the per-image consistency check, raising a misleading "Different number of stars in comparison sets" error.

**Fix:** count the comparison stars present in each image (rows per group) instead of counting nonzero fluxes.

### Tests

All fixes were developed test-first (tests confirmed failing before each fix):

- `test_no_matching_comp_stars_raises_error` -- shifts all comparison star RAs by 1 degree and asserts `calc_aij_relative_flux` raises the "no comparison stars matched" error.
- `test_comp_stars_missing_at_one_time_raises_error` -- removes the comparison-star rows at one time (keeping the target row there and the comps at all other times) and asserts the "one or more times" error.
- `test_add_relative_flux_column` (parametrized over `bjd_already_present`) -- round-trips photometry data through `add_relative_flux_column` with and without a pre-existing `bjd` column (`verbose=True`), checking the output file has a finite `bjd` column and sane relative-flux columns.
- `test_comp_star_error_uses_self_excluded_ensemble` -- hand-computed self-excluded error for a comparison-star row, plus SNR consistency.
- `test_comp_star_with_zero_flux_is_counted` -- a comp star with exactly zero counts in one image is treated as a valid measurement: correct relative fluxes, finite errors, no spurious error.

The expected errors in the `_raw_photometry_table` fixture encoded the old inconsistent behavior for comparison-star rows; they are updated to the self-excluded ensemble to match the flux calculation.

Fixes #590
Fixes #597
Fixes #605

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -80,6 +80,12 @@ Bug Fixes
 + Fixed an operator-precedence bug in ``transform_to_catalog`` that disabled
   the cross-match distance cut, letting badly matched stars bias the fitted
   transform coefficients. [#617]
++ ``calc_aij_relative_flux`` now raises an error when none of the comparison
+  stars match the photometry data, or when one or more times have no valid
+  comparison stars, instead of silently returning relative fluxes equal to
+  the raw net counts. [#618]
++ ``add_relative_flux_column`` no longer raises a ``NameError`` when the input
+  photometry data already contains a ``bjd`` column. [#618]
 
 1.4.15 (2024-08-16)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -86,6 +86,17 @@ Bug Fixes
   the raw net counts. [#618]
 + ``add_relative_flux_column`` no longer raises a ``NameError`` when the input
   photometry data already contains a ``bjd`` column. [#618]
++ The relative flux error of a comparison star is now computed against the
+  same ensemble as its relative flux, with the star itself excluded from
+  both the comparison counts and the comparison error, making the reported
+  error and SNR consistent with the flux. Previously the error used the
+  full ensemble including the star itself. [#618]
++ The per-image comparison-star consistency check in
+  ``calc_aij_relative_flux`` now counts the comparison stars present in
+  each image instead of counting nonzero fluxes, so a comparison star with
+  exactly zero net counts (a legitimate measured value) no longer triggers
+  a misleading "Different number of stars in comparison sets" error, and a
+  star with zero counts now gets a finite relative flux error. [#618]
 
 1.4.15 (2024-08-16)
 -------------------
--- a/stellarphot/differential_photometry/aij_rel_fluxes.py
+++ b/stellarphot/differential_photometry/aij_rel_fluxes.py
@@ -170,7 +170,11 @@ def calc_aij_relative_flux(
         )
 
     comp_totals = comp_fluxes.groups.aggregate(np.sum)[counts_column_name]
-    comp_num_stars = comp_fluxes.groups.aggregate(np.count_nonzero)[counts_column_name]
+    # Count the comparison stars in each image by counting the rows in each
+    # group. Counting nonzero fluxes would be wrong here -- exactly zero net
+    # counts is a legitimate measured value (net counts can even be negative
+    # after sky subtraction) and must not be mistaken for a missing star.
+    comp_num_stars = np.diff(comp_fluxes.groups.indices)
     comp_errors = comp_fluxes.groups.aggregate(add_in_quadrature)[error_column_name]
 
     comp_total_vector = np.ones_like(star_data[counts_column_name])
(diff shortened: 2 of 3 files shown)


mwcraig (commented):
**Verdict: approve.** All three bundled fixes check out individually. The base-code `UnboundLocalError` is real (on main, `bjds`/`flux_group` are assigned only inside the `if "bjd" not in …` block but used unconditionally afterward — confirmed by reading the base file at `aa7f49c`, lines 256–267). The comp-star error change was verified numerically on the fixture: the target star's flux and error are bit-for-bit unchanged, while comparison-star errors switch to the self-excluded-ensemble formula, which is the standard error propagation for the quantity actually being reported. Red/green is thorough: with the base source swapped in under the PR's test file, 13 of 17 tests fail, each for its intended reason — the two new guard tests "DID NOT RAISE", the zero-flux test hits the old misleading "Different number of stars" error, the ensemble-error test fails its assertion, and `test_add_relative_flux_column[True]` hits the `UnboundLocalError`. At the PR head all 17 pass.

Findings (nits only, neither blocking):

1. **nit (pre-existing, optional follow-up)** — `aij_rel_fluxes.py` (~lines 208–238): a degenerate ensemble with exactly one comparison star still yields inf/NaN — the lone comp excludes itself, so `comp_total_used == 0` and the division blows up. The new guards don't catch "one comp star total" (they catch zero and per-time dropouts). This was equally broken on main (the old denominator was also 0), so it's not a regression; a `min comp count >= 2` guard or a docstring note would close it.
2. **nit (behavior note)** — `add_relative_flux_column` now groups by `file` unconditionally, so when the input already has a `bjd` column the output file's row order can differ from the input order. That matches what the no-bjd path always did, and nothing downstream depends on row order, but it's an unstated change worth a line in the PR description.

*Verification: 17 tests run at head; base-source swap run to confirm each new test is red for its own reason; numerical before/after comparison of target vs. comp errors on the fixture; base NameError path confirmed directly against `aa7f49c`.*

CHANGES.rst
```diff
@@ -57,6 +57,23 @@ Bug Fixes
...
   fields. [#586]
++ ``calc_aij_relative_flux`` now raises an error when none of the comparison
+  stars match the photometry data, or when one or more times have no valid
+  comparison stars, instead of silently returning relative fluxes equal to
+  the raw net counts. [#618]
++ ``add_relative_flux_column`` no longer raises a ``NameError`` when the input
+  photometry data already contains a ``bjd`` column. [#618]
++ The relative flux error of a comparison star is now computed against the
+  same ensemble as its relative flux, with the star itself excluded from
+  both the comparison counts and the comparison error, making the reported
```
mwcraig, line 91: Compare this behavior with the latest astro image j code (it is on github) to see what AIJ does and report back.

stellarphot/differential_photometry/aij_rel_fluxes.py
```diff
@@ -247,24 +289,25 @@ def add_relative_flux_column(
     if verbose:
         print("Adding AIJ-style relative flux columns to photomotry data")
     flux_table = calc_aij_relative_flux(photometry_data, only_comp_stars)
```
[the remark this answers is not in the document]
mwcraig: Confirmed -- "photomotry" should be "photometry". (The typo actually pre-dates this PR -- it's at line 248 on main -- but this function is already being touched here, so I'll fix it on this branch along with the other review fixes.)

CHANGES.rst
mwcraig: Report back: checked the current AstroImageJ source (`AstroImageJ/astroimagej` on GitHub, `master` branch). **AIJ does exactly what this changelog entry describes** -- the comparison star itself is excluded from both the ensemble flux *and* the ensemble error.

In `Astronomy_/src/main/java/Astronomy/MultiAperture_.java` (lines ~5435-5445), the per-aperture ensemble totals are accumulated with the star's own aperture skipped:

```java
if (isRefStar[ap]) {
    for (int i = 0; i < nApertures; i++) {
        if (i != ap) {            // the star itself is excluded
            tot[i] += source;
            totVar[i] += srcVar[ap];
        }
    }
}
```

So for comp star Cn, `tot[ap]` is the sum of the *other* comps' fluxes and `totVar[ap]` the sum of their variances. Then (lines ~5539-5558):

- `rel_flux_Cn = src / tot`
- `rel_flux_err_Cn = ratio * sqrt(srcVar/src² + totVar/tot²)`
- `rel_flux_SNR_Cn = 1 / sqrt(srcVar/src² + totVar/tot²)`

Flux, error, and SNR all use the same self-excluded (N−1)-star ensemble. The file-header comment (lines ~141-152, "equation 2") documents the same: *"tot = total source counts less sky background from the 2nd through Nth apertures"*.

The formula in this PR -- `sqrt((σ_src/tot_used)² + (src·σ_tot_used/tot_used²)²)` -- is algebraically the same ratio propagation, just factored to avoid dividing by `src`, so a comp star with exactly zero net counts gets a finite error. (One small difference: AIJ short-circuits and writes `rel_flux_err = 0.0` when `src == 0` or `tot == 0`; a finite propagated error seems strictly better there.)

So: before this PR stellarphot computed the comp-star error against the *full* ensemble including the star itself, which AIJ never does; after this PR the behavior matches AIJ.

PR #619, opened 2026-07-07
Fix NaN/saturation handling and unclipped sky stats in aperture photometry

mwcraig:
This PR fixes two related NaN/saturation-handling bugs in `single_image_photometry`, both stemming from pixels above the camera's `max_data_value` being set to NaN with no downstream handling, plus a bug in the sky statistics reported when background outlier rejection is turned off. It extends the earlier NaN-handling work in #161.

Fixes #591
Fixes #592
Fixes #602

## Bug 1: saturated (NaN) pixels silently poison aperture sums and are never flagged (#591)

Only the original `ccd_image.mask` was passed to `aperture_photometry`, so any source with a single saturated pixel in its aperture or annulus got an all-NaN sum, and the downstream `aperture_net_cnts < 0` check is `False` for NaN, so the source was never flagged.

**Fix:**
- The non-finite pixels (the saturated pixels marked as NaN, plus any NaN already in the input) are now added to the image mask, so they are excluded from the aperture sums and the sky statistics.
- Because the masked sum would then silently *underestimate* the flux of a saturated source, sources whose aperture contains any saturated pixel are explicitly flagged in a new boolean `saturated` column, a warning is logged, and their `aperture_net_cnts` is set to NaN following the existing convention for bad-FWHM/bad-count sources.
- The `saturated` column is documented in the `single_image_photometry` docstring, and the "until #161 is fixed" comment in the test suite has been updated.

## Bug 2: NaN centroid escapes the shift-tolerance fallback, crashing an entire multi-image run (#592)

In the `use_coordinates="sky"` path, `centroid_sources` was called without any mask, so a fully saturated source yielded a NaN centroid. `center_diff > shift_tolerance` is `False` for NaN, so the WCS fallback never fired, and the NaN position reached `CircularAperture`, raising a `ValueError` that aborted the whole multi-image run.

**Fix:**
- The (NaN-updated) image mask is now applied during centroiding. It is applied by setting masked pixels to NaN rather than by passing `mask=` to `centroid_sources`, because photutils 2.2 raises a `ValueError` for a completely masked cutout when given an explicit mask, while non-finite data is auto-masked and yields a NaN centroid — which is exactly what the fallback logic needs.
- A non-finite `center_diff` is now treated the same as "too much shift", so a source whose centroid cannot be computed falls back to its WCS-derived position instead of crashing.
- `compute_fwhm` returns NaN for a completely masked cutout instead of letting `fit_fwhm` raise.

Broader per-image try/except error handling in `multi_image_photometry` was deliberately left out to keep this change minimal; this PR only fixes the specific NaN escape paths.

## Bug 3: sky median/std computed from bounding-box pixels when `reject_background_outliers=False` (#602)

With `reject_background_outliers=False`, `sky_per_pix_med` and `sky_per_pix_std` were computed with `np.median`/`np.std` over `mask.cutout(ccd_image)` — the raw rectangular *bounding box* of the annulus, which includes the star's core (inside the inner radius) and the corner pixels outside the annulus. For the fake test image, the reported std was 32–61 counts instead of the true background deviation of 1 count. Only these two diagnostic columns were affected; `sky_per_pix_avg` (the value actually subtracted) was already computed correctly from `annulus_sum / annulus_area`.

**Fix:** use `ApertureStats` (as the `reject_background_outliers=True` branch already does, just without the sigma clip) so the median and standard deviation describe the annulus pixels only. `ApertureStats` also respects the image mask, which now includes the saturated pixels from Bug 1.

## Tests

Written test-first (each confirmed failing before its fix):
- `test_aperture_photometry_flags_saturated_source`: a source with saturated pixels in its aperture is flagged and gets NaN net counts, while the other sources' photometry is identical to that of an image with no saturated pixels.
- `test_saturated_source_recentroiding_falls_back_to_wcs`: a completely saturated source in the sky-coordinates + recentroiding path no longer crashes the run; it falls back to the WCS-derived position and is flagged.
- `test_sky_stats_annulus_only_without_outlier_rejection`: with `reject_background_outliers=False`, the reported sky median/std reflect the flat background in the annulus rather than being inflated by the stellar core.

Full `test_photometry.py` (54 tests) and `test_detection.py`/`test_profiles.py` pass locally.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -97,6 +97,16 @@ Bug Fixes
   exactly zero net counts (a legitimate measured value) no longer triggers
   a misleading "Different number of stars in comparison sets" error, and a
   star with zero counts now gets a finite relative flux error. [#618]
++ Saturated pixels (those above the camera's maximum data value) no longer
+  silently poison the aperture sums in ``single_image_photometry``. They are
+  now masked in the photometry, and sources with saturated pixels in their
+  aperture are flagged in a new boolean ``saturated`` column and have their
+  ``aperture_net_cnts`` set to NaN. [#591]
++ A source whose centroid cannot be computed when ``use_coordinates="sky"``
+  (for example, because the source is completely saturated) no longer crashes
+  ``single_image_photometry`` -- and with it an entire multi-image run -- with
+  a ``ValueError``. The source now falls back to its WCS-derived position,
+  like sources whose centroid shifts by more than ``shift_tolerance``. [#592]
 
 1.4.15 (2024-08-16)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -107,6 +107,10 @@ Bug Fixes
   ``single_image_photometry`` -- and with it an entire multi-image run -- with
   a ``ValueError``. The source now falls back to its WCS-derived position,
   like sources whose centroid shifts by more than ``shift_tolerance``. [#592]
++ With ``reject_background_outliers=False``, the reported ``sky_per_pix_med``
+  and ``sky_per_pix_std`` are now computed from the pixels in the annulus
+  instead of from the rectangular bounding box of the annulus, which included
+  the core of the star and the corners outside the annulus. [#619]
 
 1.4.15 (2024-08-16)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -110,7 +110,7 @@ Bug Fixes
 + With ``reject_background_outliers=False``, the reported ``sky_per_pix_med``
   and ``sky_per_pix_std`` are now computed from the pixels in the annulus
   instead of from the rectangular bounding box of the annulus, which included
-  the core of the star and the corners outside the annulus. [#619]
+  the core of the star and the corners outside the annulus. [#602]
 
 1.4.15 (2024-08-16)
 -------------------
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -652,15 +652,17 @@ def single_image_photometry(
             "without clipping (set reject_background_outliers=True "
             "to perform clipping)."
         )
-        med_pp = []
-        std_pp = []
-        for mask in annuli.to_mask():
-            annulus_data = mask.cutout(ccd_image)
-            med_pp.append(np.median(annulus_data))
-            std_pp.append(np.std(annulus_data))
+        # Compute the median and standard deviation from the pixels in the
+        # annulus only. These used to be computed from the rectangular
+        # bounding box of the annulus, which includes the core of the star
+        # (inside the inner radius) and the corners outside the annulus,
+        # wildly inflating the reported standard deviation. See #602.
+        # ApertureStats also respects the image mask, which includes the
+        # saturated pixels marked as NaN above (#591).
+        ap_stats = ApertureStats(ccd_image, annuli, sum_method="center")
         photom["sky_per_pix_avg"] = photom["annulus_sum"] / photom["annulus_area"]
-        photom["sky_per_pix_med"] = np.array(med_pp) * ccd_image.unit / u.pixel
-        photom["sky_per_pix_std"] = np.array(std_pp) * ccd_image.unit / u.pixel
+        photom["sky_per_pix_med"] = ap_stats.median / u.pixel
+        photom["sky_per_pix_std"] = ap_stats.std / u.pixel
 
     # Compute counts using clipped stats on sky per pixel
     photom["aperture_net_cnts"] = photom["aperture_sum"].value - (
--- a/stellarphot/photometry/photometry.py
+++ b/stellarphot/photometry/photometry.py
@@ -236,7 +236,11 @@ def single_image_photometry(
         locations were too close to the edge of the image or to each other for
         successful aperture photometry.  If pixel (x/y) positions were used for
         the photometry, but a valid WCS header was not available for `ccd_image`,
-        the output 'ra', 'dec', and 'bjd' columns will have np.nan values
+        the output 'ra', 'dec', and 'bjd' columns will have np.nan values.
+        The output also contains a boolean 'saturated' column that is ``True``
+        for sources whose aperture contains one or more saturated (or
+        otherwise non-finite) pixels; those sources have their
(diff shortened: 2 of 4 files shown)


stellarphot/photometry/photometry.py
```diff
@@ -594,15 +652,17 @@ def single_image_photometry(
...
-            med_pp.append(np.median(annulus_data))
-            std_pp.append(np.std(annulus_data))
+        # Compute the median and standard deviation from the pixels in the
+        # annulus only. These used to be computed from the rectangular
+        # bounding box of the annulus, which includes the core of the star
+        # (inside the inner radius) and the corners outside the annulus,
+        # wildly inflating the reported standard deviation. See #602.
+        # ApertureStats also respects the image mask, which includes the
+        # saturated pixels marked as NaN above (#591).
+        ap_stats = ApertureStats(ccd_image, annuli, sum_method="center")
```
mwcraig, line 662: This is in the no-outlier-rejection branch, right?

mwcraig (commented):
**Verdict: approve, with one finding worth fixing (either here or as an immediate follow-up).** All three diagnoses are correct on main and all three fixes work: the three regression tests genuinely fail against main's `photometry.py`/`source_detection.py` (verified by selective revert in a scratch worktree), and the full photometry + detection suites pass at the head (100 passed, 16 skipped). The photutils API claims in the PR description were verified by direct execution against photutils 2.2.0: `centroid_sources` with `mask=` really does raise `ValueError` on a fully masked cutout, NaN data auto-masks to a NaN centroid, and the warning is exactly the `AstropyUserWarning` the code filters. The `compute_fwhm` all-masked guard is load-bearing, not defensive fluff — reverting only `source_detection.py` makes the saturation test die with `ValueError: zero-size array` in the FWHM fit.

Findings:

1. **medium** — `stellarphot/photometry/photometry.py:663` — in the `reject_background_outliers=False` branch, `sky_per_pix_avg = annulus_sum / annulus_area` divides a mask-excluding sum by the full geometric area. A source with saturated/NaN pixels in its *annulus* only (not its aperture) therefore gets a silently low sky and inflated net counts, and is not flagged — `saturated` only checks the aperture. Confirmed empirically: saturating 3 annulus-only pixels biased `sky_per_pix_avg` by −0.0103 adu/pix and net counts by +4.6 adu with `saturated=False`, while the clipped branch showed ~zero bias on the same image. The bias scales as `aperture_area × sky × n_bad/annulus_area`, so bright sky plus a saturated neighbor could reach thousands of counts. This is still strictly better than main (which NaN-poisons the same case, unflagged), hence not blocking — but the one-line fix is to use `ap_stats.mean / u.pixel` here, consistent with how the clipped branch already handles the mask (note it slightly changes values vs. main for exact-method photometry). Extending the `saturated` flag to annulus contamination would also work.
2. **low** — `photometry.py:691` — `bad_fwhm = (fwhm_x < 1) | (fwhm_y < 1)` is `False` for the NaN that the new `compute_fwhm` guard returns — the same NaN-comparison escape this PR fixes elsewhere. It's harmless today only because a fully masked cutout implies the aperture center is masked, so the saturation flag catches those sources anyway; the correctness is coupled to that coincidence. Suggest ORing in `np.isnan(fwhm_x) | np.isnan(fwhm_y)`.
3. **nit** — `CHANGES.rst` — the sky-stats bullet cites `[#619]` while the other two new bullets cite their issue numbers (`[#591]`, `[#592]`); the matching issue is #602.
4. **nit** — the new `saturated` column is documented in the `single_image_photometry` docstring but not in `PhotometryData`'s column documentation; it survives via the extra-columns-appended path in `_validate_columns`. Worth a mention in the `PhotometryData` docs.
5. **nit** — `photometry.py:662` — `sum_method="center"` on this `ApertureStats` call is inert (only `median`/`std` are consumed); evidently copied from `clipped_sky_per_pix_stats`. Could be dropped for clarity.

One note on the tests: the saturation test fails on main first via `KeyError: 'saturated'` (the column doesn't exist there), but its behavioral assertions have real teeth — main demonstrably produces unflagged NaN sums since `net_cnts < 0` is `False` for NaN. The sky-std assertion is a genuine guard (main: 31.9–60.6 vs. expected 1.0 at rtol=0.2); the sky-median assertion also passes on main and is decorative.

*Verification: red/green by selective reverts (both files, then `source_detection.py` alone) in a scratch worktree; full `test_photometry.py` + `test_detection.py` run at head; quantitative annulus-contamination probe across all four clean/dirty × clip/no-clip combinations; photutils 2.2.0 behavior exercised directly. Findings 1 and 2 were re-confirmed against the code at the PR head by the orchestrating session.*

CHANGES.rst
```diff
@@ -57,6 +57,20 @@ Bug Fixes
...
+  ``aperture_net_cnts`` set to NaN. [#591]
++ A source whose centroid cannot be computed when ``use_coordinates="sky"``
+  (for example, because the source is completely saturated) no longer crashes
+  ``single_image_photometry`` -- and with it an entire multi-image run -- with
+  a ``ValueError``. The source now falls back to its WCS-derived position,
+  like sources whose centroid shifts by more than ``shift_tolerance``. [#592]
++ With ``reject_background_outliers=False``, the reported ``sky_per_pix_med``
+  and ``sky_per_pix_std`` are now computed from the pixels in the annulus
+  instead of from the rectangular bounding box of the annulus, which included
+  the core of the star and the corners outside the annulus. [#619]
```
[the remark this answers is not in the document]
mwcraig: Agreed. The two neighboring entries added in this same PR already cite the issues being fixed ([#591], [#592]), and per the decision on the identical Copilot comment in #620, the changelog should cite the issue -- so this should end with [#602]. Will fix on the branch.

stellarphot/photometry/photometry.py
mwcraig: Yes -- this line is inside the `else:` of `if photometry_options.reject_background_outliers:` (the branch that logs the "SUGGESTION: ... set reject_background_outliers=True" warning just above). The clipped path is untouched by this PR and still computes its stats via `clipped_sky_per_pix_stats`; only the no-rejection path changed, from statistics over the annulus's rectangular bounding-box cutout to annulus-only statistics via `ApertureStats`.

PR #624, opened 2026-07-07
Get TOI info from ExoFOP's single-target endpoint instead of the full table

mwcraig:
Fixes #623

`TOI.from_tic_id` downloaded the entire ExoFOP TOI table (3.7 MB that the server takes most of a minute to generate) to extract a single row, which intermittently timed out on CI and failed five remote-data tests.

### Changes

- **`TOI.from_tic_id` now uses ExoFOP's single-target endpoint** (`target.php?id=<tic>&json`, ~39 KB / a few seconds). It selects the TESS-project (`prov == "toi"`) entry from the provenance-structured `planet_parameters` list — keeping the old "exactly one entry, else `RuntimeError`" behavior — and reads the TESS magnitude from the `magnitudes` block. ExoFOP serves numbers as strings (including leading-dot decimals and empty strings for missing values), so a small `_exofop_float` helper does the conversion and raises a clear `RuntimeError` when a value is unusable.
- **Coordinates now come from the same ExoFOP response** (`coordinates` block), so the separate MAST `get_tic_info` query is gone from `from_tic_id` — one fewer flaky service in the chain. For the test TIC the ExoFOP and MAST positions agree to <0.1″. (Note: TOI JSON files written with the old code could differ from new ones at the sub-arcsecond level.)
- **Offline tests with a captured server response** (per the discussion on the issue): a real `target.php?id=236158940&json` payload is checked in as test data, and a new `TestTOIFromTicIdOffline` class covers parsing of the captured payload, provenance selection (user-supplied CTOI parameters are ignored), wrong-entry-count and missing-TESS-magnitude errors, and empty-value handling. The tests also pin that `from_tic_id` never calls `download_file` or `get_tic_info`.
- **`SERVER_DOWN_ERRORS` broadened and shared**: moved to `stellarphot/conftest.py` and extended with `requests.Timeout`, `requests.HTTPError`, `urllib.error.URLError` (astropy/astroquery paths), and `TimeoutError`. The previously unguarded `TestTOI::test_from_tic_id` and GUI `test_click_yes_confirm` are now wrapped so genuine service outages xfail instead of failing, and `test_target_file`'s two except branches collapse into one.
- Changelog entry; codespell skip for the captured payload (its real field names, e.g. `fobid`, read as typos).

### Testing

- TDD: the new offline tests were written first and observed failing against the old code (deterministically, with no network), then pass after the rewrite.
- Live `--remote-data=any` run of `stellarphot/io/tests/test_tess.py`: all 21 tests pass; `test_from_tic_id` takes ~3 s instead of ~47 s+ on a cold cache.
- Full offline suite and `pre-commit` pass locally (the GUI test module can't be collected in my local env due to an unrelated astrowidgets/astro-image-display-api version mismatch; CI will exercise it).


stellarphot/io/tests/test_tess.py
```diff
@@ -164,6 +159,185 @@ def fake_get(*args, **kwargs):  # noqa: ARG001
...
+
+        def fail_get_tic_info(*args, **kwargs):  # noqa: ARG001
+            raise AssertionError("from_tic_id must not query MAST (#623)")
+
+        def fail_download(*args, **kwargs):  # noqa: ARG001
+            raise AssertionError(
+                "from_tic_id must not download the full TOI table (#623)"
+            )
+
+        monkeypatch.setattr("stellarphot.io.tess.requests.get", fake_get)
```
mwcraig: Can these be replaced with pytest-mock?

stellarphot/io/tests/test_tess.py
```diff
@@ -164,6 +159,185 @@ def fake_get(*args, **kwargs):  # noqa: ARG001
...
+        assert toi.tess_mag_error == pytest.approx(0.0145)
+
+        # Offline version of the round-trip regression test for #427
+        new_toi = TOI.model_validate_json(toi.model_dump_json())
+        assert toi.coord.separation(new_toi.coord).arcsecond < 0.01
+
+    def test_from_tic_id_only_uses_toi_provenance(
+        self, monkeypatch, tess_tic_expected_values
+    ):
+        # A user-supplied CTOI with a different period follows the TOI
```
mwcraig: Add more explanation to this test.

stellarphot/io/tests/test_tess.py
```diff
@@ -164,6 +159,185 @@ def fake_get(*args, **kwargs):  # noqa: ARG001
...
+            # Only a user provenance -> no TOI entries
+            ([{"prov": "user", "prov_num": "1"}, TOI_PLANET_ENTRY], 0),
+            # Two planets in the TOI section
+            (
+                [{"prov": "toi", "prov_num": "1"}, TOI_PLANET_ENTRY, TOI_PLANET_ENTRY],
+                2,
+            ),
+        ],
+    )
+    def test_from_tic_id_wrong_number_of_tois_raises(
```
mwcraig, line 326: Add comments to this test (and the following ones) explaining what the prupose of the test is.

stellarphot/io/tess.py
```diff
@@ -39,6 +38,19 @@
 )
 
 
+def _exofop_float(value, description):
```
mwcraig, line 41: Is there really not some function in the standard lib or other packages that are already dependencies that will do this for us?

stellarphot/io/tess.py
```diff
@@ -326,35 +338,71 @@ class TOI(BaseModel):
...
-        toi_table = Table.read(toi_table, format="ascii.csv")
-        toi_table = toi_table[toi_table["TIC ID"] == tic_id]
-        if len(toi_table) != 1:  # pragma: no cover
-            raise RuntimeError(f"Found {len(toi_table)} rows in table, expected one.")
-        toi_table = toi_table[0]
+        # Use the single-target endpoint (a few tens of kB) instead of the full
+        # TOI table (several MB that takes the server most of a minute to
+        # generate), which frequently timed out. See #623. The trailing bare
+        # "json" (no "=value") is the form the ExoFOP server expects.
+        response = requests.get(f"{EXOFOP_TARGET_URL}?id={tic_id}&json", timeout=30)
```
mwcraig, line 349: Does this provide any cacheing?

stellarphot/io/tess.py
```diff
@@ -326,35 +338,71 @@ class TOI(BaseModel):
...
+        response.raise_for_status()
+        target_info = response.json()
+
```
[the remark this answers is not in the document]
mwcraig: Good catch on the intent, though the mechanics are a bit different than described. Verified against the live endpoint: for a nonexistent TIC ID, ExoFOP returns **HTTP 200 with a plain-text body** (`No TIC object found for "999999999999"`), so `raise_for_status()` passes and the crash happens at `response.json()` — a `JSONDecodeError` that never mentions the TIC ID — before any `KeyError` could be reached. A partial-but-valid JSON payload would indeed hit the `KeyError` path.

Either way the conclusion stands: the old code raised a clear `RuntimeError` for a missing target and the new code doesn't. Plan:

1. Wrap `response.json()` in `try/except`, re-raising as `RuntimeError(f"ExoFOP returned an unusable response for TIC {tic_id}: ...")`.
2. Check for `planet_parameters`, `magnitudes`, and `coordinates` before use, raising a descriptive `RuntimeError` naming the TIC ID if any is absent.

Will implement on this branch with matching offline tests.

stellarphot/io/tests/test_tess.py
```diff
@@ -164,6 +159,185 @@ def fake_get(*args, **kwargs):  # noqa: ARG001
...
+                "from_tic_id must not download the full TOI table (#623)"
+            )
+
+        monkeypatch.setattr("stellarphot.io.tess.requests.get", fake_get)
+        monkeypatch.setattr("stellarphot.io.tess.get_tic_info", fail_get_tic_info)
+        # raising=False because once #623 is fixed download_file is no longer
+        # imported in stellarphot.io.tess and this setattr becomes a no-op.
+        monkeypatch.setattr(
+            "stellarphot.io.tess.download_file", fail_download, raising=False
+        )
```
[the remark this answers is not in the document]
mwcraig: Partially agree. One nuance: the `raising=False` patch is not permanently dead — `monkeypatch.setattr(..., raising=False)` still *sets* the attribute, so if a module-level `from astropy.utils.data import download_file` were ever re-added to `tess.py`, this patch would rebind it at test time and the guard would fire again. The gap it can't cover is a *function-level* import inside `from_tic_id`, which would grab the real `download_file` at call time.

Patching `astropy.utils.data.download_file` closes that path, since a function-level import resolves the name at call time, after the patch is in place. It's cheap insurance — will add it alongside the existing patch (which stays as the module-level-import guard).

stellarphot/io/tests/test_tess.py
mwcraig: Yes. `pytest-mock` is already available — it ships as a dependency of `pytest-astropy` (which is in the `test` extra), and the repo already uses the `mocker` fixture in `settings/tests/test_settings_file.py` and the `gui` tests, so this would also make `test_tess.py` consistent with existing convention.

Concretely, `patch_network` becomes:

```python
def patch_network(self, mocker, payload):
    fake_get = mocker.patch(
        "stellarphot.io.tess.requests.get",
        return_value=FakeExoFOPResponse(payload),
    )
    mocker.patch(
        "stellarphot.io.tess.get_tic_info",
        side_effect=AssertionError("from_tic_id must not query MAST (#623)"),
    )
    ...
    return fake_get
```

with the hand-rolled `calls` dict replaced by `fake_get.call_args` in the assertions. Will refactor.

stellarphot/io/tests/test_tess.py
mwcraig: Will do. Planned expansion, for the record:

```python
def test_from_tic_id_only_uses_toi_provenance(...):
    # ExoFOP's "planet_parameters" is a flat list in which entries containing
    # a "prov" key act as section headers; the entries that follow belong to
    # that provenance (TESS-project TOI, user-submitted CTOI, ...) until the
    # next header. This test builds a payload containing both a TOI section
    # and a user section whose entry has a deliberately different period
    # (99.9 d), then checks that from_tic_id returns the TESS-project value
    # (2.677573 d) -- i.e. that user/CTOI parameters are never picked up.
```

stellarphot/io/tests/test_tess.py
mwcraig: Will do — each of the three error-path tests will get a comment stating what it guards:

- `test_from_tic_id_wrong_number_of_tois_raises`: the TOI section of the payload must contain exactly one planet; zero entries (e.g. only a user-submitted CTOI section) or two entries (multi-planet system) should raise a `RuntimeError` reporting the count, mirroring the old full-table code's "expected one row" check.
- `test_from_tic_id_missing_tess_mag_raises`: a payload whose `magnitudes` list has no `"band": "TESS"` entry should raise a `RuntimeError` naming the TIC ID rather than silently using a different band.
- `test_from_tic_id_empty_value_raises`: ExoFOP serves missing numbers as empty strings; conversion must fail with a `RuntimeError` that names the specific field (here "transit duration error") instead of a bare `ValueError` from `float("")`.

stellarphot/io/tess.py
mwcraig: The conversion itself is just the builtin `float()` — it already handles everything ExoFOP serves, including leading-dot strings like `".0018336"`. The helper exists solely to attach *which field* failed: `float("")` raises `ValueError: could not convert string to float: ''`, which is useless when ten different fields go through the conversion and ExoFOP routinely serves empty strings for missing values.

I couldn't find a convert-with-context function in the stdlib or in our existing dependencies. Pydantic can coerce str→float, but it can't take this over: most of these values are combined with units (`* u.hour`, `Time(...)`) *before* the model sees them, so validation would fire too late and the error would name a model field, not the ExoFOP key. So the wrapper stays, unless you'd rather inline `float()` and accept the anonymous `ValueError`.

stellarphot/io/tess.py
mwcraig: No — plain `requests.get` has no caching, and that's a deliberate change. The old code used `download_file(TOI_TABLE_URL, cache=True, ...)`, which cached the multi-MB TOI table in the astropy download cache **indefinitely**: every later call reused the stale table (stale epoch/period ephemerides included) until the user manually cleared the cache. Since TOI parameters get refined over time, that was arguably a bug.

The single-target response is tens of kB and returns in about a second, so fetching fresh on every call is cheap and always current — and `from_tic_id` is only called during per-target setup anyway. If we ever want caching, `requests-cache` with a short TTL would be the right shape (astropy's `cache=True` would just recreate the staleness problem), but I don't think this call site needs it.

PR #639, opened 2026-07-18
Added pixi support to pyproject.toml

aneugart:
Added support for pixi.

* Added necessary tables to pyproject.toml to support pixi.
* Added pixi rules to .gitignore
* Added .gitattributes with default settings from pixi init.

To run with pixi:
* `pixi run lab`


pyproject.toml
```diff
@@ -253,3 +253,21 @@ filterwarnings = [
     # regardless of category.
     'ignore:The TestRunner.*deprecated:'
 ]
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64"]
```
JuanCab: I suggest supporting macOS on intel and Apple Silicon, look up those platform 'codes' and add them.  Once this test is made, I will pull these files and test that it works on a macOS machine.

JuanCab (changes requested):
Make the suggested change and then we can support it.

pyproject.toml
```diff
@@ -253,3 +253,36 @@ filterwarnings = [
...
+all = { features = ["all"], solve-group = "default" }
+docs = { features = ["docs"], solve-group = "default" }
+exoplanet = { features = ["exoplanet"], solve-group = "default" }
+gui = { features = ["gui"], solve-group = "default" }
+test = { features = ["test"], solve-group = "default" }
+test-py311 = { features = ["test", "py311"]}
+test-py312 = { features = ["test", "py312"]}
+test-py313 = { features = ["test", "py313"]}
+
+[tool.pixi.feature.all.tasks]
```
JuanCab, line 279: Now confirmed to work on macOS and Linux 64.

JuanCab (dismissed):
Merge approved.

pyproject.toml
JuanCab: This closes both #630 and #631.

JuanCab (approved):
Finally fixed and passed all the tests.

PR #649, opened 2026-07-28
Fix no_catalog_error without colname_map and PS1→JC passband lookup

mwcraig:
Two small correctness fixes from #606 (items a and d), both developed test-first:

- **`CatalogData(..., no_catalog_error=True)` without `colname_map`**: the NaN `mag_error` auto-fill was gated on `colname_map is not None`, so the flag was silently ignored when the input table already used the standard column names — a missing `mag_error` column still raised. The gate now accepts `colname_map=None`. New regression test builds a standard-named table without `mag_error` and asserts it validates with an all-NaN `mag_error` column (it raised before the fix).

- **`PanStarrs1ToJohnsonCousinsMixin.__call__`**: the g−r color was computed as `from_magnitudes[..., 0] - from_magnitudes[..., 1]`, silently assuming gp1 and rp1 are the first two passbands. It now looks the bands up by name via the `from_band_index` dict, and raises a clear `ValueError` naming any missing band. Behavioral no-op for the shipped passband order; new tests pin that reordering the passbands doesn't change the result and that a missing gp1/rp1 raises.

Verified: both new tests failed before the fixes (`ValueError: data['mag_error'] is missing...` and a wrong-color mismatch / `KeyError`), pass after; full local suite green (793 passed, 47 skipped; remote-data tests skipped as usual).

This PR was written by Claude at Matt's direction.

https://claude.ai/code/session_01TEsxZvxjLqPN5PU8A14dpz


mwcraig (commented):
Looks good

PR #650, opened 2026-07-28
Structure cleanups: drop dead module, document bottleneck, share in-frame check

mwcraig:
Structure/dependency cleanups; completes #608.

- **Removed `stellarphot/utils/catalog_search.py`** — a 0-byte module with zero references repo-wide.
- **Documented the `bottleneck` dependency** in `pyproject.toml` (kept, per discussion): nothing imports it directly, but astropy's `sigma_clip` — which the photometry code calls heavily — uses it opportunistically as an accelerator. Comment matches the style of the neighboring `requests`/`regions` comments.
- **Extracted the duplicated in-frame containment check** into a shared `stellarphot.core.coords_in_frame(coords, wcs, shape, padding=0)` helper and rewired both call sites (`CatalogData.from_vizier`'s clip block and `comparison_utils.in_field`). `shape` is an explicit `(nx, ny)` argument because `wcs.pixel_shape` is frequently `None`. The helper lives in `core.py` rather than a new `utils` module because `core` importing anything from the `stellarphot.utils` package would run `utils/__init__.py → comparison_utils → catalogs → core` mid-import — a genuine circular-import failure; the `comparison_utils → core` direction is safe.

One deliberate, minor behavior change, noted in the changelog: `in_field` previously used strict bounds while `from_vizier` used inclusive ones; both now use the inclusive semantics, so a source landing exactly on the frame edge counts as in-field where `in_field` previously excluded it.

Developed test-first: new `TestCoordsInFrame` unit tests (padding margin, inclusive edges, non-square shape axis order) failed red before the helper existed and pass after; the pre-existing `test_in_field_non_square_image`, `from_vizier` clip tests, and the full local suite stay green (793 passed, 47 skipped).

This PR was written by Claude at Matt's direction.

https://claude.ai/code/session_01TEsxZvxjLqPN5PU8A14dpz

### DIFF
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -84,10 +84,13 @@ def coords_in_frame(coords, wcs, shape, padding=0):
 
     shape : tuple of int
         Pixel shape of the image, as ``(nx, ny)`` -- the same axis order as
-        `astropy.wcs.WCS.pixel_shape`. This is a required argument, rather
-        than being read from ``wcs.pixel_shape``, because ``pixel_shape`` is
-        frequently ``None`` (it is only set if the WCS was constructed from
-        a header with ``NAXIS`` keywords, or set explicitly).
+        `astropy.wcs.WCS.pixel_shape`, and the *reverse* of the numpy
+        ``array.shape`` order ``(ny, nx)``. If you are starting from an
+        array, pass ``array.shape[::-1]``. This is a required argument,
+        rather than being read from ``wcs.pixel_shape``, because
+        ``pixel_shape`` is frequently ``None`` (it is only set if the WCS
+        was constructed from a header with ``NAXIS`` keywords, or set
+        explicitly).
 
     padding : int, optional
         Minimum number of pixels required between a coordinate and the edge
--- a/stellarphot/core.py
+++ b/stellarphot/core.py
@@ -65,6 +65,51 @@ VIZIER_SERVERS = (
 VIZIER_TIMEOUT = 30
 
 
+def coords_in_frame(coords, wcs, shape, padding=0):
+    """
+    Determine which sky coordinates fall within an image frame.
+
+    This is the shared containment check used both by
+    `CatalogData.from_vizier` (to clip a Vizier catalog to the field of
+    view) and by `stellarphot.utils.comparison_utils.in_field`.
+
+    Parameters
+    ----------
+
+    coords : `astropy.coordinates.SkyCoord`
+        Coordinates to test.
+
+    wcs : `astropy.wcs.WCS`
+        WCS solution used to convert ``coords`` to pixel positions.
+
+    shape : tuple of int
+        Pixel shape of the image, as ``(nx, ny)`` -- the same axis order as
+        `astropy.wcs.WCS.pixel_shape`. This is a required argument, rather
+        than being read from ``wcs.pixel_shape``, because ``pixel_shape`` is
+        frequently ``None`` (it is only set if the WCS was constructed from
+        a header with ``NAXIS`` keywords, or set explicitly).
+
+    padding : int, optional
+        Minimum number of pixels required between a coordinate and the edge
+        of the frame for that coordinate to be considered in-frame. Default
+        is 0.
+
+    Returns
+    -------
+
+    `numpy.ndarray` of bool
+        Boolean array, the same length as ``coords``, ``True`` for the
+        coordinates that fall within the frame (and at least ``padding``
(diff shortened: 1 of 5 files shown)


stellarphot/tests/test_core.py
```diff
@@ -1084,6 +1085,68 @@ def test_from_vizier_with_header_no_wcs_raise_error():
...
+        # Corners exactly on the frame edge should be included; a point just
+        # past the edge should not.
+        wcs = _FakeWCS(x=[0.0, 400.0, 401.0], y=[0.0, 200.0, 100.0])
+        coords = _fake_coords(3)
+
+        mask = coords_in_frame(coords, wcs, (400, 200))
+
+        assert list(mask) == [True, True, False]
+
+    def test_non_square_shape_axes_are_not_swapped(self):
```
mwcraig, line 1139: Is this test actually meaningful since the x/y positions are hardcoded in the fake WCS?

stellarphot/core.py
```diff
@@ -65,6 +65,51 @@
...
+    ----------
+
+    coords : `astropy.coordinates.SkyCoord`
+        Coordinates to test.
+
+    wcs : `astropy.wcs.WCS`
+        WCS solution used to convert ``coords`` to pixel positions.
+
+    shape : tuple of int
+        Pixel shape of the image, as ``(nx, ny)`` -- the same axis order as
```
mwcraig, line 86: Is this opposite the usual numpy order? If it is then we should clarify that

stellarphot/tests/test_core.py
mwcraig: Yes — the hardcoded positions are the point, not a weakness. The unit under test here is not the WCS transform; it's the boundary arithmetic in `coords_in_frame`, specifically *which element of `shape` bounds which pixel coordinate*. The fake WCS pins the pixel positions so the test exercises only that comparison.

Concretely: the helper unpacks `nx, ny = shape` and checks `x <= nx`. If that unpacking were flipped to numpy order (`ny, nx = shape`), then with `shape=(400, 200)` the bound on `x` becomes 200, the hardcoded `x=300` fails it, and this test goes red. So it does pin real behavior — the `(nx, ny)` contract that `from_vizier` relies on when it passes `wcs.pixel_shape` straight through.

Two supporting notes:

- The fake WCS also serves a second purpose (noted in its docstring): a real world→pixel→world round trip lands at `-7e-12` instead of `0.0` right at pixel edges, which makes edge-exact assertions flaky. Faking the transform removes that noise.
- The end-to-end path with a *real* WCS and header isn't untested — the pre-existing `test_in_field_non_square_image` and the `from_vizier` clip tests cover that integration side.

One fair observation: `test_edge_pixel_counts_as_inside` in the same class would *also* catch an axis swap (its corner point `(400, 200)` fails under swapped bounds), so this test is partially redundant as a tripwire. Its value is diagnostic isolation — an interior point that fails *only* under an axis swap, not entangled with the edge-inclusivity semantics. Happy to drop it if you'd rather keep the class minimal, but I'd lean toward keeping it.

stellarphot/core.py
mwcraig: Confirmed — it is the opposite of numpy order, and that's deliberate. `(nx, ny)` matches `astropy.wcs.WCS.pixel_shape` (FITS/pixel axis order), while numpy's `array.shape` is `(ny, nx)`. Verified against astropy directly: a header with `NAXIS1=400, NAXIS2=200` gives `pixel_shape == (400, 200)` and `array_shape == (200, 400)`.

The choice lets `from_vizier` pass `wcs.pixel_shape` straight through, and the one numpy-shaped call site (`in_field`) already flips explicitly with a comment (`ccdy, ccdx = ccd.shape`). But you're right that the docstring only implies the numpy mismatch via the `pixel_shape` reference — worth stating outright. Proposed wording:

```
shape : tuple of int
    Pixel shape of the image, as ``(nx, ny)`` -- the same axis order as
    `astropy.wcs.WCS.pixel_shape`, and the *reverse* of the numpy
    ``array.shape`` order ``(ny, nx)``. If you are starting from an
    array, pass ``array.shape[::-1]``. This is a required argument,
    rather than being read from ``wcs.pixel_shape``, because
    ``pixel_shape`` is frequently ``None`` (it is only set if the WCS
    was constructed from a header with ``NAXIS`` keywords, or set
    explicitly).
```

Say the word and I'll push that docstring tweak to the branch.

mwcraig: Done — docstring updated in 93a200a to state the reversal outright, with the `array.shape[::-1]` hint.

PR #655, opened 2026-07-28
Make sky annulus track per-image FWHM with variable apertures

mwcraig:
Fixes #654.

## The problem

With `variable_aperture=True`, `single_image_photometry` measures the FWHM of each image and scales the aperture with it, but the sky annulus (and the edge-of-image rejection padding) was always computed from the static `fwhm_estimate` in the settings. This affects **every** image photometered in variable mode, not just an initial estimate: the annulus is rebuilt per image but always from the same stale number. On any frame whose measured FWHM satisfies `radius × measured_fwhm > radius × fwhm_estimate + gap`, the aperture reaches into its own sky annulus, contaminating the background estimate — worst on exactly the bad-seeing frames variable apertures are meant to handle.

## The fix

- `PhotometryApertures` (in `stellarphot/settings/models.py`) gains `inner_annulus_pixels(fwhm)` and `outer_annulus_pixels(fwhm)` methods, mirroring the existing `radius_pixels(fwhm)`.
- The `inner_annulus` / `outer_annulus` properties now delegate to the new methods with `fwhm_estimate`, so GUI/display behavior and the fixed-aperture path are unchanged.
- `single_image_photometry` builds the `CircularAnnulus` and the edge-of-image `padding` from the new methods using the per-image FWHM already in scope. In fixed mode that FWHM *is* `fwhm_estimate`, so nothing changes there.

## Tests

Written first and confirmed failing before the production change:

- `test_annulus_pixels_methods` in `stellarphot/settings/tests/test_models.py` covers the new methods in both variable and fixed modes and checks the properties still follow `fwhm_estimate`.
- `test_photometry_variable_aperture` in `stellarphot/photometry/tests/test_photometry.py` now asserts that the `annulus_inner`/`annulus_outer` output columns track each image's measured FWHM (frames with FWHM 5, 7.5 and 10 px) and that the aperture never overlaps its own annulus. Before the fix this failed on the 7.5 and 10 px frames, exactly as the issue predicts.

Full `stellarphot/settings` + `stellarphot/photometry` suites pass (367 passed, 16 expected skips).

---

https://claude.ai/code/session_01C7Caue41k2785ggZeAFAhW

### DIFF
--- a/stellarphot/settings/models.py
+++ b/stellarphot/settings/models.py
@@ -422,16 +422,18 @@ class PhotometryApertures(BaseModelWithTableRep):
     @property
     def inner_annulus(self):
         """
-        Radius of the inner annulus in pixels.
+        Radius of the inner annulus in pixels, based on the FWHM estimate
+        in the settings.
         """
-        return self.radius_pixels(self.fwhm_estimate) + self.gap
+        return self.inner_annulus_pixels(self.fwhm_estimate)
 
     @property
     def outer_annulus(self):
         """
-        Radius of the outer annulus in pixels.
+        Radius of the outer annulus in pixels, based on the FWHM estimate
+        in the settings.
         """
-        return self.inner_annulus + self.annulus_width
+        return self.outer_annulus_pixels(self.fwhm_estimate)
 
     def radius_pixels(self, fwhm):
         """
@@ -442,6 +444,20 @@ class PhotometryApertures(BaseModelWithTableRep):
             return fwhm * self.radius
         else:
             return self.radius
+
+    def inner_annulus_pixels(self, fwhm):
+        """
+        Return the inner annulus radius in pixels for the given FWHM,
+        depending on whether the aperture is variable or not.
+        """
+        return self.radius_pixels(fwhm) + self.gap
+
+    def outer_annulus_pixels(self, fwhm):
+        """
+        Return the outer annulus radius in pixels for the given FWHM,
+        depending on whether the aperture is variable or not.
+        """
+        return self.inner_annulus_pixels(fwhm) + self.annulus_width
 
 
 class PhotometryFileSettings(BaseModelWithTableRep):
--- a/stellarphot/settings/tests/test_models.py
+++ b/stellarphot/settings/tests/test_models.py
@@ -644,6 +644,41 @@ class TestPhotometryApertureSettings:
         radius_pix = ap_set.radius_pixels(fwhm)
         assert radius_pix == pytest.approx(7.5, rel=1e-6)
 
+    @pytest.mark.parametrize(
+        "variable_aperture,radius",
+        [
+            (True, 1.5),
+            (False, 5.0),
+        ],
+    )
+    def test_annulus_pixels_methods(self, variable_aperture, radius):
+        # The annulus radii should follow the FWHM passed in, just like
+        # radius_pixels does, so that in variable-aperture mode the annulus
+        # can track the per-image FWHM. See #654.
+        settings = deepcopy(TEST_APERTURE_SETTINGS)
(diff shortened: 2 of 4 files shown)


stellarphot/settings/tests/test_models.py
```diff
@@ -644,6 +644,41 @@ def test_create_aperture_settings_variable_aperture(self):
...
+        # can track the per-image FWHM. See #654.
+        settings = deepcopy(TEST_APERTURE_SETTINGS)
+        settings["variable_aperture"] = variable_aperture
+        settings["radius"] = radius
+        ap_set = PhotometryApertures(**settings)
+
+        # Deliberately different from the settings' fwhm_estimate
+        fwhm = 2 * settings["fwhm_estimate"]
+        if variable_aperture:
+            expected_inner = radius * fwhm + settings["gap"]
```
mwcraig, line 666: Shouldn't the gap be multiplied by the FWHM?

stellarphot/settings/models.py
```diff
@@ -443,6 +445,20 @@ def radius_pixels(self, fwhm):
         else:
             return self.radius
 
+    def inner_annulus_pixels(self, fwhm):
+        """
+        Return the inner annulus radius in pixels for the given FWHM,
+        depending on whether the aperture is variable or not.
+        """
+        return self.radius_pixels(fwhm) + self.gap
```
mwcraig, line 453: Same comment here -- I thought we decided that gap and width need to be scaled by FWH<?

mwcraig: Yes — that decision stands, and this line is where it will eventually change. The scaling is deliberately *not* in this PR, though.

As written, only `radius` is FWHM-scaled in variable mode (`radius_pixels` returns `fwhm * self.radius`); `gap` and `annulus_width` remain fixed pixel counts, so the annulus *position* tracks the per-image FWHM but its gap and width do not. That is the minimal fix from #654: it removes the failure mode where the scaled aperture grows into (or past) an annulus computed from the static `fwhm_estimate`, and with it the seeing-correlated annulus-position error.

Scaling `gap` and `annulus_width` is the agreed follow-up per the #654 second-round review ("Stance A, amended": in variable mode, gap ≈ 2×FWHM and width ≈ 1.5×FWHM defaults). It is gated on the settings-format versioning in #656, because it changes the *meaning* of the numbers in every saved settings file with `variable_aperture=True` — that semantic change should land behind a version field with an explicit migration, not silently in a bugfix.

To be explicit about what this PR does and does not buy: with a fixed-pixel gap the residual self-contamination bias still grows with FWHM — per the #654 closed-form analysis, a 5 px gap leaves roughly 22–43 mmag at FWHM 17 px (β 3.5–2.5). So this PR alone does not meet a <1 mmag target on the worst-seeing frames; it is the "necessary but not sufficient" piece, with the sufficiency piece queued behind #656.

*Reply written by Claude at @mwcraig's direction.*

stellarphot/settings/tests/test_models.py
mwcraig: The test matches the code as this PR intends it: in variable mode only `radius` is FWHM-scaled, so `expected_inner = radius * fwhm + settings["gap"]` — the gap is a fixed pixel count added on top. If the gap were scaled here the test would (correctly) fail against `inner_annulus_pixels`, which is `radius_pixels(fwhm) + self.gap`.

Scaling `gap` (and `annulus_width`) with FWHM is the agreed follow-up from #654, gated on the settings-format versioning in #656 since it changes what saved settings numbers mean. Fuller rationale in my reply on the `models.py` thread; when that follow-up lands, this expectation changes with it.

*Reply written by Claude at @mwcraig's direction.*

PR #656, opened 2026-07-30
Add settings-format version field and load-time migration

mwcraig:
Prerequisite for the variable-aperture geometry fix discussed in #654 (see [this comment](https://github.com/feder-observatory/stellarphot/issues/654#issuecomment-5120141234), amendment 3): the settings format needs to be versioned before the meaning of the annulus settings changes, so that old saved settings are never silently reinterpreted.

## What this does

- Adds a `settings_version` integer field to `PhotometrySettings` (and therefore to `PartialPhotometrySettings`). The current format is **2**; files written before the field existed are **format 1**. The value is a format generation, bumped only when the meaning of saved settings changes — it is deliberately not tied to release tags.
- Migrates format 1 files when they are read through `PhotometryWorkingDirSettings.load()`:
  - `variable_aperture=True`: the user's `radius`, `fwhm_estimate` and `variable_aperture` are kept, but `gap` and `annulus_width` are **reset to the model defaults** with a prominent `PhotometrySettingsMigrationWarning` explaining that variable-aperture photometry done with older versions is biased (#654) and should be redone. The reset values are read from the model field defaults, so when the geometry-change PR updates those defaults the migration follows automatically.
  - Fixed-aperture files load unchanged with no warning.
  - Either way the file carries `settings_version: 2` the next time it is saved.
- A file with a `settings_version` newer than the code understands raises `NewerFormatError` on every validation path. It intentionally does **not** subclass `ValueError`: pydantic folds `ValueError` from validators into `ValidationError`, and this error needs to escape unchanged so `save()` cannot mistake a too-new file for a missing one and overwrite it, and the GUI's `except ValueError` fallbacks don't silently show empty settings.

## Design notes

- The migration is a `model_validator(mode="before")` gated on validation context (`{"settings_file": True}`, supplied only by `load()`). Settings constructed in code — e.g. from GUI widget values — also lack a `settings_version` key and must not be treated as old files; the context gate prevents that.
- `_make_partial_model` now nulls the defaults of *required* fields only, so partial settings files record the actual version instead of `null`.
- `BaseModelWithTableRep` no longer defines a pass-through `__init__`. It was vestigial (its original body, registering table representers, was removed long ago) and it made pydantic construct instances through `__init__` during `model_validate*`, silently dropping the validation context.
- Settings embedded in the metadata of existing photometry tables are read as-is with no migration or warning — tables record photometry that has already been done.

## Testing

- New model-level tests in `TestPriorVersionsCompatibility` (defaulted version, partial-model version not null, migration only with context, fixed-mode untouched, too-new error).
- New `PhotometryWorkingDirSettings` tests covering full and partial format 1 files, upgrade-on-save, refusing to clobber a newer-format file, and loading the real 2.0.0-alpha sample settings file (which is a format 1 file with `variable_aperture: true`).
- Full test suite passes locally.

---
*This PR was written by Claude at @mwcraig's direction; the migration behavior (keep the aperture radius, reset the annulus geometry, warn loudly) and the integer version scheme are @mwcraig's decisions.*

https://claude.ai/code/session_01C7Caue41k2785ggZeAFAhW


stellarphot/settings/tests/test_settings_file.py
```diff
@@ -593,3 +600,171 @@ def test_save_update_completing_partial_makes_full(self):
...
+        assert apertures.variable_aperture is True
+        assert apertures.radius == partial["photometry_apertures"]["radius"]
+        assert apertures.gap == PhotometryApertures.model_fields["gap"].default
+        assert (
+            apertures.annulus_width
+            == PhotometryApertures.model_fields["annulus_width"].default
+        )
+        assert loaded.settings_version == PHOTOMETRY_SETTINGS_FORMAT_VERSION
+
+    def test_load_partial_format1_no_apertures_no_warning(self):
```
mwcraig, line 1142: Add comments here explaining what the test does.

stellarphot/settings/tests/test_settings_file.py
```diff
@@ -593,3 +600,171 @@ def test_save_update_completing_partial_makes_full(self):
...
+        )
+        raw = json.loads(settings_file.partial_settings_file.read_text())
+        assert raw["settings_version"] == PHOTOMETRY_SETTINGS_FORMAT_VERSION
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            PhotometryWorkingDirSettings().load()
+
+    @pytest.mark.parametrize("full_settings", [True, False])
+    def test_load_newer_format_version_raises(self, full_settings):
```
mwcraig, line 1188: Explain

stellarphot/settings/tests/test_settings_file.py
```diff
@@ -593,3 +600,171 @@ def test_save_update_completing_partial_makes_full(self):
...
+        if full_settings:
+            file = settings_file.settings_file
+        else:
+            file = settings_file.partial_settings_file
+        file.write_text(json.dumps(newer))
+
+        with pytest.raises(NewerFormatError, match="newer version of stellarphot"):
+            settings_file.load()
+
+    def test_save_refuses_to_clobber_newer_format_file(self):
```
mwcraig, line 1205: Explain

mwcraig (commented):
In addition to the inline comments, there is an additional issue: the default values for the annulus gap and width (and aperture for that matter) were chosen assuming a fixed aperture size. Is there a way to have different defaults for the fixed aperture and variable aperture cases?

mwcraig (commented):
Critical review, done by checking out the branch and running verification code against it (pydantic 2.12.5). Overall the mechanism is sound — I verified the context gate, the save-clobber protection, and the partial-file version retention all behave as described, and the pydantic custom-`__init__` claim is real (a passthrough `__init__` really does drop the validation context; verified below). But I found one behavioral problem I think should block merging as-is, one significant design inconsistency, and a couple of structural issues.

## 1. The migration resets the annulus to a 1-pixel-wide ring that photometry will happily run with, headlessly

`PhotometryApertures.model_fields["gap"].default` and `["annulus_width"].default` are both **1** today. So the migration takes a realistic old file and produces a nearly degenerate sky annulus. Verified with the golden file shipped in this PR:

```python
golden = json.loads(Path("stellarphot/settings/tests/data/sample_photometry_settings_2.0.0alpha.json").read_text())
wd = PhotometryWorkingDirSettings()
wd.settings_file.write_text(json.dumps(golden))
loaded = wd.load()          # PhotometrySettingsMigrationWarning fires
ap = loaded.photometry_apertures
print(ap.model_dump())
# {'variable_aperture': True, 'radius': 1.4, 'gap': 1.0, 'annulus_width': 1.0, 'fwhm_estimate': 10.098...}
print(ap.inner_annulus, ap.outer_annulus)
# 15.137805709022222 16.137805709022222   <- a 1-pixel-wide sky annulus
```

The old values were `gap=5.0, annulus_width=15.0`. The warning is the only guardrail, and the main consumer of `load()` never shows it to the user: `PhotometryRunner` executes `photometry_runner.ipynb` through papermill (`stellarphot/gui/custom_widgets.py:1059`), and that notebook does

```python
wd_settings.load()
ap_photom = AperturePhotometry(settings=wd_settings.settings)
photometry = ap_photom(...)
```

so the `PhotometrySettingsMigrationWarning` lands in the output of a generated notebook nobody reads, and photometry runs to completion with a 1-px annulus. The PR's stated goal is to stop old settings from being *silently* misinterpreted, but on this path it replaces "annulus doesn't scale" with "annulus is 1 px wide", still silently.

Suggestions, in rough order of preference:

- Land this together with (or after) the geometry PR that gives `gap`/`annulus_width` meaningful defaults, so migration never produces placeholder values; or
- make a migrated-but-not-yet-reviewed file *unrunnable*: e.g. the migration raises (like `NewerFormatError` does) when invoked from a non-interactive load, or nulls the aperture fields so only a `PartialPhotometrySettings` results and `AperturePhotometry` refuses to start until the user re-saves; or
- have the migration keep the settings loadable for *review* but mark them (a flag in the model or on `PhotometryWorkingDirSettings`) so `AperturePhotometry` can refuse to run on unreviewed migrated settings.

## 2. `check_format_version_not_newer` makes photometry *data files* from a newer stellarphot unreadable

The version check runs on every validation path, including the table-metadata path that the PR itself argues (comment in `table_representations.py`) is a historical record where migration deliberately does not apply. The same logic applies to the too-new check — a table records photometry that has already been done — but the check still fires there:

```python
t = Table({"a": [1, 2]})
t.meta["photometry_settings"] = PhotometrySettings(**TEST_PHOTOMETRY_SETTINGS)
serialize_models_in_table_meta(t.meta)
t.meta["photometry_settings"]["settings_version"] = 99   # written by future stellarphot
t.write("phot.ecsv")

t2 = Table.read("phot.ecsv")
deserialize_models_in_table_meta(t2.meta)
# NewerFormatError: These settings have settings_version 99, but this version of
# stellarphot only understands settings_version 2 or earlier...
```

`BaseEnhancedTable.read` calls `deserialize_models_in_table_meta` (`stellarphot/core.py:402`), so once format 3 exists, a `PhotometryData` ECSV produced by a newer stellarphot cannot be opened at all by a 2.2.x install — even just to look at the fluxes. Sharing data files between machines/collaborators on different versions seems like the common case this will hit.

The context mechanism you added is exactly the right tool: pass a context (e.g. `{"table_meta": True}`) from `deserialize_models_in_table_meta` / the YAML `_constructor` and skip the too-new check there — or catch `NewerFormatError` at those two call sites and leave the raw dict in the metadata.

## 3. `__all__` is now doing double duty, and the exception classes leak into the YAML machinery

Adding `NewerFormatError` and `PhotometrySettingsMigrationWarning` to `models.__all__` is right for the public API, but `table_representations` consumes `__all__` in three places as "the list of serializable models". Verified:

```python
_generate_old_table_representers()
NewerFormatError in AstropyDumper.yaml_representers        # True
"!PhotometrySettingsMigrationWarning" in AstropyLoader.yaml_constructors  # True
```

So exception classes now have YAML representers whose body calls `model.model_dump_json()`, constructors that call `cls.model_validate_json(...)`, and they're in `known_models` in `deserialize_models_in_table_meta` (a stray `"_model_name": "NewerFormatError"` would die with `AttributeError` instead of a sane error). None of this triggers today — and `FwhmMethods` (a `StrEnum`) already had the same latent problem — but the new `# Everything in __all__ must be a class` comment is a fragile contract enforced by nothing. Filtering in `table_representations` instead would fix all three consumers and remove the comment's constraint:

```python
for model_name in models.__all__:
    model_class = getattr(models, model_name)
    if isinstance(model_class, type) and issubclass(model_class, BaseModel):
        ...
```

## 4. Compatibility note worth documenting

Older releases (≤ 2.1.x) reading a format-2 file don't get anything like `NewerFormatError` — because of `extra="forbid"` they fail with the cryptic `Extra inputs are not permitted [type=extra_forbidden]` on `settings_version` (verified against `main`). Nothing this PR can do retroactively, but since the friendly error only exists going forward, a sentence in CHANGES telling users what the *old*-version failure looks like ("if you see `Extra inputs are not permitted` on `settings_version`, the file was written by stellarphot ≥ 2.2") would save some confusion.

## Verified claims (no action needed)

- The `BaseModelWithTableRep.__init__` removal rationale is correct. On pydantic 2.12.5, a model with a passthrough `__init__` receives `info.context = None` during `model_validate(..., context=...)`; without the `__init__` the context arrives intact. The migration tests would catch a regression if someone re-adds an `__init__`.
- `save()` genuinely refuses to clobber a newer-format file on all paths I tried (full and partial, `update` True/False), because the internal `load()` raises before any write.
- No circular import from the new `constants.py` → `models.py` import.
- Overwriting a format-1 file with fresh full settings preserves the new values correctly (the spurious extra migration warning during `save()`'s internal `load()` is acknowledged in the tests; it's noise but harmless).

## Nit

- `stacklevel=2` in `_migrate_settings_file`'s `warnings.warn` points into pydantic's validator machinery, not user code. Since no stacklevel is meaningful from inside a validator, plain `warnings.warn(..., PhotometrySettingsMigrationWarning)` is at least honest about the source.

stellarphot/settings/models.py
```diff
@@ -1144,6 +1200,75 @@ class PhotometrySettings(BaseModelWithTableRep):
...
+        gate, settings constructed in code -- for example from GUI widget
+        values, which have no settings_version key either -- would be
+        "migrated" too.
+        """
+        context = info.context or {}
+        if not context.get("settings_file") or not isinstance(data, dict):
+            return data
+        if data.get("settings_version") is not None:
+            return data
+
```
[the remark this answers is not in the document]
mwcraig: The described behavior is real — verified on this branch (pydantic 2.13.4): a file containing `"settings_version": 1` with `variable_aperture: true` loads with no warning and its old `gap`/`annulus_width` intact, and after `save()` the file on disk still says `"settings_version": 1`, so it skips migration forever.

Two corrections to the framing, though:

- No released stellarphot has ever *written* `"settings_version": 1`. The field first appears in this PR with value 2, and format 1 is defined (see `docs/dev/settings.rst`) as "the field is absent". The only way to produce the file above is to hand-edit one.
- "contradicting the intent that old files upgrade to the current format on next save" overstates it: every file an old version actually wrote (no key) *does* upgrade on next save — that is exactly what `test_format1_file_upgraded_on_save` asserts (the raw saved file contains `settings_version: 2`). The intent is not contradicted for any file stellarphot has produced.

That said, treating an explicit `1` the same as a missing key is consistent with the documented meaning of the field and is cheap hardening against hand-edited files. Note the suggested change has a second required piece, as you say: the migration would then also have to overwrite the version in the data — today the upgrade to 2 happens because the *missing* field picks up the model default, whereas an explicit `1` would be kept. Whether to fold this hardening into this PR is @mwcraig's call.

*(Reply written by Claude at @mwcraig's direction.)*

stellarphot/settings/models.py
```diff
@@ -34,22 +35,55 @@
...
     "LoggingSettings",
+    "NewerFormatError",
     "PartialPhotometrySettings",
     "PassbandMap",
     "PhotometryApertures",
     "PhotometryFileSettings",
     "PhotometryRunSettings",
     "PhotometrySettings",
+    "PhotometrySettingsMigrationWarning",
     "PhotometryOptionalSettings",
```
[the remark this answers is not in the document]
mwcraig: The crash scenario is real — verified on this branch:

```python
deserialize_models_in_table_meta({"x": {"_model_name": "NewerFormatError", "value": "{}"}})
# AttributeError: type object 'NewerFormatError' has no attribute 'model_validate'
```

But the suggested remedy rests on an invariant that did not hold before this PR: `FwhmMethods` is a `StrEnum`, has been in `models.__all__` on `main` all along, and fails identically today:

```python
deserialize_models_in_table_meta({"x": {"_model_name": "FwhmMethods", "value": '"fit"'}})
# AttributeError: type object 'FwhmMethods' has no attribute 'model_validate'
```

So "keep `models.__all__` restricted to pydantic models" and moving only the two new classes into `stellarphot/settings/__init__.py` would leave the same latent crash in place, and `__all__` would remain an unenforced contract with three separate consumers in `table_representations`.

The more robust resolution is to filter where `__all__` is consumed — `isinstance(obj, type) and issubclass(obj, BaseModel)` in `generate_table_representers`, `_generate_old_table_representers`, and `deserialize_models_in_table_meta` — which also fixes the pre-existing `FwhmMethods` case and lets the fragile "everything in `__all__` must be a class" comment be deleted. That is the direction planned for this PR (same conclusion as point 3 of the long review on this PR).

*(Reply written by Claude at @mwcraig's direction.)*

stellarphot/settings/tests/test_settings_file.py
mwcraig: Explanation (to be added as comments in the test in the next push):

This test pins the *negative* case: the migration must not fire when there is nothing to migrate. It writes a format-1 **partial** settings file in the old on-disk style — every key present, `null` for unset values, no `settings_version`, and only `camera` actually filled in — so there is no aperture section for the migration to touch. It then asserts that

- `load()` is completely silent (`warnings.simplefilter("error")` promotes any warning, including `PhotometrySettingsMigrationWarning`, to a test failure),
- the camera value survives the load, and
- the loaded settings report the current format version (the missing field picks up the model default), so the next save writes an up-to-date file.

Rationale: a migration warning that fired on every old file regardless of content would train users to ignore it. It must be reserved for files where `gap`/`annulus_width` were actually reset.

*(Reply written by Claude at @mwcraig's direction.)*

stellarphot/settings/tests/test_settings_file.py
mwcraig: Explanation (to be added as comments in the test in the next push):

This test writes a settings file claiming `settings_version` one greater than this code understands (`PHOTOMETRY_SETTINGS_FORMAT_VERSION + 1`, i.e. a file from a hypothetical future stellarphot) and asserts that `load()` raises `NewerFormatError` carrying the "written by a newer version of stellarphot" message rather than mis-reading the file under current-format assumptions.

It is parametrized over `full_settings` because `PhotometryWorkingDirSettings.load()` validates the two files through separate calls — `PhotometrySettings.model_validate_json` for the full settings file and `PartialPhotometrySettings.model_validate_json` for the partial one (`settings_files.py`, in `load()`) — and both paths must refuse a too-new file. The partial path is the one that would be easy to break accidentally, since `PartialPhotometrySettings` is generated by `_make_partial_model` and could plausibly lose the field validator in a refactor.

*(Reply written by Claude at @mwcraig's direction.)*

stellarphot/settings/tests/test_settings_file.py
mwcraig: Explanation (to be added as comments in the test in the next push):

This is the data-loss guard, and it pins the design decision that `NewerFormatError` subclasses `Exception` rather than `ValueError`. Setup: a full settings file from a "future" format is on disk, and the test calls `save(partial, update=True)`. `save()` begins with

```python
try:
    _ = self.load()
except ValueError:
    # ... we have no settings files. We can proceed to save the settings.
    pass
```

i.e. it interprets a `ValueError` as "nothing usable on disk, safe to write". If `NewerFormatError` were a `ValueError` — or were raised inside a validator as one, which pydantic folds into `ValidationError` and `load()` then re-raises as `ValueError` — that `except` would swallow it and `save()` would overwrite a file written by a newer stellarphot with a downgraded one.

The two asserts check both halves of the guarantee: `NewerFormatError` propagates out of `save()`, and the bytes on disk are unchanged afterward.

*(Reply written by Claude at @mwcraig's direction.)*

stellarphot/settings/tests/test_settings_file.py
mwcraig: Done in 21f91fc — the test now carries the explanation as comments: it pins the negative case (migration must not fire when there is nothing to migrate, so the warning stays reserved for files where `gap`/`annulus_width` were actually reset), describes the old on-disk partial-file style being reproduced, and notes that `simplefilter("error")` turns any warning on load into a failure and that the missing `settings_version` picks up the current default.

stellarphot/settings/tests/test_settings_file.py
mwcraig: Done in 21f91fc — the test now explains in comments that a file claiming a too-new `settings_version` must raise `NewerFormatError` rather than be mis-read, and why it is parametrized over full/partial: the two files are validated through separate `model_validate_json` calls, and the partial path is the fragile one since `PartialPhotometrySettings` is generated by `_make_partial_model` and could lose the validator in a refactor.

stellarphot/settings/tests/test_settings_file.py
mwcraig: Done in 21f91fc — the test now documents that it is the data-loss guard pinning the decision that `NewerFormatError` subclasses `Exception` rather than `ValueError` (a `ValueError` would be swallowed by the `except ValueError` at the top of `save()` and the newer file overwritten), and that the two asserts check both halves of the guarantee: the error propagates out of `save()`, and the bytes on disk are unchanged.

CHANGES.rst
```diff
@@ -1,6 +1,27 @@
 2.2.0 (unreleased)
 -------------------
 
+New Features
+^^^^^^^^^^^^
++ Saved photometry settings files now carry a ``settings_version`` field that
```
mwcraig: Shorten this.

stellarphot/settings/settings_files.py
```diff
@@ -22,8 +22,9 @@
     "SettingsFileReadError",
 ]
 
-# We will have to version settings formats, I think. Hopefully this changes rarely
-# or never.
+# This versions the *directory* in which the global camera/observatory/passband
+# settings are stored, via PlatformDirs below. The version of the on-disk
+# settings file *format* is PHOTOMETRY_SETTINGS_FORMAT_VERSION in models.py.
 SETTINGS_FILE_VERSION = "2"  # value chosen to match major version of stellarphot
```
mwcraig: Wait --- what exactly does this version and how is it different from the version number we are introducing in this PR?

stellarphot/settings/models.py
```diff
@@ -1171,6 +1224,78 @@ class PhotometrySettings(BaseModelWithTableRep):
...
+        apertures = data.get("photometry_apertures")
+        if isinstance(apertures, dict) and apertures.get("variable_aperture"):
+            gap_default = PhotometryApertures.model_fields["gap"].default
+            width_default = PhotometryApertures.model_fields["annulus_width"].default
+            data = data | {
+                "photometry_apertures": apertures
+                | {"gap": gap_default, "annulus_width": width_default}
+            }
+            warnings.warn(
+                "\nIMPORTANT: your saved photometry settings were created by "
```
mwcraig: Super wordy -- where exactly does this end up being presented to the user?

stellarphot/settings/tests/test_models.py
```diff
@@ -460,6 +471,103 @@ def test_reading_2_0_0_alpha_files(self):
...
+        """
+        Settings dict as written by stellarphot before the settings_version
+        field existed, i.e. settings format 1.
+        """
+        old_style = deepcopy(TEST_PHOTOMETRY_SETTINGS)
+        old_style.pop("settings_version", None)
+        old_style["photometry_apertures"]["variable_aperture"] = variable_aperture
+        return old_style
+
+    def test_settings_version_defaults_to_current(self):
```
mwcraig, line 490: Add explanatory comments to this testt and all others and the PR that do not have them

docs/dev/settings.rst
```diff
@@ -25,5 +25,35 @@ related to the aperture photometry. It is used in
...
+Settings format versioning
+==========================
+
+Saved photometry settings files carry a ``settings_version`` field, an integer
+that identifies the version of the on-disk settings *format* (it is not the
+stellarphot version). Files written before the field existed are format ``1``;
+the current format version is ``PHOTOMETRY_SETTINGS_FORMAT_VERSION`` in
+``settings/models.py``.
+
+When ``PhotometryWorkingDirSettings.load`` reads a settings file it passes the
```
mwcraig, line 40: Why are implementation details in the user documentation?

CHANGES.rst
mwcraig: Agreed — at 15 lines it is the longest entry in the file. Proposed replacement (to be applied in the next push unless you'd like it trimmed differently):

```rst
+ Saved photometry settings files now carry a ``settings_version`` field;
  files without it are format ``1``. On load, format-1 settings with
  ``variable_aperture=True`` get their ``gap`` and ``annulus_width`` reset
  to defaults, with a warning, because the old variable-aperture annulus
  geometry biased the photometry (see issue ``#654``). A file written by a
  newer stellarphot raises ``NewerFormatError`` instead of being misread
  or overwritten, and older stellarphot versions fail on the new files
  with a pydantic error about the unexpected ``settings_version`` field. [#656]
```

This keeps the four load-bearing facts (version field, loud variable-aperture reset, newer-file protection, old-version behavior) and drops the mechanics (which values are kept vs reset, table-metadata details) that are documented in the warning text and docstrings.

stellarphot/settings/settings_files.py
mwcraig: `SETTINGS_FILE_VERSION` predates this PR (it has been on `main` since the 2.0 settings work) and versions a completely different thing: the *location* of the global settings store, not the content of any file. It is passed to `PlatformDirs("stellarphot", version=SETTINGS_FILE_VERSION)` (`settings_files.py`, in `SavedSettings`), so it becomes part of the OS-specific directory path where the shared cameras/observatories/passband maps live (e.g. `~/Library/Application Support/stellarphot/2` on macOS). Bumping it points stellarphot at a fresh, empty directory — a "start over" switch for the global store, with no migration semantics at all.

`PHOTOMETRY_SETTINGS_FORMAT_VERSION` (this PR) versions the *schema of the contents* of the per-working-directory photometry settings file, and drives the load-time migration.

So: one names *where the global store lives*, the other says *how to interpret what's inside a photometry settings file*. They can change independently. The comment this PR added is exactly to answer the question you just asked; if the name itself is the problem, `SETTINGS_DIR_VERSION` would be more honest, but note it is re-exported in `__all__`, so renaming is a (minor) API change.

stellarphot/settings/models.py
mwcraig: Where it surfaces, concretely:

1. **`ReviewSettings` GUI** — the primary surface. It is a `PhotometrySettingsWarning`, the category the banner from #663 displays: the full text lands in an `ipw.HTML` banner at the top of the widget and stays until dismissed.
2. **Scripts/notebooks calling `PhotometryWorkingDirSettings.load()` directly** — ordinary Python warning to stderr.
3. **The papermill-executed photometry notebook** — captured in the output notebook written to the working directory (recorded, not actively surfaced).

Agreed it is too wordy for a banner. Proposed replacement (~55 words vs ~95, same facts):

> IMPORTANT: your settings were saved by an older stellarphot with variable_aperture=True. The old variable-aperture geometry biased photometry (see issue #654), so 'gap' and 'annulus_width' have been reset to their defaults (gap={gap_default}, annulus_width={width_default}). Review and re-save your aperture settings; variable-aperture photometry done with older versions should be redone.

Happy to apply in the next push.

stellarphot/settings/tests/test_models.py
mwcraig: Will do in the next push: this one (`test_settings_version_defaults_to_current`) plus a sweep of every test in the PR diff for missing explanations — from a quick pass, the other bare ones are `test_format1_file_upgraded_on_save` in `test_settings_file.py` and the `_format_1_settings` helper's remaining users; the tests added today (`test_table_representations.py`, the too-new-metadata test in `test_core.py`, and the three from the earlier threads) already carry them.

docs/dev/settings.rst
mwcraig: This file is the *developer* documentation, not the user documentation: `docs/dev/settings.rst` lives in the `dev/` tree, which `docs/index.rst` pulls in as the "Contributing" section (`dev/index.rst`), and this same file already opens with contributor notes about the ipyautoui workarounds. The section was added deliberately as a contributor checklist for the next person who changes the settings format — bump `PHOTOMETRY_SETTINGS_FORMAT_VERSION`, extend `_migrate_settings_file`, add tests to the named test classes — which has no user-facing home.

The user-facing story (what the migration warning means, "upgrade stellarphot" for too-new files) is carried by the warning text and the changelog instead.

If the objection is that the *Contributing* section shouldn't name specific functions and test classes: that specificity is the point of a checklist — vague instructions here would send the next format change hunting through the code this PR just wrote. But if you'd rather slim it to principles and let the code comments carry the specifics, say the word.

stellarphot/settings/settings_files.py
mwcraig: The purpose of the directory versioning was to help track which settings go with which version of stellarphot, i.e. a crude way of tracking settings versions that anticipated that at some point the version style would change. Given that this version is already 2, perhaps we make it 3 and make the two version numbers the same instead of keeping them separate. The "implementation" here is to ask more questions about this.

stellarphot/settings/models.py
mwcraig: Question about this --don't the PRs merged before this automatically migrate settings already? In that case, does this PR issue a special message in addition to or instead of that generic "your settings have been updated" message?

This item needs more discussion.

CHANGES.rst
mwcraig: Applied in 6f9bff8 — the entry is now the 8-line version proposed above.

stellarphot/settings/tests/test_models.py
mwcraig: Done in 6f9bff8. Comments added to `test_settings_version_defaults_to_current` (this thread) and, from the sweep of the PR diff, to the other tests that lacked an explanation: `test_migration_of_unversioned_variable_aperture_with_context` in `test_models.py`, and `test_load_format1_variable_aperture_migrates_and_warns`, `test_format1_file_upgraded_on_save`, and `test_load_partial_format1_variable_aperture_warns` in `test_settings_file.py`. The golden-file test also gained a note on why it exists (it guards against the format-1 test helper drifting from what old versions actually wrote). The remaining tests in the diff already carried explanations.

stellarphot/settings/models.py
mwcraig: To the question: no — nothing merged before this PR migrates settings, and there is no generic "your settings have been updated" message anywhere. #655 changed only how the annulus is *computed* at photometry time and never touches settings files. #663 built the banner *display* for `PhotometrySettingsWarning`, but nothing on `main` emits a warning of that category — #663's own changelog entry says "Nothing emits the warning yet; the settings-format migration that will is being developed separately", i.e. this PR. So this warning is the first and only message the banner will show; it is not in addition to anything. (The only other migration in the codebase, the stellarphot 1→2 photometry *data* migration in `version_migrator.py`, is unrelated to settings files.)

Per discussion with @mwcraig, the warning is now an even shorter two-sentence message (57c11f0): the reset itself plus "review and re-save", with the details (what was kept, why old variable-aperture photometry should be redone) delegated to issue #654 and the changelog.

stellarphot/settings/settings_files.py
mwcraig: Per discussion with @mwcraig, deciding whether to unify the two numbers at `3` is deferred to a separate conversation; this PR keeps them as they are for now. The open questions for that discussion: whether bumping the directory version should copy the existing global store (cameras/observatories/passband maps) forward into the new directory or start fresh, and whether coupling the two numbers — every future format bump relocating the store — is acceptable. Renumbering this PR's format to 3 stays free until 2.2.0 is released, since no released version writes `settings_version: 2`.

PR #657, opened 2026-07-30
Show settings-load problems in a banner in ReviewSettings

mwcraig:
Previously, any failure to read the saved settings files in the working directory was silently swallowed by `ReviewSettings` (`except ValueError` around `PhotometryWorkingDirSettings().load()`), so the widget came up with defaults and no indication that an existing settings file had been ignored.

This PR adds a banner above the settings tabs that reports:

- an existing settings file that cannot be read (the widget still starts from defaults and tells the user why; the unreadable file is preserved by renaming it with a `.bak` suffix before any save would overwrite it), and
- any warnings generated while loading settings.

The warnings path is groundwork for #656: when the settings-format migration lands, its migration warning will be displayed in this banner automatically, addressing the concern that migration messages would otherwise never be seen by GUI users. Only the new `PhotometrySettingsWarning` category (a `UserWarning` subclass) is routed to the banner, so plain `UserWarning`s and deprecation noise from dependency internals stay out of it. Note for #656: the migration warning must be emitted as `PhotometrySettingsWarning` to appear in the banner.

Also fixes a latent crash this exposed: `PhotometryWorkingDirSettings.save(partial, update=True)` with an existing-but-unreadable full settings file raised `AttributeError` (`self._settings` is `None` after the failed load) — exactly what `ReviewSettings`' automatic default-saving does during construction, which would have killed the widget before the banner could be shown.

Changes from the review of this PR:

- `save()` never destroys a settings file it could not read: whether the unreadable file is the partial or the full one, it is renamed to `<name>.json.bak` before the write or delete that would have clobbered it, and existing backups are never overwritten (numbered `.bak1`, `.bak2`, ... are used instead). Writes go to a temporary file that atomically replaces the target, and the partial settings file is disposed of only after the full settings are safely on disk.
- Banner text derived from file contents (pydantic error text, warning messages) is HTML-escaped at message-construction time; the full error is shown in a `<pre>` block inside a collapsed `<details>` element.
- Warnings emitted by a `load()` that then raises are still shown alongside the error, instead of being dropped.
- `current_settings` is now a side-effect-free snapshot; reloading from disk and updating the banner happens only through an explicit `_refresh()`.
- Dismissing the load-error message does not suppress a later, *different* load error: dismissals are fingerprinted on the error's full text (first lines of pydantic errors are too generic to tell two corruptions of the same file apart), so a new incident that reuses the single load-error slot is shown again.
[...]

PR #661, opened 2026-08-08
Fix three crashes/dead code in settings save and ReviewSettings

mwcraig:
Extracted from #657 as part of decomposing that PR into small, reviewable pieces. These are the pure bug fixes — each exists on `main` today and is independent of the settings banner work.

1. **`PhotometryWorkingDirSettings.save(partial, update=True)` crashed with `AttributeError` when an existing full settings file could not be read.** The failed `load()` is swallowed, leaving `self._settings` as `None`, and the merge step then called `None.model_dump()`. Since `ReviewSettings.__init__` autosaves through exactly this path, the widget could not be constructed at all in a working directory with a corrupt settings file. Now the partial settings are saved on their own — there is nothing to merge them into.
2. **Collapsing every section of a `ReviewSettings` accordion raised `TypeError`.** An accordion reports `selected_index = None` when all sections are collapsed, and the selection observer indexed a list with it.
3. **Selecting a tab whose setting is not saved on disk never marked the tab.** The observer rebound a local variable (`setting_widget = SaveStatus.SETTING_NOT_SAVED`) instead of setting the widget's badge.

Each fix has a regression test that fails on `main`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -69,6 +69,14 @@ Bug Fixes
   to original-image pixel coordinates with an off-by-0.5 formula), and it no
   longer mutates the caller's ``CCDData.mask`` in place when flagging pixels
   above ``max_adu``. [#651]
++ ``PhotometryWorkingDirSettings.save(partial_settings, update=True)`` no
+  longer raises ``AttributeError`` when an existing full settings file
+  cannot be read; the partial settings are saved on their own since there
+  is nothing to merge them into. In ``ReviewSettings``, collapsing every
+  section of an accordion no longer raises ``TypeError``, and selecting a
+  tab whose setting is not saved on disk now actually marks the tab as not
+  saved (the selection observer previously rebound a local variable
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -69,6 +69,14 @@ Bug Fixes
   to original-image pixel coordinates with an off-by-0.5 formula), and it no
   longer mutates the caller's ``CCDData.mask`` in place when flagging pixels
   above ``max_adu``. [#651]
++ ``PhotometryWorkingDirSettings.save(partial_settings, update=True)`` no
```
mwcraig, line 72: Shorten this up a bit

mwcraig: This request arrived after the PR's last commit and the entry merged at 8 lines. It has now been shortened to 5 lines (all three fixes kept, the parenthetical mechanics dropped) in commit 14a1a46 on the open PR #656, which was already editing ``CHANGES.rst``.

PR #662, opened 2026-08-08
Make settings writes safe: atomic replace and .bak preservation of unreadable files

mwcraig:
Extracted from #657 as part of decomposing that PR into small, reviewable pieces. Built on top of #661 (the first two commits here are that PR; this one should be rebased or merged after it lands).

On `main`, three write paths can silently destroy a user's settings:

1. `save()` opens the settings file with `open("w")` **before** serializing, so a pydantic serialization error, a full disk, or a killed process leaves a zero-length or half-written settings file.
2. Saving full settings **unlinks the partial settings file before writing**, so a failed write can leave the directory with no settings at all.
3. A settings file that exists but cannot be parsed is silently overwritten by the next save — including the autosave `ReviewSettings.__init__` performs, so simply opening the widget can destroy a hand-recoverable file.

This PR makes those paths safe:

- Settings are serialized first and written to a **temporary file that atomically replaces** the target.
- The partial settings file is disposed of **only after** new full settings are safely on disk.
- An unreadable settings file is **renamed to `.bak`** (numbered `.bak1`, `.bak2`, ... — existing backups are never overwritten) instead of being overwritten or deleted, and a save sets aside an unreadable full settings file even after a partial-only write, so a save never leaves the directory unloadable.
- `load()` now parses **both** settings files before raising and latches which were unreadable, so `save()` merges into whichever file was readable; encoding and OS errors during a read are reported as `ValueError` like corrupt JSON instead of escaping raw.

No widget/UI changes. The larger banner/load-error UX from #657 is deliberately not included.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -77,6 +77,18 @@ Bug Fixes
   tab whose setting is not saved on disk now actually marks the tab as not
   saved (the selection observer previously rebound a local variable
   instead of setting the badge). [#661]
++ ``PhotometryWorkingDirSettings`` no longer destroys settings on failed
+  or conflicting writes. Settings are written to a temporary file that
+  atomically replaces the target, so an interrupted write cannot
+  truncate an existing settings file; the partial settings file is
+  disposed of only after new full settings are safely on disk; and a
+  settings file that exists but cannot be read is renamed to a ``.bak``
+  backup (numbered ``.bak1``, ``.bak2``, ... if needed) instead of being
+  silently overwritten or deleted. ``load()`` now parses both settings
+  files before raising so a save can merge into whichever file was
+  readable, and a file that is unreadable because of an encoding or
+  operating-system error is reported the same way as corrupt JSON
+  instead of raising an unexpected exception type. [#662]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -77,17 +77,13 @@ Bug Fixes
   tab whose setting is not saved on disk now actually marks the tab as not
   saved (the selection observer previously rebound a local variable
   instead of setting the badge). [#661]
-+ ``PhotometryWorkingDirSettings`` no longer destroys settings on failed
-  writes. Settings are written to a temporary file that atomically
-  replaces the target, the partial settings file is removed only after
-  full settings are safely on disk, and an unreadable settings file is
-  set aside as a numbered ``.bak`` backup instead of being overwritten.
-  Encoding and OS errors while reading settings now raise
-  ``SettingsFileReadError``, a ``ValueError`` subclass. A partial save
-  no longer resurrects values from a conflicting partial settings file;
-  the conflicting file is set aside as a ``.bak`` backup. Saved cameras,
-  observatories, and passband maps are now also written with an atomic
-  replace. [#662]
++ Saving settings can no longer destroy existing ones: all settings files
+  (including saved cameras, observatories, and passband maps) are written
+  via an atomic temporary-file replace, unreadable or conflicting files
+  are set aside as numbered ``.bak`` backups instead of being overwritten,
+  and a partial save no longer resurrects stale values from a conflicting
+  partial settings file. Encoding and OS errors while reading settings now
+  raise ``SettingsFileReadError``, a ``ValueError`` subclass. [#662]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -83,7 +83,11 @@ Bug Fixes
   full settings are safely on disk, and an unreadable settings file is
   set aside as a numbered ``.bak`` backup instead of being overwritten.
   Encoding and OS errors while reading settings now raise
-  ``SettingsFileReadError``, a ``ValueError`` subclass. [#662]
+  ``SettingsFileReadError``, a ``ValueError`` subclass. A partial save
+  no longer resurrects values from a conflicting partial settings file;
+  the conflicting file is set aside as a ``.bak`` backup. Saved cameras,
+  observatories, and passband maps are now also written with an atomic
+  replace. [#662]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,17 +78,12 @@ Bug Fixes
   saved (the selection observer previously rebound a local variable
   instead of setting the badge). [#661]
 + ``PhotometryWorkingDirSettings`` no longer destroys settings on failed
-  or conflicting writes. Settings are written to a temporary file that
-  atomically replaces the target, so an interrupted write cannot
-  truncate an existing settings file; the partial settings file is
-  disposed of only after new full settings are safely on disk; and a
-  settings file that exists but cannot be read is renamed to a ``.bak``
-  backup (numbered ``.bak1``, ``.bak2``, ... if needed) instead of being
-  silently overwritten or deleted. ``load()`` now parses both settings
-  files before raising so a save can merge into whichever file was
-  readable, and a file that is unreadable because of an encoding or
-  operating-system error is reported the same way as corrupt JSON
-  instead of raising an unexpected exception type. [#662]
+  writes. Settings are written to a temporary file that atomically
+  replaces the target, the partial settings file is removed only after
+  full settings are safely on disk, and an unreadable settings file is
+  set aside as a numbered ``.bak`` backup instead of being overwritten.
+  Encoding and OS errors while reading settings now raise
+  ``SettingsFileReadError``, a ``ValueError`` subclass. [#662]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/stellarphot/settings/settings_files.py
+++ b/stellarphot/settings/settings_files.py
@@ -291,6 +291,8 @@ class PhotometryWorkingDirSettings:
         )
         self._partial_settings = None
         self._settings = None
+        self._full_settings_unreadable = False
+        self._partial_settings_unreadable = False
 
     @property
     def settings(self):
@@ -403,15 +405,38 @@ class PhotometryWorkingDirSettings:
         Finally, if we are passed a partial setting and there is a full setting on disk,
         then the partial settings are merged with the full settings, and the full
         settings are saved.
+
+        An existing settings file that cannot be read is never overwritten in
+        place; it is renamed with a ``.bak`` suffix before the new settings
+        are written, and a save also sets aside, with the same ``.bak``
+        naming, any settings file its pre-save load found unreadable even
+        when the save does not rewrite that particular file. Existing backups
+        are never overwritten; if a ``.bak`` file already exists, numbered
+        suffixes (``.bak1``, ``.bak2``, ...) are used instead. The settings
+        are written to a temporary file that atomically replaces the target,
+        and the partial settings file is disposed of only after new full
+        settings are safely on disk, so a failed write cannot truncate or
+        destroy existing settings.
         """
         full_settings = False
 
         try:
             _ = self.load()
         except ValueError:
-            # If we catch a ValueError, then we are in a situation where we have no
-            # settings files. We can proceed to save the settings.
+            # load() raises ValueError when there are no settings files at
+            # all and when a file exists but cannot be read. The save
+            # proceeds in both cases; the unreadable-file case relies on the
+            # flags below so the problem file is set aside as .bak rather
+            # than destroyed.
             pass
+
+        # Whether each existing settings file failed to parse during the
+        # bookkeeping load above. load() latches these flags and parses both
+        # files before raising, so they are accurate even when only one of
+        # the two files is unreadable. They decide below whether a file
+        # about to be replaced is renamed aside instead of destroyed.
+        unreadable_full = self._full_settings_unreadable
+        unreadable_partial = self._partial_settings_unreadable
 
         match settings:
             case PartialPhotometrySettings():
@@ -442,7 +467,8 @@ class PhotometryWorkingDirSettings:
                     # If the settings file exists but could not be read then
                     # self._settings is None and there is nothing to merge the
                     # partial settings into. The partial settings are saved on
-                    # their own.
+                    # their own and the unreadable file is renamed with a .bak
+                    # suffix once the new settings are safely on disk.
 
                 # Are we updating or replacing partial settings?
                 if update and self._partial_settings is not None:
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -77,6 +77,18 @@ Bug Fixes
   tab whose setting is not saved on disk now actually marks the tab as not
   saved (the selection observer previously rebound a local variable
   instead of setting the badge). [#661]
++ ``PhotometryWorkingDirSettings`` no longer destroys settings on failed
```
mwcraig: shorten this

stellarphot/settings/settings_files.py
```diff
@@ -403,16 +405,39 @@ def save(self, settings, update=False):
...
-            # settings files. We can proceed to save the settings.
+            # load() raises ValueError when there are no settings files at
+            # all and when a file exists but cannot be read. The save
+            # proceeds in both cases; the unreadable-file case relies on the
+            # flags below so the problem file is set aside as .bak rather
+            # than destroyed.
             pass
 
+        # Whether each existing settings file failed to parse during the
+        # bookkeeping load above. load() latches these flags and parses both
```
mwcraig: What do you mean by "latches" here?

stellarphot/settings/settings_files.py
```diff
@@ -479,53 +505,138 @@ def save(self, settings, update=False):
...
+            if self._partial_settings_file.exists():
+                if unreadable_partial:
+                    self._move_aside(self._partial_settings_file)
+                else:
+                    self._partial_settings_file.unlink()
+            self._partial_settings = None
+
+    def _move_aside(self, path):
+        """
+        Rename ``path`` to a backup name that does not already exist and
```
mwcraig: Make this a numpy-style docstring

stellarphot/settings/settings_files.py
```diff
@@ -479,53 +505,138 @@ def save(self, settings, update=False):
...
-            f.write(settings.model_dump_json(indent=4))
+        # to the file. Write to a temporary file in the same directory and then
+        # atomically replace the target, so that a failure at any point during
+        # serialization or writing leaves any existing settings file untouched.
+        json_data = settings.model_dump_json(indent=4)
+        tmp_file = file.with_name(file.name + ".tmp")
+        try:
+            tmp_file.write_text(json_data, encoding=ENCODING)
+            if set_aside_target:
+                self._move_aside(file)
```
mwcraig: **Ordering bug (confirmed): the move-aside happens before the atomic replace, so a failed `replace()` leaves no settings file at all.**

When the existing `photometry_settings.json` is unreadable, `save()` renames it to `.bak` via `_move_aside(file)` and *then* runs `tmp_file.replace(file)`. If `replace()` raises (disk full, transient I/O error), the `finally:` block deletes the tmp file — leaving the target path with no file anywhere: only the `.bak` remains. That contradicts the guarantee stated in the comment above (a failed write can never destroy existing settings), and a subsequent `load()` reports the settings file doesn't exist.

None of the new tests cover this failure point — they mock `model_dump_json`/`write_text`, but not `Path.replace` after a successful move-aside.

A simple fix: **copy** the unreadable original to `.bak` instead of renaming it, so the original stays in place until `replace()` succeeds. (Alternatively, restore the `.bak` to its original name if `replace()` raises.)

---
*This comment was written by Claude at my direction.*

stellarphot/settings/settings_files.py
```diff
@@ -479,53 +505,138 @@ def save(self, settings, update=False):
...
-            with self._settings_file.open(encoding=ENCODING) as f:
-                content = f.read()
-
             try:
+                with self._settings_file.open(encoding=ENCODING) as f:
+                    content = f.read()
                 self._settings = PhotometrySettings.model_validate_json(content)
-            except ValidationError as e:
-                raise ValueError(f"Error loading settings: {e}") from e
+            except (ValidationError, OSError, UnicodeDecodeError) as e:
```
mwcraig: **Error masking (confirmed): folding `OSError`/`UnicodeDecodeError` into the generic `ValueError` makes the GUI silently swallow real read failures.**

`ReviewSettings` in `stellarphot/gui/custom_widgets.py` (lines ~770–772 and ~892–894) does:

```python
try:
    self._current_settings = PhotometryWorkingDirSettings().load()
except ValueError:
    self._current_settings = PartialPhotometrySettings()
```

Before this PR, a `PermissionError` or `UnicodeDecodeError` on a real, *valid* settings file propagated and failed loudly. Now the same condition becomes `ValueError` and the widget just shows a blank settings form as if nothing was ever saved — and if the user re-enters and saves, their original file survives only as a `.bak` the GUI never mentions.

Consider either a distinct exception type for I/O-level failures (so the GUI can distinguish "no/invalid settings" from "couldn't read your settings"), or surfacing the error text in the GUI fallback path.

---
*This comment was written by Claude at my direction.*

stellarphot/settings/settings_files.py
```diff
@@ -479,53 +505,138 @@ def save(self, settings, update=False):
...
-            with self._partial_settings_file.open(encoding=ENCODING) as f:
-                content = f.read()
+        errors = []
+        last_exc = None
 
+        # Load PartialPhotometrySettings first, if it exists. A read failure
+        # (ValidationError, or an OSError/UnicodeDecodeError while opening or
+        # decoding the file) leaves self._partial_settings as None, exactly
+        # like the missing-file case.
+        if self._partial_settings_file.exists():
```
mwcraig: **Duplication: this partial-file `try/except` block and the full-file one below (lines ~624–632) are near-identical ~10-line copies.**

They differ only in which file, model class, `*_is_unreadable` flag, and error label they use. This PR already had to widen the `except` tuple in both copies — a future maintainer touching only one copy silently regresses the other file type to letting raw exceptions escape.

Suggest a small private helper, e.g. `_try_load(path, model_cls) -> (model | None, error_msg | None)`, called once per file.

---
*This comment was written by Claude at my direction.*

stellarphot/settings/tests/test_settings_file.py
```diff
@@ -484,7 +488,264 @@ def test_save_partial_update_with_unreadable_full_settings(self):
 
         assert settings_file.partial_settings_file.exists()
         assert settings_file.partial_settings == partial_settings
+        assert not settings_file.settings_file.exists()
+        backup_file = settings_file.settings_file.with_name(
```
mwcraig: **Test duplication: the `X.with_name(X.name + ".bak")` expression is repeated verbatim in roughly ten of the new tests** (≈ lines 492, 511, 530, 549, 581, 627, 645, 659, 706, 724, 745).

A one-line helper (or fixture) would remove all of them, e.g.:

```python
def bak_path(path, suffix=".bak"):
    return path.with_name(path.name + suffix)
```

Beyond brevity, this keeps the tests honest: if the backup naming convention ever changes, a single missed inline copy would leave that test asserting against a stale path and silently no longer verifying the backup behavior.

---
*This comment was written by Claude at my direction.*

stellarphot/settings/settings_files.py
```diff
@@ -475,53 +505,138 @@ def save(self, settings, update=False):
...
         # Write the settings to a file. The settings themselves are models, so we
         # are guaranteed to write the correct model type (partial or full settings)
-        # to the file.
-        with file.open("w", encoding=ENCODING) as f:
-            f.write(settings.model_dump_json(indent=4))
+        # to the file. Write to a temporary file in the same directory and then
+        # atomically replace the target, so that a failure at any point during
+        # serialization or writing leaves any existing settings file untouched.
+        json_data = settings.model_dump_json(indent=4)
+        tmp_file = file.with_name(file.name + ".tmp")
```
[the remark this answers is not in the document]
mwcraig: The observation is technically correct — the temporary name is fixed per target file — but concurrent saves to the same working directory are outside this code's concurrency model. `save()` is synchronous, single-user, widget-driven code; there is no supported scenario in which two saves overlap. The same tradeoff is already made explicitly (with a comment) in `_move_aside`, which accepts a narrow `exists()`/`replace()` race for the backup name.

Worth noting that a unique temporary name would not actually make overlapping saves safe: two concurrent saves would still race at `tmp.replace(target)` (last writer silently wins) and at `_move_aside`, so `mkstemp` removes only the `FileNotFoundError` symptom, not the underlying race. Since both writers hold fully serialized, valid JSON, the worst outcome with the current scheme is the same last-writer-wins result.

That said, switching to `tempfile.mkstemp(dir=file.parent, prefix=file.name, suffix=".tmp")` is cheap hardening and removes the crash mode — happy to make that change if you'd like it in this PR.

CHANGES.rst
mwcraig: Agreed — it currently runs 13 lines. Proposed replacement (6 lines):

```rst
+ ``PhotometryWorkingDirSettings`` no longer destroys settings on failed
  writes. Settings are written to a temporary file that atomically
  replaces the target, the partial settings file is removed only after
  full settings are safely on disk, and an unreadable settings file is
  set aside as a numbered ``.bak`` backup instead of being overwritten.
  Encoding and OS errors while reading settings now raise ``ValueError``
  like corrupt JSON. [#662]
```

stellarphot/settings/settings_files.py
mwcraig: "Latches" was meant in the electronics sense: `load()` sets `_full_settings_unreadable` / `_partial_settings_unreadable` and they *stay set on the instance even though `load()` then raises* — so the state survives the exception for `save()` to consult afterward. That persistence-past-the-raise is the point being emphasized, since a reader might otherwise assume the flags are lost when the load fails.

Agreed it's jargon. Proposed rewording:

```python
# Whether each existing settings file failed to parse during the
# bookkeeping load above. load() records these flags on the instance
# before raising, and it parses both files before raising, so the
# flags are accurate even when only one of the two files is
# unreadable. They decide below whether a file about to be replaced
# is renamed aside instead of destroyed.
```

stellarphot/settings/settings_files.py
mwcraig: Agreed. Proposed:

```python
def _move_aside(self, path):
    """
    Rename a file to a backup name that does not already exist.

    The first available of ``<name>.bak``, ``<name>.bak1``,
    ``<name>.bak2``, ... is used, so an earlier backup -- which may
    hold the only copy of settings from a previous corruption -- is
    never overwritten.

    Parameters
    ----------
    path : `pathlib.Path`
        The file to rename.

    Returns
    -------
    `pathlib.Path`
        The path the file was renamed to.
    """
```

stellarphot/settings/settings_files.py
mwcraig: Verified against the branch — this is real. The sequence at `settings_files.py:525-533` is `write_text(tmp)` → `_move_aside(file)` → `tmp.replace(file)`, and the `finally` unlinks the tmp file. So if `replace()` raises after a successful move-aside, the canonical path ends up with no file: the contents survive only in the `.bak`, but `load()` then reports the settings file doesn't exist, which contradicts the comment at lines 508-513 ("leaves the unreadable file in place under its original name instead of leaving no settings file at all").

Also confirmed the test-coverage gap: the failure-injection tests mock `model_dump_json` (line ~684) and `Path.write_text` (lines ~701, ~719), but nothing mocks `Path.replace`, so this window is untested.

Two mitigating notes on severity: `replace()` is a same-directory rename, so its realistic failure modes are much rarer than a data write's, and no bytes are ever lost — recovery is renaming the `.bak` back. Still worth fixing since the whole point of this PR is that a failed save never leaves the directory in a worse state.

Of the two suggested fixes I'd take the **copy** variant: copy the unreadable original to the reserved backup name (`open(backup, "x")` semantics preserve the never-overwrite guarantee), leave the original in place, then `replace()`. If the copy fails the save aborts with the original untouched; if `replace()` fails the original is still at its canonical path. Restore-on-failure works too but adds a second rename that can itself fail. Plus a test that patches `Path.replace` to raise after the move-aside path is taken.

Happy to implement this on the branch.

stellarphot/settings/settings_files.py
mwcraig: Confirmed: both `ReviewSettings` call sites (`custom_widgets.py:769-772` and `891-894`) catch `ValueError` and silently fall back to empty `PartialPhotometrySettings()`, and before this PR the file read happened outside the `try`, so a `PermissionError` on a perfectly valid file propagated instead of presenting a blank form.

Two mitigating points, then a proposal. First, the widened `except` is what makes the `.bak` set-aside reachable for OS-level failures at all — with the old behavior, a transiently unreadable file crashed the widget outright, which is arguably worse UX than the blank form, and either way the file itself is never destroyed. Second, surfacing load errors in the GUI is exactly the banner UX from #657 that this PR's description deliberately defers.

Proposal for this PR: introduce `class SettingsFileReadError(ValueError)` and raise it for the `OSError`/`UnicodeDecodeError` cases (keeping plain `ValueError` for corrupt/conflicting JSON). Because it subclasses `ValueError`, nothing changes behaviorally now, but the follow-up banner PR can distinguish "you have no settings" from "your settings exist but could not be read" without touching this module again. The alternative — making the GUI display error text now — would pull the #657 UX scope into this PR.

Happy to implement the exception subclass on the branch if that split sounds right.

stellarphot/settings/settings_files.py
mwcraig: Agreed — the two blocks differ only in path, model class, flag, and label, and the widened `except` tuple did have to be edited in both. A helper along these lines removes the drift risk:

```python
def _try_load(self, path, model_cls, label):
    """Parse one settings file; return (model | None, error | None)."""
    if not path.exists():
        return None, None
    try:
        with path.open(encoding=ENCODING) as f:
            return model_cls.model_validate_json(f.read()), None
    except (ValidationError, OSError, UnicodeDecodeError) as e:
        return None, (f"Error loading {label}: {e}", e)
```

with `load()` calling it once per file and setting the `*_unreadable` flag when the error slot is non-None. Happy to implement on the branch.

stellarphot/settings/tests/test_settings_file.py
mwcraig: Agreed — the expression appears verbatim at all the cited lines. A module-level helper in the test file:

```python
def bak_path(path, suffix=".bak"):
    return path.with_name(path.name + suffix)
```

covers both the plain `.bak` and the numbered `.bak1`/`.bak2` assertions (`bak_path(f, ".bak1")`), and keeps every test pointing at the one definition of the naming convention. Happy to implement on the branch.

stellarphot/settings/settings_files.py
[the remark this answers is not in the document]
mwcraig: Do it

stellarphot/settings/tests/test_settings_file.py
```diff
@@ -484,7 +495,298 @@ def test_save_partial_update_with_unreadable_full_settings(self):
 
         assert settings_file.partial_settings_file.exists()
         assert settings_file.partial_settings == partial_settings
+        assert not settings_file.settings_file.exists()
+        backup_file = bak_path(settings_file.settings_file)
+        assert backup_file.read_text() == bad_content
+
+    def test_save_partial_with_unreadable_partial_settings_makes_backup(self):
```
mwcraig: Can this and the next few tests be factored out into a single parametrized test?

stellarphot/settings/settings_files.py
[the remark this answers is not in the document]
mwcraig: Done in 5a991a3 — `_atomic_write_json` now uses `tempfile.mkstemp(dir=file.parent, prefix=file.name + ".", suffix=".tmp")`, so each save gets a unique temporary name and only cleans up its own file.

stellarphot/settings/tests/test_settings_file.py
mwcraig: Yes — this test plus the two below it (`test_save_full_with_unreadable_full_settings_makes_backup`, `test_save_full_with_unreadable_partial_settings_makes_backup`) and the one above (`test_save_partial_update_with_unreadable_full_settings`) all share the same skeleton: corrupt one file, save something, assert which files exist afterward and that the corrupt bytes survived as `.bak`. They collapse into one parametrized test over (which file is corrupt, what gets saved):

```python
@pytest.mark.parametrize(
    "corrupt, saved, expect_full",
    [
        # unreadable full + incomplete partial save: partial written, full set aside
        ("full", "partial", False),
        # unreadable partial + incomplete partial save: new partial written
        ("partial", "partial", False),
        # unreadable full + complete partial save: full written
        ("full", "complete", True),
        # unreadable partial + full save: full written, partial set aside
        ("partial", "full", True),
    ],
)
def test_save_with_unreadable_settings_file_makes_backup(
    self, corrupt, saved, expect_full
):
    settings_file = PhotometryWorkingDirSettings()
    bad_content = '{"pasta": "carbonara"}'
    corrupt_path = (
        settings_file.settings_file
        if corrupt == "full"
        else settings_file.partial_settings_file
    )
    corrupt_path.write_text(bad_content)

    if saved == "full":
        to_save = PhotometrySettings(**TEST_PHOTOMETRY_SETTINGS)
        settings_file.save(to_save)
    else:
        to_save = (
            PartialPhotometrySettings(**TEST_PHOTOMETRY_SETTINGS)
            if saved == "complete"
            else PartialPhotometrySettings(camera=Camera.model_validate_json(CAMERA))
        )
        settings_file.save(to_save, update=True)

    assert settings_file.settings_file.exists() == expect_full
    assert settings_file.partial_settings_file.exists() != expect_full
    if expect_full:
        assert settings_file.settings == to_save
    else:
        assert settings_file.partial_settings == to_save
    assert bak_path(corrupt_path).read_text() == bad_content
```

The one per-case assertion that doesn't survive verbatim is the cross-check in the current second test (`settings == settings` for a complete partial); it's covered by the `expect_full` branch above. The per-test explanatory comments would move into the param list comments. Happy to apply this on the branch.

mwcraig: Done in b586ffd — the four tests are now one parametrized `test_save_with_unreadable_settings_file_makes_backup` with the four (corrupt file, saved settings) cases.

CHANGES.rst
mwcraig: Shortened in d362001 — the entry (which had grown back to 12 lines after the later fixes were folded in) is now 7 lines covering all the changes.

PR #663, opened 2026-08-08
Show settings-migration warnings in a banner in ReviewSettings

mwcraig:
Extracted from #657 as part of decomposing that PR into small, reviewable pieces. This is the piece #657 was originally opened for: a banner in `ReviewSettings` for warning the user that their settings file was updated from an earlier version.

- Adds a `PhotometrySettingsWarning` category (exported from `stellarphot.settings`) for user-actionable problems with saved photometry settings files.
- `ReviewSettings` now captures warnings of that category raised while loading the working-directory settings and displays them in a banner above the settings, instead of letting them disappear into the notebook log. Other warning categories stay out of the banner, and warning text is HTML-escaped before display.

**This PR is a display channel only** — nothing emits `PhotometrySettingsWarning` yet. The settings-format version field and load-time migration that will produce the warning are being added in #656; once both are in, opening `ReviewSettings` on a directory with an old-format settings file will show the migration notice in the banner.

Independent of #661 and #662 (touches different code paths); only the changelog will need a trivial conflict resolution between these PRs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,14 +47,12 @@ Other Changes and Additions
   assertions to the settings JSON/table round-trip tests; and pinned the
   ``klims`` workaround for pytransit's ``RoadRunnerModel`` with a regression
   test at the maximum allowed planet radius. [#653]
-+ Added a ``stellarphot.settings.PhotometrySettingsWarning`` category for
-  user-actionable problems with saved photometry settings files.
-  ``ReviewSettings`` now shows warnings of that category raised while
-  loading the working-directory settings in a banner above the settings,
-  instead of letting them disappear into the notebook log; other warning
-  categories stay out of the banner. Nothing emits the warning yet -- the
-  settings-format version field and load-time migration that will produce
-  it are being added separately (#656). [#663]
++ ``ReviewSettings`` now displays warnings of the new
+  ``stellarphot.settings.PhotometrySettingsWarning`` category, raised while
+  loading the working-directory settings, in a dismissible banner above the
+  settings instead of letting them disappear into the notebook log. Nothing
+  emits the warning yet; the settings-format migration that will is being
+  added in #656. [#663]
 
 Bug Fixes
 ^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,14 @@ Other Changes and Additions
   assertions to the settings JSON/table round-trip tests; and pinned the
   ``klims`` workaround for pytransit's ``RoadRunnerModel`` with a regression
   test at the maximum allowed planet radius. [#653]
++ Added a ``stellarphot.settings.PhotometrySettingsWarning`` category for
+  user-actionable problems with saved photometry settings files.
+  ``ReviewSettings`` now shows warnings of that category raised while
+  loading the working-directory settings in a banner above the settings,
+  instead of letting them disappear into the notebook log; other warning
+  categories stay out of the banner. Nothing emits the warning yet -- the
+  settings-format version field and load-time migration that will produce
+  it are being added separately (#656). [#663]
 
 Bug Fixes
 ^^^^^^^^^
--- a/stellarphot/gui/custom_widgets.py
+++ b/stellarphot/gui/custom_widgets.py
@@ -32,6 +32,12 @@ __all__ = ["ChooseOrMakeNew", "Confirm", "SettingWithTitle"]
 
 DEFAULT_BUTTON_WIDTH = "300px"
 
+# Colors for the settings-warning banner in ReviewSettings; the dismiss
+# button is styled to match the banner border.
+_BANNER_BORDER_COLOR = "#ffc107"
+_BANNER_TEXT_COLOR = "#664d03"
+_BANNER_BACKGROUND_COLOR = "#fff3cd"
+
 
 class ChooseOrMakeNew(ipw.VBox):
     """
@@ -768,19 +774,36 @@ class ReviewSettings(ipw.VBox):
     `~stellarphot.settings.PhotometrySettingsWarning` category (e.g. a
     settings-format migration), the warning is displayed in a banner above
     the settings instead of being lost in the notebook log. Other warning
-    categories stay out of the banner.
+    categories stay out of the banner. The banner stays up until the user
+    dismisses it, even though later reloads of the settings -- which happen
+    on every tab selection -- may no longer raise the warning.
     """
 
     def __init__(self, settings, style="tabs", *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         # Banner for messages about the saved settings. It is hidden unless
-        # there is a message to show.
-        self._banner = ipw.HTML()
-        self._banner.layout.display = "none"
-        self._banner_messages = []
+        # there is a message to show, and once shown it stays up until the
+        # user dismisses it.
+        self._banner_html = ipw.HTML()
+        self._banner_dismiss = ipw.Button(
+            description="✕",
+            tooltip="Dismiss this message",
+            layout=ipw.Layout(width="2.5em"),
+        )
+        self._banner_dismiss.style.button_color = _BANNER_BORDER_COLOR
+        self._banner_dismiss.style.text_color = _BANNER_TEXT_COLOR
+        self._banner = ipw.HBox([self._banner_html, self._banner_dismiss])
 
-        # Get a copy of whatever settings may have already been saved.
-        self._current_settings = self._load_working_dir_settings()
+        def dismiss_banner(_):
+            self._banner.layout.display = "none"
+
+        self._banner_dismiss.on_click(dismiss_banner)
+
+        # Get a copy of whatever settings may have already been saved,
+        # showing any warnings the load raised in the banner. This is the
+        # only place the banner is set: reloads by the current_settings
+        # property must not clear a notice the user has not yet dismissed.
+        self._update_banner(self._load_working_dir_settings())
 
         self._setting_widgets = []
         self._plain_names = []
@@ -899,16 +922,16 @@ class ReviewSettings(ipw.VBox):
         """
         The current settings in the widget.
         """
-        self._current_settings = self._load_working_dir_settings()
+        self._load_working_dir_settings()
         return self._current_settings
 
     def _load_working_dir_settings(self):
         """
-        Load settings from the working directory, routing any
-        `~stellarphot.settings.PhotometrySettingsWarning` the load generates
-        into the banner instead of letting it be lost in the notebook log.
+        Load settings from the working directory into
+        ``self._current_settings`` and return the messages of any
+        `~stellarphot.settings.PhotometrySettingsWarning` the load
+        generated, for display in the banner instead of the notebook log.
         """
-        recorded = []
         try:
             # Only PhotometrySettingsWarning -- the category for
             # user-actionable problems with the saved settings, such as a
@@ -922,29 +945,30 @@ class ReviewSettings(ipw.VBox):
                 loaded = PhotometryWorkingDirSettings().load()
         except ValueError:
             # No readable settings in the working directory; start from
-            # defaults, keeping any warnings recorded before the error.
+            # defaults, keeping any warnings recorded before the error --
+            # recorded is bound by the with statement before anything in
+            # its body can raise.
             loaded = PartialPhotometrySettings()
 
-        self._banner_messages = [str(warning.message) for warning in recorded]
-        self._update_banner()
-        return loaded
+        self._current_settings = loaded
+        return [str(warning.message) for warning in recorded]
 
-    def _update_banner(self):
+    def _update_banner(self, messages):
         """
         Show the banner if there are messages, hide it if there are none.
         """
-        if self._banner_messages:
-            content = "".join(
-                f"<p>⚠️ {html.escape(msg)}</p>" for msg in self._banner_messages
-            )
-            self._banner.value = (
-                "<div style='border: 2px solid #ffc107; border-radius: 4px; "
-                "background-color: #fff3cd; color: #664d03; "
+        if messages:
+            content = "".join(f"<p>⚠️ {html.escape(msg)}</p>" for msg in messages)
+            self._banner_html.value = (
+                f"<div style='border: 2px solid {_BANNER_BORDER_COLOR}; "
+                f"border-radius: 4px; "
+                f"background-color: {_BANNER_BACKGROUND_COLOR}; "
+                f"color: {_BANNER_TEXT_COLOR}; "
                 f"padding: 0.25em 1em;'>{content}</div>"
             )
             self._banner.layout.display = "flex"
         else:
-            self._banner.value = ""
+            self._banner_html.value = ""
             self._banner.layout.display = "none"
 
     def _observe_tab_selection(self, change):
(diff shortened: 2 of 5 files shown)


mwcraig (commented):
Is there a way for the user to dismiss the banner after it displays?

CHANGES.rst
```diff
@@ -47,6 +47,14 @@ Other Changes and Additions
   assertions to the settings JSON/table round-trip tests; and pinned the
   ``klims`` workaround for pytransit's ``RoadRunnerModel`` with a regression
   test at the maximum allowed planet radius. [#653]
++ Added a ``stellarphot.settings.PhotometrySettingsWarning`` category for
```
mwcraig: Shorten this up

stellarphot/gui/custom_widgets.py
```diff
@@ -888,12 +899,53 @@ def current_settings(self):
...
+        recorded = []
         try:
-            self._current_settings = PhotometryWorkingDirSettings().load()
+            # Only PhotometrySettingsWarning -- the category for
+            # user-actionable problems with the saved settings, such as a
+            # settings-format migration -- belongs in the banner. Warnings
+            # of other categories raised by libraries on the load path are
+            # not aimed at the user of this widget, so they are suppressed
+            # here rather than displayed.
+            with warnings.catch_warnings(record=True) as recorded:
```
mwcraig, line 942: Just to make sure I understand this, this completely drops all other warnings, right?

stellarphot/gui/custom_widgets.py
```diff
@@ -888,12 +899,53 @@ def current_settings(self):
         """
         The current settings in the widget.
         """
+        self._current_settings = self._load_working_dir_settings()
```
mwcraig: Two things about the property:

1. **It now has UI side effects, triggered on every tab click.** `_observe_tab_selection` reads `self.current_settings`, so each tab click re-reads disk and rewrites the banner. While the on-disk file is old-format that's fine (the migration warning re-fires each load, so the banner persists), but the moment the file is re-saved in the new format, the next tab click silently clears the banner. That may be exactly the desired behavior — the message describes a condition that no longer exists — but a user who half-read the notice loses it on their first click. If the banner should be sticky, only update it from `__init__`.

2. **Minor: let the helper own the assignment.** Both `__init__` and this property do `self._current_settings = self._load_working_dir_settings()`; if the helper assigns `self._current_settings` itself and returns it, the property body collapses to one line.

stellarphot/gui/custom_widgets.py
```diff
@@ -761,15 +764,23 @@ class ReviewSettings(ipw.VBox):
...
+    categories stay out of the banner.
     """
 
     def __init__(self, settings, style="tabs", *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
+        # Banner for messages about the saved settings. It is hidden unless
+        # there is a message to show.
+        self._banner = ipw.HTML()
+        self._banner.layout.display = "none"
+        self._banner_messages = []
```
mwcraig: These two lines are dead: both are unconditionally overwritten a few lines later by `_load_working_dir_settings()` → `_update_banner()`. Only `self._banner = ipw.HTML()` needs to precede the load.

stellarphot/gui/custom_widgets.py
```diff
@@ -888,12 +899,53 @@ def current_settings(self):
...
+        self._current_settings = self._load_working_dir_settings()
+        return self._current_settings
+
+    def _load_working_dir_settings(self):
+        """
+        Load settings from the working directory, routing any
+        `~stellarphot.settings.PhotometrySettingsWarning` the load generates
+        into the banner instead of letting it be lost in the notebook log.
+        """
+        recorded = []
```
mwcraig: `recorded = []` is dead code — the `as recorded` binding happens at `__enter__`, before anything in the body can raise, so `recorded` is always bound in the `except` branch. (Verified empirically that the pre-exception warnings survive.)

stellarphot/gui/custom_widgets.py
```diff
@@ -888,12 +899,53 @@ def current_settings(self):
...
+            # user-actionable problems with the saved settings, such as a
+            # settings-format migration -- belongs in the banner. Warnings
+            # of other categories raised by libraries on the load path are
+            # not aimed at the user of this widget, so they are suppressed
+            # here rather than displayed.
+            with warnings.catch_warnings(record=True) as recorded:
+                warnings.simplefilter("ignore")
+                warnings.simplefilter("always", PhotometrySettingsWarning)
+                loaded = PhotometryWorkingDirSettings().load()
         except ValueError:
```
mwcraig, line 946: **Test gap:** this "warnings recorded before the load raises `ValueError` still reach the banner" path has an explanatory comment but no test — and it's the path #656's migrate-then-still-fail scenario will hit. The behavior is correct (verified on the branch), but a test would keep it that way.

stellarphot/gui/custom_widgets.py
```diff
@@ -888,12 +899,53 @@ def current_settings(self):
...
-            self._current_settings = PartialPhotometrySettings()
+            # No readable settings in the working directory; start from
+            # defaults, keeping any warnings recorded before the error.
+            loaded = PartialPhotometrySettings()
 
-        return self._current_settings
+        self._banner_messages = [str(warning.message) for warning in recorded]
```
mwcraig: `_banner_messages` doesn't need to be instance state — it's written in exactly one place and read in exactly one place. Passing the list as a parameter (`_update_banner(messages)`) deletes the attribute and the write-then-read coupling between the two methods. Nothing in the tests touches it.

stellarphot/gui/tests/test_custom_widgets.py
```diff
@@ -1254,6 +1256,75 @@ def test_selecting_table_without_saved_setting_sets_proper_badge(self):
...
+        # loading settings (e.g. a settings-format migration) should be
+        # displayed in the banner.
+        def fake_load(_self):
+            warnings.warn(
+                "Settings were migrated", PhotometrySettingsWarning, stacklevel=2
+            )
+            return PartialPhotometrySettings()
+
+        mocker.patch.object(
+            settings_files.PhotometryWorkingDirSettings, "load", fake_load
```
mwcraig: This `fake_load` + `mocker.patch.object` block is repeated nearly verbatim in the three warning tests, differing only in message/category. A small helper on the test class (`_patch_load_to_warn(mocker, message, category)`) removes two of the three copies — and gives the missing warn-then-raise test a natural home.

mwcraig (commented):

Critical review focused on bugs and simplification. The merge-base diff against `main` contains only the two banner commits, so none of the #661/#662 material contaminates the review surface. The tricky parts were verified empirically on the branch: all 6 new tests pass, and warnings recorded before a `ValueError` during load really do survive into the banner as the code comment claims.

**No real bugs found.** The filter ordering is correct (`simplefilter("always", PhotometrySettingsWarning)` is inserted ahead of the blanket `"ignore"`, so only the wanted category is recorded), the HTML escaping works, `stellarphot.settings` re-exports the new warning class via `from .settings_files import *`, and the tests' patch of `settings_files.PhotometryWorkingDirSettings.load` hits the same class object the widget uses.

On dedup more broadly: the banner doesn't duplicate an existing pattern elsewhere in the codebase (the existing `catch_warnings` uses in `photometry/` and `fits_opener.py` are suppression-oriented and not worth unifying with this). The main structural choice in the PR — extracting `_load_working_dir_settings` so `__init__` and the property share one load path — is the right dedup move and replaces what used to be a copy-pasted try/except.

Details are in the inline comments: one behavioral judgment call (banner clearing on tab click), one test gap, and four small simplifications.

stellarphot/gui/custom_widgets.py
```diff
@@ -888,12 +899,53 @@ def current_settings(self):
...
+            # Only PhotometrySettingsWarning -- the category for
+            # user-actionable problems with the saved settings, such as a
+            # settings-format migration -- belongs in the banner. Warnings
+            # of other categories raised by libraries on the load path are
+            # not aimed at the user of this widget, so they are suppressed
+            # here rather than displayed.
+            with warnings.catch_warnings(record=True) as recorded:
+                warnings.simplefilter("ignore")
+                warnings.simplefilter("always", PhotometrySettingsWarning)
+                loaded = PhotometryWorkingDirSettings().load()
```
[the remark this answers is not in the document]
mwcraig: Confirmed — both premises check out, and the problem is slightly worse than stated.

- Every `save()` starts with a bookkeeping `_ = self.load()` (`settings_files.py:584`).
- `ReviewSettings.__init__` triggers saves during construction: the autosave of default-constructible settings (`custom_widgets.py:865`), **and** the already-saved `ChooseOrMakeNew` path — setting `value = None` then `saved_value` (lines 816–817) fires the `_choose_existing` observer → `save_wd` → `save(update=True)`. So construction saves in essentially every case, not just when defaults are autosaved.

Once #656's load-time migration warning exists, the sequence on an old-format file is: initial load (banner captures the warning) → widget values set → `save()` → bookkeeping `load()` of the still-old-format file **outside** the capture context → warning escapes to the notebook log → the save then rewrites the file in the new format. And because `catch_warnings` restores the warning registry on exit, the banner's capture doesn't consume the "show once" slot — the leaked warning will actually print.

Side effect worth noting: since construction itself rewrites the file in migrated format, the next `current_settings` read (any tab click) reloads cleanly and clears the banner — which feeds the stickiness discussion in the thread at line 902.

Agreed on the fix: suppress `PhotometrySettingsWarning` around the widget-triggered saves only (e.g. a small `catch_warnings` + `simplefilter("ignore", PhotometrySettingsWarning)` inside `save_wd` in `_add_saving_to_widget`), leaving ordinary callers' filters intact and keeping the banner as the single display channel. Happy to implement that on this branch, or fold it into the #656 integration since nothing emits the warning until then.

CHANGES.rst
mwcraig: Proposed shorter version (8 lines → 5, drops the category-definition preamble and the "other categories" detail, which the docstring already covers):

```rst
+ ``ReviewSettings`` now displays warnings of the new
  ``stellarphot.settings.PhotometrySettingsWarning`` category, raised while
  loading the working-directory settings, in a banner above the settings
  instead of letting them disappear into the notebook log. Nothing emits
  the warning yet; the settings-format migration that will is being added
  in #656. [#663]
```

stellarphot/gui/custom_widgets.py
mwcraig: Yes — within this one `load()` call, every warning that is not a `PhotometrySettingsWarning` is dropped entirely: `simplefilter("ignore")` means it is neither recorded nor printed, and nothing re-emits it after the block. `PhotometrySettingsWarning` matches the more specific `simplefilter("always", ...)` entry (filters are consulted newest-first), so it is recorded for the banner (and not printed). `catch_warnings` restores the previous filters and the warning registry on exit, so nothing outside this call site is affected.

If you'd rather not lose the other categories, the alternative is to record everything (`simplefilter("always")` alone), route `PhotometrySettingsWarning` to the banner, and re-emit the rest with `warnings.warn_explicit(...)` after the block so they follow the caller's normal filters. I suppressed them on the theory that library noise during a GUI load isn't aimed at the widget user, but it's a small change if you want the pass-through behavior.

stellarphot/gui/custom_widgets.py
mwcraig: On 1 — agreed, and having now traced the save path, it's worse than "the moment the file is re-saved": construction itself re-saves. The already-saved `ChooseOrMakeNew` path (`value = None` then `saved_value`, lines 816–817) fires the save observer, and `save()` rewrites the file — in migrated format once #656 lands. So the *first* tab click after opening the widget reloads a clean file and clears the banner; the user gets essentially one glance at it.

Recommendation: make the banner sticky — set it only from `__init__`, and drop the reload's banner update (the property keeps refreshing `_current_settings`; it just stops touching the banner). Combined with your review-level question about dismissal, I'd add an explicit ✕ button (turn `_banner` into an `HBox` of the HTML plus a small `Button` that hides it). That gives: banner appears when the widget opens on an old-format file, stays until the user dismisses it, never silently vanishes on a tab click.

On 2 — agreed; if `_load_working_dir_settings` assigns `self._current_settings` itself and returns it, the property collapses to `return self._load_working_dir_settings()` and `__init__` just calls it. Will fold both into the same change if you want the sticky-banner version.

stellarphot/gui/custom_widgets.py
mwcraig: Correct — `_load_working_dir_settings()` unconditionally assigns `_banner_messages` and calls `_update_banner()`, which sets `layout.display` on every path (to `"none"` when there are no messages), so both initializations at lines 779–780 are overwritten before anything can observe them. Only `self._banner = ipw.HTML()` needs to precede the load. Will remove.

(If the banner goes sticky per the line-902 thread and `_banner_messages` becomes a parameter per the line-928 thread, both lines disappear as part of that restructuring anyway.)

stellarphot/gui/custom_widgets.py
mwcraig: Correct — the `with` statement's `__enter__` binds `recorded` before any code that can raise `ValueError` runs (the `PhotometryWorkingDirSettings()` constructor and `load()` are both inside the body), so `recorded` is always bound when the `except` branch reads it. The `recorded = []` at line 911 is dead. Will remove and let the code lean on the binding, with the warn-then-raise test from the line-923 thread pinning the behavior.

stellarphot/gui/custom_widgets.py
mwcraig: Agreed. A test where the patched `load()` warns with `PhotometrySettingsWarning` and then raises `ValueError`, asserting the message still lands in the banner, covers exactly the migrate-then-still-fail path #656 will exercise — and it's the only thing keeping the `recorded`-binding subtlety (line-911 thread) honest once the dead initialization is removed. Will add it, using the shared `fake_load` helper proposed in the test-file thread.

stellarphot/gui/custom_widgets.py
mwcraig: Agreed — single write, single read, nothing else touches it, so `_update_banner(messages)` is strictly better than the write-then-read attribute coupling. This also composes cleanly with the sticky-banner option from the line-902 thread: `__init__` calls `_update_banner(messages)` once and the reload path simply never calls it. Will apply.

stellarphot/gui/tests/test_custom_widgets.py
mwcraig: Agreed — the three blocks differ only in message/category, and a `_patch_load_to_warn(mocker, message, category, raise_after=False)` helper collapses them; the `raise_after` flag gives the missing warn-then-raise test (line-923 thread) its natural home. Will apply together with that new test.

stellarphot/gui/custom_widgets.py
mwcraig: Keep the current behavior.

stellarphot/gui/custom_widgets.py
mwcraig: Add the button; style its color to match the warning box border. Sticky banner is better.

stellarphot/gui/tests/test_custom_widgets.py
```diff
@@ -1320,43 +1330,114 @@ def test_banner_ignores_other_warning_categories(self, mocker, category):
...
+            settings_files.PhotometryWorkingDirSettings,
+            "load",
+            lambda _self: PartialPhotometrySettings(),
+        )
+        review_settings._container.selected_index = 1
+
+        assert "Settings were migrated" in review_settings._banner_html.value
+        assert review_settings._banner.layout.display == "flex"
+
+    def test_banner_dismiss_button_hides_banner(self, mocker):
```
mwcraig, line 1388: Add a descriptive comment

mwcraig: Added in 4c78e73 — the comment notes the banner is sticky, so the dismiss button is the one way to hide it.

PR #671, opened 2026-08-10
Relax vsx apass bounds

mwcraig:
Fixes #668

### DIFF
--- a/stellarphot/utils/tests/test_magnitude_transforms.py
+++ b/stellarphot/utils/tests/test_magnitude_transforms.py
@@ -17,27 +17,84 @@ from ..magnitude_transforms import (
 )
 
 
-def generate_input_mags(n_stars):
-    # Generate n_stars with magnitude in range 10 to 15
+def _generate_input_mags(n_stars):
+    """
+    Generate instrumental magnitudes in the range 10 to 15.
+
+    The random number generator is seeded, so the magnitudes are the same
+    from one run to the next.
+
+    Parameters
+    ----------
+
+    n_stars : int
+        Number of magnitudes to generate.
+
+    Returns
+    -------
+    `astropy.table.Column`
+        Column of instrumental magnitudes named ``instrumental``.
+    """
     rg = np.random.default_rng(1024)
     input_mags = rg.integers(0, high=50, size=n_stars) / 10 + 10
     instr_mags = Column(name="instrumental", data=input_mags)
     return instr_mags
 
 
-def generate_catalog_mags(instrument_mags, color, model):
+def _generate_catalog_mags(instrument_mags, color, model):
     """
-    Generate catalog magnitudes from instrumental magnitudes
-    given a model that relates the two.
+    Generate catalog magnitudes from instrumental magnitudes.
+
+    Parameters
+    ----------
+
+    instrument_mags : `astropy.table.Column`
+        Instrumental magnitudes.
+
+    color : `astropy.table.Column`
+        Color of each star.
+
+    model : `astropy.modeling.Model`
+        Model relating the color to the difference between the catalog and
+        instrumental magnitudes.
+
+    Returns
+    -------
+    `astropy.table.Column`
+        Catalog magnitudes.
     """
     return instrument_mags + model(color)
 
 
-def generate_star_coordinates(
+def _generate_star_coordinates(
     n_stars, ra_start=180 * u.degree, dec_start=45 * u.degree, separation=10 * u.arcsec
 ):
     """
-    Generate RA/Dec coordinates for a set of stars.
+    Generate RA/Dec coordinates for a set of stars on a square grid.
+
+    Parameters
+    ----------
+
+    n_stars : int
+        Number of coordinates to generate.
+
+    ra_start : `astropy.units.Quantity`, optional
+        Right ascension of the first star.
+
+    dec_start : `astropy.units.Quantity`, optional
+        Declination of the first star.
+
+    separation : `astropy.units.Quantity`, optional
+        Spacing between adjacent grid positions in right ascension and in
+        declination.
+
+    Returns
+    -------
+    ra : `astropy.units.Quantity`
+        Right ascension of each star.
+
+    dec : `astropy.units.Quantity`
+        Declination of each star.
     """
     # The plus one guarantees we'll have enough positions for
     # all of our stars.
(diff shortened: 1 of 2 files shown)


stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -351,24 +351,24 @@ def passband_columns(self, passbands=None, transformer=None):  # noqa: ARG002
...
+# fit model exactly with a=b=c=d=0, so this is the only non-zero coefficient.
+FAKE_CATALOG_ZERO_POINT = 20.0
 
-    # Good stars sit exactly on top of their catalog counterparts.
-    ra, dec = generate_star_coordinates(n_good)
 
-    good_mags = np.linspace(-10.0, -5.0, num=n_good)
-    color = np.linspace(0.0, 1.0, num=n_good)
+def generate_fake_catalog(n_stars):
+    """
```
mwcraig, line 437: Make this a numpy-style docstring

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -379,32 +379,38 @@ def test_transform_to_catalog_excludes_distant_matches(mocker):
...
-    # only the distance cut can remove it from the fit.
-    bad_ra = ra[0]
-    bad_dec = dec[0] - 1 * u.degree
-    bad_mag = catalog["mag_R"][0] - zero_point - 0.5
+    return catalog, ra, dec, instrumental
+
+
+def generate_observed_table(ra, dec, instrumental):
+    """
+    Generate observations in the form transform_to_catalog expects, i.e. a
```
mwcraig: Make this a numpy-style docstring

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -379,32 +379,38 @@ def test_transform_to_catalog_excludes_distant_matches(mocker):
...
-    observed = observed.group_by("file")
 
-    # Replace the catalog fetch with our synthetic catalog so no
-    # network access is needed.
+    return observed.group_by("file")
+
+
+def run_transform_to_catalog(mocker, catalog, observed):
+    """
+    Run transform_to_catalog against a synthetic catalog rather than a
```
mwcraig: Make this a numpy-style docstring

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -351,24 +351,24 @@ def passband_columns(self, passbands=None, transformer=None):  # noqa: ARG002
...
+# Zero point of the synthetic catalog below. The catalog magnitudes follow the
+# fit model exactly with a=b=c=d=0, so this is the only non-zero coefficient.
+FAKE_CATALOG_ZERO_POINT = 20.0
 
-    # Good stars sit exactly on top of their catalog counterparts.
-    ra, dec = generate_star_coordinates(n_good)
 
-    good_mags = np.linspace(-10.0, -5.0, num=n_good)
-    color = np.linspace(0.0, 1.0, num=n_good)
+def generate_fake_catalog(n_stars):
```
mwcraig: This is only for internal use, so prefix it with a `_`

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -379,32 +379,38 @@ def test_transform_to_catalog_excludes_distant_matches(mocker):
...
-    # chosen so that the bogus match is 0.5 magnitude off the true zero
-    # point -- close enough to survive the median-based outlier cut, so
-    # only the distance cut can remove it from the fit.
-    bad_ra = ra[0]
-    bad_dec = dec[0] - 1 * u.degree
-    bad_mag = catalog["mag_R"][0] - zero_point - 0.5
+    return catalog, ra, dec, instrumental
+
+
+def generate_observed_table(ra, dec, instrumental):
```
mwcraig: This is only for internal use, so prefix it with a `_`

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -379,32 +379,38 @@ def test_transform_to_catalog_excludes_distant_matches(mocker):
...
         }
     )
-    observed = observed.group_by("file")
 
-    # Replace the catalog fetch with our synthetic catalog so no
-    # network access is needed.
+    return observed.group_by("file")
+
+
+def run_transform_to_catalog(mocker, catalog, observed):
```
mwcraig: This is only for internal use, so prefix it with a `_`

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed. Here is the numpy-style version, matching the flavor used in the module under test (one-line summary, blank line after the `Parameters` underline, blank line between entries — as in `calibrated_from_instrumental`):

```python
def _generate_fake_catalog(n_stars):
    """
    Generate a catalog whose magnitudes are an exact fit to the transform model.

    Catalog magnitudes are generated with a=b=c=d=0 and
    z=FAKE_CATALOG_ZERO_POINT, so a fit to them recovers those values exactly.

    Parameters
    ----------

    n_stars : int
        Number of stars to generate.

    Returns
    -------
    catalog : `FakeCatalogTable`
        Catalog with columns ``ra``, ``dec``, ``mag_R`` and ``mag_I``.

    ra : `astropy.units.Quantity`
        Right ascension of each star.

    dec : `astropy.units.Quantity`
        Declination of each star.

    instrumental : `numpy.ndarray`
        Instrumental magnitudes from which the catalog magnitudes were generated.
    """
```

For context on why it was not written that way to begin with: the pre-existing helpers in this file — `generate_input_mags`, `generate_catalog_mags`, `generate_star_coordinates` and `generate_tables` — are all free-prose docstrings with no sections, so I matched the local idiom rather than the `stellarphot/utils/magnitude_transforms.py` idiom. Your call is the better one, since the source module is consistently numpy-style. Happy to convert those four pre-existing helpers in the same pass if you want the file internally consistent; say the word and I will, otherwise I will restrict the change to the three new helpers.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed:

```python
def _generate_observed_table(ra, dec, instrumental):
    """
    Generate observations in the form ``transform_to_catalog`` expects.

    Parameters
    ----------

    ra : `astropy.units.Quantity`
        Right ascension of each observed star.

    dec : `astropy.units.Quantity`
        Declination of each observed star.

    instrumental : `numpy.ndarray`
        Instrumental magnitude of each observed star.

    Returns
    -------
    `astropy.table.Table`
        Observations of a single image, grouped by file name, with the
        ``file``, ``passband``, ``ra``, ``dec``, ``mag_inst`` and ``mag_error``
        columns `transform_to_catalog` requires.
    """
```

Worth noting the `ra`/`dec` parameters must be `Quantity` rather than plain arrays — the body calls `.to_value("degree")` on them — so documenting the type here is doing real work, not just ceremony.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed:

```python
def _run_transform_to_catalog(mocker, catalog, observed):
    """
    Run ``transform_to_catalog`` against a synthetic catalog.

    Patching the catalog fetch keeps the test offline, so it does not need
    the ``remote_data`` marker.

    Parameters
    ----------

    mocker : `pytest_mock.MockerFixture`
        Fixture used to patch the `apass_dr9` catalog fetch.

    catalog : `FakeCatalogTable`
        Catalog to use in place of the fetched one.

    observed : `astropy.table.Table`
        Observations to transform, grouped by file name.

    Returns
    -------
    `astropy.table.Table`
        The observations with the calibrated magnitude, fit coefficient and
        matched-catalog columns added.
    """
```

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — `generate_fake_catalog` → `_generate_fake_catalog`.

I checked that the rename is safe before agreeing: `grep -rn "generate_fake_catalog\|generate_observed_table\|run_transform_to_catalog"` across the repo returns hits only in this file (the three definitions and their six call sites), so nothing imports them. And there is plenty of precedent for the convention in the test suite already — `_repeat` and `_raw_photometry_table` in `test_aij_rel_fluxes.py`, `_make_photometry_table` in `test_aij_io.py`, `_fake_coords` in `test_core.py`, `_ey_uma_header` in `test_catalogs.py`, among others.

Two related things in the same block that are also file-local, if you want the convention applied consistently rather than only to the three functions you flagged:

- `FakeCatalogTable` (line 344) → `_FakeCatalogTable`
- `FAKE_CATALOG_ZERO_POINT` (line 356) → `_FAKE_CATALOG_ZERO_POINT`

I left both alone because you did not mark them, and renaming the class in particular touches the docstrings I am rewriting for your other comments. Tell me either way and I will make it consistent in one pass.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — `generate_observed_table` → `_generate_observed_table`. Two call sites to update, at lines 443 and 496.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — `run_transform_to_catalog` → `_run_transform_to_catalog`. Two call sites to update, at lines 449 and 502.

The leading underscore is worth a little extra here, since `_run_transform_to_catalog` sits one character away from the `transform_to_catalog` it wraps; the prefix makes it obvious at each call site which one is the helper and which is the function under test.

PR #676, opened 2026-08-10
Consolidate magnitude_transforms.py onto one tested pipeline

mwcraig:
Fixes #601.

`magnitude_transforms.py` had two parallel calibration pipelines, and the split fell the wrong way round. `transform_to_catalog` — the function users' calibrated magnitudes actually flow through, and the one the shipped notebook calls — had two narrow regression tests from the #617/#668 coordinate-matching bugs. `calculate_transform_coefficients` / `transform_magnitudes` were thoroughly tested and called by nothing outside their own tests.

This keeps the pipeline that is used, deletes the one that is not, and gives the survivor real tests.

## What changed

**Pipeline B removed.** `calculate_transform_coefficients` and `transform_magnitudes` are gone, along with the `transform_to_catalog` helpers `opts_to_str` and `calc_residual`, which lmfit's own reporting supersedes. `filter_transform` and `calibrated_from_instrumental` are untouched.

**The fitter is now lmfit, not `scipy.optimize.curve_fit`.** This is not cosmetic. The old bounds were doing two unrelated jobs at once: genuinely constraining `a`/`c`/`z`, and *faking fixed parameters* for `b` and `d` with a box 1e-6 wide, because `curve_fit` has no concept of a fixed parameter. lmfit expresses the two separately, so the signature changes:

```python
vary=("a", "c", "z")   # terms that are fit; every other term is held at exactly 0
expected=None          # {term: (low, high)}; post-fit warning only, default {"z": (18, 22)}
```

replacing `a_delta`, `a_cen`, `b_delta`, `c_delta`, `d_delta` and `zero_point_range`.

The important behavioral change is that `expected` **never constrains the fit**. A true zero point outside `zero_point_range` used to rail at the boundary and report a confident wrong answer with no signal at all. It is now fit freely and warned about, by name and value.

`verbose` is gone; the prints are now `AstropyUserWarning`s. Because the suite runs with `filterwarnings = ["error", ...]`, every test that does not expect a warning is now also asserting the function stayed quiet — previously the tests could not distinguish "fit succeeded" from "filled the column with NaN".

The per-image progress line is dropped rather than converted. Warning on every *successful* fit would defeat the point of the conversion, and the coefficients it printed are already columns in the output table.

[...]

PR #682, opened 2026-08-11
Report per-image transform problems as log messages instead of warnings

mwcraig:
Closes #679.

`transform_to_catalog` reported problems with individual images -- no usable
stars, an underdetermined or failed fit, a term outside its expected range --
as warnings raised inside the per-image loop. Under `-W error` any one of them
aborted the whole calibration, discarding the results of every image already
fit (issue #679).

Review discussion here landed on a simpler position than the deferral this PR
originally implemented: these messages are per-image data-quality notes, not
API-misuse warnings, so they are now `logger.warning` calls -- the same
pattern `single_image_photometry` and `multi_image_photometry` already use.
With nothing in the loop able to escalate, #679's failure mode no longer
exists, and the deferral machinery and `try/finally` from earlier revisions of
this branch are gone with it.

The split this leaves, matching the rest of the package:

- **Anticipated per-image data problems** are logged and the image is skipped;
  its rows stay NaN and every other image keeps its results.
- **Unanticipated exceptions** (e.g. lmfit crashing on data that passed the
  checks) are bugs, and crash loudly with no salvage attempt -- partial
  results are not worth preserving, since the fix is to re-run anyway.
- The pre-loop warning about calling with no `obs_error_column` stays a real
  warning: it is advice about the call itself, fires before any work is done,
  and escalating it discards nothing.

### Also

`_write_output_columns` is a new helper next to `_merge_with_existing`; output
column order is unchanged. This is groundwork -- four follow-up PRs (#678,
#681, #677, #674) each add columns, and this turns each of those into one dict
entry instead of another edit to the same straight-line block.

This is the first of five stacked PRs off the #676 follow-ups; the rest build
on this branch.

---

### DIFF
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -1,3 +1,4 @@
+import logging
 import warnings
 
 import lmfit
@@ -8,6 +9,8 @@ from astropy.utils.exceptions import AstropyUserWarning
 
 from ..catalogs import apass_dr9, refcat2
 from .magnitude_system_transforms import transform_apass_bands, transform_refcat2_bands
+
+logger = logging.getLogger(__name__)
 
 __all__ = [
     "calibrated_from_instrumental",
@@ -465,15 +468,6 @@ def transform_to_catalog(
         If ``True``, add the calibrated magnitude to the input table, otherwise return
         a copy.
 
-        Warnings about an individual image are held until every image's
-        results have been written, so a warning does not cost the images that
-        were fit successfully their results. That does not stop the call
-        raising when warnings are errors -- under ``-W error`` the deferred
-        warnings still escalate, only after the columns have been written. So
-        with ``in_place=True`` the table the caller handed in is complete when
-        the call raises, while with ``in_place=False`` the copy that would
-        have been returned is lost.
-
     fit_diff : bool, optional
         If ``True``, fit the difference between the instrumental and catalog magnitude
         instead of the treating the catalog mag as the dependent variable.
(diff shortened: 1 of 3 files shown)


stellarphot/utils/magnitude_transforms.py
```diff
@@ -437,6 +465,15 @@ def transform_to_catalog(
         If ``True``, add the calibrated magnitude to the input table, otherwise return
         a copy.
 
+        Warnings about an individual image are held until every image's
```
mwcraig: This is way too long to be part of an • argument doc string. Either shorter it or move it to notes and refer to that note here.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -551,197 +588,204 @@ def transform_to_catalog(
...
+            output_columns[_CAL_MAG_ERROR_COLUMN] = scaled_errors
+
+        output_columns[_CAL_MAG_COLUMN] = cal_mags
+        output_columns.update(coefficients)
+        output_columns["mag_cat"] = cat_mags
+        output_columns["color_cat"] = cat_colors
+
+        _write_output_columns(result, output_columns, in_passband)
+
+    for message in deferred_warnings:
```
mwcraig: So all of the warnings we are deferring are ones we generate? If that is the case could we maybe issue log messages instead of warnings and simplify this whole thing?

stellarphot/utils/magnitude_transforms.py
```diff
@@ -551,197 +588,204 @@ def transform_to_catalog(
...
+            # a negative number working out the uncertainties, and the
+            # RuntimeWarning that produces is an outright exception for anyone
+            # running with warnings as errors. Whether the terms can be told
+            # apart from *each other* is a question about the fit and is asked
+            # of it below.
+            if fit_mag.size <= len(vary):
+                deferred_warnings.append(
+                    f"Fit for {file[0]} is underdetermined: {fit_mag.size} usable "
+                    f"star(s) cannot determine {len(vary)} term(s) {list(vary)}; "
+                    "the fit has no degrees of freedom left"
```
mwcraig: Drop this line of the message.

mwcraig (commented):
I am no longer convinced that this is a problem that needs to be solved. Real users will not run with -W error. Our test suite controls the data we feed into the tests so we can check that the warnings we expect are raised when we expect them. Any other the they occur in the test suite really should be treated as an error in the test I think.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — this paragraph is about the function's warning behavior, not about `in_place` specifically, so it doesn't belong here. If we keep the deferral machinery, the move would be:

Under `in_place`, one sentence:

```
    in_place : bool, optional
        If ``True``, add the calibrated magnitude to the input table, otherwise return
        a copy. See Notes for how warnings interact with this choice.
```

and the rest of the paragraph goes to Notes.

One dependency worth noting: if the question at line 788 lands on converting the deferred warnings to log messages, this paragraph mostly gets *deleted* rather than moved — there would no longer be an escalation to warn about. So it's probably worth settling that thread first.

stellarphot/utils/magnitude_transforms.py
mwcraig: Yes — every deferred warning is generated in this function. `deferred_warnings` only ever gets our own f-strings appended; warnings raised by lmfit/numpy inside the loop are *not* deferred (the `try/finally` is what protects the results if one of those escalates).

And converting them to log messages would simplify real things:

- The `deferred_warnings` list, the re-raise loop at the bottom, and the whole "-W error still raises, but after the columns are written" caveat in the `in_place` docstring all disappear — which also resolves your comment at line 468 by deletion.
- It matches the codebase precedent: `single_image_photometry` / `multi_image_photometry` already report per-image problems via `logger.warning(...)` and keep going. This loop is the same shape.
- The #679 failure mode goes away outright, since log messages never escalate under `-W error`.

What it would *not* remove is the `try/finally` — that guards against real exceptions (lmfit blowing up, third-party warnings escalating), not our warnings, so it stays either way.

The costs, so we choose with eyes open:

- Callers lose the ability to catch these with the `warnings` machinery; the ~10 new tests using `pytest.warns` would switch to `caplog`.
- A run where *every* image fails would return an all-NaN table without raising, even under `-W error`. The message still reaches stderr by default (`logger.warning` prints via logging's last-resort handler when nothing is configured), but nothing programmatic fires unless the caller checks the result.

My suggestion: convert the six in-loop per-image messages to `logger.warning`, but keep the pre-loop `obs_error_column is None` message as a real warning — that one is "you probably called this wrong" advice about the API call, which is classic `warnings` territory, and it fires before any work is done so deferral was never needed for it.

Happy to implement that on this branch if you want to go that way.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — "the fit has no degrees of freedom left" restates what "N usable star(s) cannot determine M term(s)" already says. It's also a free deletion: the tests match on `"underdetermined"`, not on that phrase, so nothing else changes. Will drop it when applying the review fixes.

PR #683, opened 2026-08-11
Apply one match cut to everything the catalog match produces

mwcraig:
Closes #678. Closes #681.

Stacked on #682 — please merge that first; this branch contains its commit.

Two issues that rewrite the same twenty lines. Split, they conflict line for
line, and #681 itself says they are best fixed together.

### #678 — one cut for everything the match produces

`mag_cal` was NaN'd beyond 1.5 arcsec but `mag_cat`, `color_cat` and
`mag_cal_error` were not, so a star whose nearest catalog entry is a degree away
still came back with that unrelated star's catalog magnitude and colour, and
with a finite calibrated error attached to a NaN calibrated magnitude — which
makes `mag_cal_error` unsafe to select rows on.

Now a single `matched = d2d.arcsecond <= 1.5` decides all of them. The two bare
literals become `_FIT_MATCH_ARCSEC = 1.0` and `_CAL_MATCH_ARCSEC = 1.5`, with
the reason the band between them exists — V2480 Cyg's VSX position is ~1.3
arcsec from its APASS DR9 position — written down next to them rather than
buried in a comment further down.

`mag_cal_error` moves into the per-image loop, where the distances and
calibrated magnitudes it depends on actually live, and gains the invariant that
makes the complaint go away:

```python
# A star with no calibrated magnitude has no calibrated error either,
# whatever the reason it has no magnitude.
cal_error[~np.isfinite(cal_mag)] = np.nan
```

That is stronger than repeating the distance cut — it also covers a star with no
instrumental magnitude — and gives the rule worth relying on: **a finite
`mag_cal_error` always has a finite `mag_cal`.**

### #681 — do not demand a colour nobody asked for

A star with no catalog colour was dropped from the fit even when no colour term
was being fit, and its `mag_cal` came out NaN because `0.0 * np.nan` is NaN. So
a catalog with sparse colours could produce "No good data" for an image whose
fit never needed a colour at all.

A star is now required to have a colour only when `c` or `d` is in `vary`, and
the model is evaluated with a neutral zero standing in for missing colours
(`np.where`, not `np.nan_to_num`, which would turn an infinite colour into a
very large finite one). The raw colour is what still goes into `color_cat`, so a
star whose colour is genuinely unknown reports NaN.

### Behaviour change worth noting

Rows with a NaN `mag_inst` now get NaN `mag_cal_error` instead of a scaled
instrumental error. No test asserted the old value, but it is a real change and
is in `CHANGES.rst`.

---

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -152,17 +152,11 @@ Bug Fixes
   aborts the calibration; output columns are written after all images are
   processed. [#682]
 + ``transform_to_catalog()`` now applies one match-distance cut to everything
-  it derives from the catalog match. ``mag_cat`` and ``color_cat`` used to be
-  taken from the nearest catalog entry however far away it was, and
-  ``mag_cal_error`` was reported for every row with a usable input error, so
-  both could describe an unrelated star on a row whose ``mag_cal`` was NaN.
-  All of them are now NaN beyond 1.5 arcsec, the same limit ``mag_cal``
-  follows, and ``mag_cal_error`` is NaN wherever ``mag_cal`` is -- including
-  for a star with no instrumental magnitude, which previously got a finite
-  calibrated error. A star is also no longer required to have a catalog color
-  unless a color term (``"c"`` or ``"d"``) is actually being fit, so stars the
-  catalog has no color for can be fit and calibrated when ``vary`` does not
-  include one. [#678, #681]
+  it derives from the catalog match: ``mag_cat``, ``color_cat`` and
+  ``mag_cal_error`` are NaN wherever ``mag_cal`` is, instead of reporting the
+  nearest catalog star's values however far away it was. [#678]
++ ``transform_to_catalog()`` no longer requires a star to have a catalog
+  color unless a color term (``"c"`` or ``"d"``) is actually being fit. [#681]
 
 2.1.2 (2026-07-19)
 -------------------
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -146,6 +146,23 @@ Bug Fixes
   and a partial save no longer resurrects stale values from a conflicting
   partial settings file. Encoding and OS errors while reading settings now
   raise ``SettingsFileReadError``, a ``ValueError`` subclass. [#662]
++ ``transform_to_catalog()`` now writes its results to the table before any of
+  the warnings about individual images are raised, so running with warnings as
+  errors (``-W error``) no longer discards the results of every image that was
+  fit successfully. With ``in_place=True`` the caller's table is complete even
+  when a deferred warning escalates and the call raises. [#679]
++ ``transform_to_catalog()`` now applies one match-distance cut to everything
```
mwcraig, line 154: This is way too long.maybe break it into two entries, one for each bug, or just shorten

mwcraig: Agreed — it's covering two different bugs, so two entries, one per issue:

```rst
+ ``transform_to_catalog()`` now applies one match-distance cut to everything
  it derives from the catalog match: ``mag_cat``, ``color_cat`` and
  ``mag_cal_error`` are NaN wherever ``mag_cal`` is, instead of reporting the
  nearest catalog star's values however far away it was. [#678]
+ ``transform_to_catalog()`` no longer requires a star to have a catalog
  color unless a color term (``"c"`` or ``"d"``) is actually being fit. [#681]
```

The "star with no instrumental magnitude used to get a finite calibrated error" detail folds into the first entry's "NaN wherever ``mag_cal`` is" without needing its own sentence. Happy to apply this on the branch.

PR #684, opened 2026-08-11
Report how well each image's transform fit

mwcraig:
Closes #677.

`transform_to_catalog` reported the transform coefficients but nothing about how
well they were determined, so there was no way to tell a tight fit from a
hopeless one except by eye. Six new columns, repeated down every row like
`a`…`z` already are: **`a_error`, `b_error`, `c_error`, `d_error`, `z_error`,
`fit_redchi`**. No change to the return contract.

The uncertainties are lmfit's own, read straight off the fit result: a term not
in `vary` is held at exactly zero so its uncertainty is exactly `0.0`, and an
image that was never fit stays NaN. `fit_redchi` is the summed squared
residuals per degree of freedom (stars fit minus terms varied).

### Two things the docstring has to say

**`fit_redchi` means two different things.** It is a reduced chi-square only
when `obs_error_column` is given. Unweighted, the weights are the scalar 1.0
and the same column holds the residual scatter in **mag²** — same name,
completely different scale.

**`scale_covar` is on**, so every `*_error` is scaled by `redchi`. The reported
coefficient uncertainties therefore reflect the *observed* scatter rather than
the quoted instrumental errors — robust when the quoted errors are wrong, but
it means `a_error` and `fit_redchi` are not independent diagnostics. Whether to
keep that convention is #690.

### The numbers are real uncertainties, and the tests say so

- **`z_error` falls by ~√10 from 25 to 250 stars**: measured 0.02311 → 0.006706,
  ratio **3.45** against √10 = 3.162, averaged over 5 noise seeds — at 25 stars
  a single seed's `√redchi` fluctuates ~15%, so a single-seed assertion would
  have been a statement about one draw.
- **`fit_redchi` = 1.008** when `mag_error` matches the injected 0.02 mag noise,
  **100.83** when `mag_error` is 10× too small, and **≈ (0.02 mag)² in mag²**
  when no error column is given at all.

### Changed in review round 1

The first version gated the uncertainties on the covariance being present and
finite, and tested that gate with mocked fit results. Review pointed out,
correctly, that neither state is reachable past the guards upstream, so the
gate, the helper that held it, and both mocked tests are gone; the
covariance-usability verdict returns with the #674 follow-up, where the matrix
is actually consumed. Also from review: the changelog entry was cut to size,
the redchi wording fixed (lmfit divides by `nfree`, not `N`), and the noisy
test helper folded into `_generate_observed_table`.

---

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,16 @@ New Features
   to the variable-aperture defaults (multiples of the measured FWHM, not
   pixels), with a warning (see issue ``#654``). A file written by a
   newer stellarphot raises ``NewerFormatError``. [#656]
++ ``transform_to_catalog()`` now reports how good each image's fit was, in six
+  new columns repeated down every row of that image alongside the coefficients
+  themselves: ``a_error`` through ``z_error``, the uncertainty of each
+  coefficient, and ``fit_redchi``. A term that is not being fit is held at
+  exactly zero and its uncertainty is exactly zero; the uncertainties are NaN,
+  while the coefficients are still reported, for a fit that converged but left
+  no usable covariance behind. ``fit_redchi`` is a reduced chi-square only
+  when ``obs_error_column`` is given -- unweighted it is the mean squared
+  residual in mag squared -- and, because ``lmfit`` scales the covariance by
+  it, it and the ``*_error`` columns are not independent diagnostics. [#677]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,16 +18,12 @@ New Features
   to the variable-aperture defaults (multiples of the measured FWHM, not
   pixels), with a warning (see issue ``#654``). A file written by a
   newer stellarphot raises ``NewerFormatError``. [#656]
-+ ``transform_to_catalog()`` now reports how good each image's fit was, in six
-  new columns repeated down every row of that image alongside the coefficients
-  themselves: ``a_error`` through ``z_error``, the uncertainty of each
-  coefficient, and ``fit_redchi``. A term that is not being fit is held at
-  exactly zero and its uncertainty is exactly zero; the uncertainties are NaN,
-  while the coefficients are still reported, for a fit that converged but left
-  no usable covariance behind. ``fit_redchi`` is a reduced chi-square only
-  when ``obs_error_column`` is given -- unweighted it is the mean squared
-  residual in mag squared -- and, because ``lmfit`` scales the covariance by
-  it, it and the ``*_error`` columns are not independent diagnostics. [#677]
++ ``transform_to_catalog()`` now reports how well each image's transform
+  fit, in six new columns repeated down the image's rows like the
+  coefficients themselves: ``a_error`` through ``z_error``, the uncertainty
+  of each coefficient, and ``fit_redchi``, the summed squared residuals per
+  degree of freedom -- in units of the supplied errors, or in mag² when no
+  ``obs_error_column`` is given. [#677]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -214,59 +214,6 @@ def _underdetermined_reason(fit_result, vary):
     return None
 
 
-def _coefficient_uncertainties(fit_result):
-    """
-    Uncertainty of each coefficient of a fit, when the fit can supply them.
-
-    This is the single place the question "does this fit have a usable
-    covariance?" is answered, so that everything derived from the covariance
-    agrees about whether there is one to derive anything from.
-
-    Note that ``fit_result.errorbars`` is deliberately *not* what is checked.
-    `lmfit` clears that flag whenever any varied term comes out with a
-    standard error of exactly zero, and that is what a perfect fit gives:
-    data that follows the model to the last bit has a covariance that really
-    is zero, and zero is the right answer there rather than a missing one.
-    What makes the uncertainties unavailable is ``covar`` being absent
-    altogether, or having non-finite entries -- an inverse that came back with
-    NaN in it looks like a matrix and is not one, and anything that reads it
-    without checking reports a confident wrong number instead of failing.
-
-    Parameters
-    ----------
-
-    fit_result : `lmfit.minimizer.MinimizerResult`
-        Result of fitting one image.
-
-    Returns
-    -------
-    uncertainties : dict
-        Standard error of each term of the transform model, keyed by term
-        name. A term that was not varied was held at exactly zero and is
-        therefore known exactly, so its uncertainty is ``0.0``. Every value is
-        NaN if the covariance cannot be used.
-
-    covar : `numpy.ndarray` or None
-        The covariance matrix, or `None` if it cannot be used. Anything that
-        needs the correlations between the terms as well as their individual
-        uncertainties should take the matrix from here rather than asking the
-        fit result again, so that there is only ever one verdict.
-    """
-    covar = fit_result.covar
-
-    if covar is None or not np.all(np.isfinite(covar)):
-        return {name: np.nan for name in _COEFF_NAMES}, None
-
-    return {
-        name: (
-            np.nan
-            if fit_result.params[name].stderr is None
-            else float(fit_result.params[name].stderr)
-        )
-        for name in _COEFF_NAMES
-    }, covar
-
-
 def _check_known_terms(terms, argument_name):
     """
     Raise if any of ``terms`` is not a coefficient of the transform model.
@@ -579,18 +526,19 @@ def transform_to_catalog(
     for one image rather than any one star, so each of them is the same on
     every row of that image. A term that is not in ``vary`` is held at exactly
     zero and is therefore known exactly, so its uncertainty is exactly zero
-    rather than a small number. The uncertainties are NaN, while the
-    coefficients themselves are still reported, for an image whose fit
-    converged but left no usable covariance behind.
+    rather than a small number.
 
-    ``fit_redchi`` is a reduced chi-square only when ``obs_error_column`` is
-    given, because only then is anything dividing the residuals: it is the
-    mean squared residual in units of the errors that were supplied, and a
-    value near one says the model misses the stars by about as much as they
-    claim to be uncertain. Without an error column the fit is unweighted and
-    the same column holds the mean squared residual in **mag squared** -- the
-    same name and a completely different scale, so values from a weighted and
-    an unweighted fit must not be compared with each other.
+    ``fit_redchi`` is `lmfit`'s reduced chi-square: the summed squared
+    residuals per degree of freedom, i.e. divided by the number of stars fit
+    minus the number of terms varied. It is a chi-square only when
+    ``obs_error_column`` is given, because only then is anything dividing the
+    residuals: the residuals are in units of the errors that were supplied,
+    and a value near one says the model misses the stars by about as much as
+    they claim to be uncertain. Without an error column the fit is unweighted
+    and the same column holds the summed squared residuals per degree of
+    freedom in **mag squared** -- the same name and a completely different
+    scale, so values from a weighted and an unweighted fit must not be
+    compared with each other.
 
     The reported uncertainties are scaled by ``fit_redchi``, because `lmfit`'s
     ``scale_covar`` is left on. They therefore describe the scatter actually
@@ -829,15 +777,18 @@ def transform_to_catalog(
 
         values = {name: fit_result.params[name].value for name in _COEFF_NAMES}
 
-        # How well the fit pinned each term down, and how well the model it
-        # landed on describes the stars. Both are answers about the image
-        # rather than about any one star, and both are reported even when
-        # they are the only thing that would tell a caller the coefficients
-        # below are not worth applying.
-        # The covariance matrix comes back too, but nothing here needs the
-        # correlations between the terms -- only the individual
-        # uncertainties, which are its diagonal.
-        uncertainties, _ = _coefficient_uncertainties(fit_result)
+        # How well the fit pinned each term down. lmfit's stderr is exactly
+        # 0.0 for a term that was not varied -- held at zero, so known
+        # exactly -- and None in the states where no uncertainty could be
+        # worked out at all.
+        uncertainties = {
+            name: (
+                np.nan
+                if fit_result.params[name].stderr is None
+                else float(fit_result.params[name].stderr)
+            )
+            for name in _COEFF_NAMES
+        }
 
         # The expected ranges are a check on the answer, not a constraint on
         # the fit, so a value outside its range is reported and kept.
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -35,6 +35,14 @@ _DEFAULT_EXPECTED = {"z": (18.0, 22.0)}
 _CAL_MAG_COLUMN = "mag_cal"
 _CAL_MAG_ERROR_COLUMN = "mag_cal_error"
 
+# Suffix of the column reporting the uncertainty of a fit coefficient, so that
+# ``a`` is reported alongside ``a_error``. Matches the mag/mag_error and
+# mag_cal/mag_cal_error pairs already in the table.
+_COEFF_ERROR_SUFFIX = "_error"
+
+# Name of the column reporting how well one image's transform fit its stars.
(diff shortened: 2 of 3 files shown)


stellarphot/utils/magnitude_transforms.py
```diff
@@ -206,6 +214,59 @@ def _underdetermined_reason(fit_result, vary):
...
 
+def _coefficient_uncertainties(fit_result):
+    """
+    Uncertainty of each coefficient of a fit, when the fit can supply them.
+
+    This is the single place the question "does this fit have a usable
+    covariance?" is answered, so that everything derived from the covariance
+    agrees about whether there is one to derive anything from.
+
+    Note that ``fit_result.errorbars`` is deliberately *not* what is checked.
```
mwcraig: Move to the Notes section

stellarphot/utils/magnitude_transforms.py
```diff
@@ -206,6 +214,59 @@ def _underdetermined_reason(fit_result, vary):
...
+    `lmfit` clears that flag whenever any varied term comes out with a
+    standard error of exactly zero, and that is what a perfect fit gives:
```
mwcraig: The standard error is NEVER EVER ZERO on real data. There is no reason to avoid using `fit_result.errobars`

stellarphot/utils/magnitude_transforms.py
```diff
@@ -513,11 +575,39 @@ def transform_to_catalog(
     Notes
     -----
 
+    The coefficients, their uncertainties and ``fit_redchi`` describe the fit
+    for one image rather than any one star, so each of them is the same on
+    every row of that image. A term that is not in ``vary`` is held at exactly
+    zero and is therefore known exactly, so its uncertainty is exactly zero
+    rather than a small number. The uncertainties are NaN, while the
```
mwcraig: That only happens in a fake perfect-fit case, right?

stellarphot/utils/magnitude_transforms.py
```diff
@@ -513,11 +575,39 @@ def transform_to_catalog(
...
+    given, because only then is anything dividing the residuals: it is the
+    mean squared residual in units of the errors that were supplied, and a
+    value near one says the model misses the stars by about as much as they
+    claim to be uncertain. Without an error column the fit is unweighted and
+    the same column holds the mean squared residual in **mag squared** -- the
+    same name and a completely different scale, so values from a weighted and
+    an unweighted fit must not be compared with each other.
+
+    The reported uncertainties are scaled by ``fit_redchi``, because `lmfit`'s
+    ``scale_covar`` is left on. They therefore describe the scatter actually
```
mwcraig, line 544: Say more about what this means  -- I doni't think the code description needs to be longer, but explain in a reply what you mean.

CHANGES.rst
```diff
@@ -18,6 +18,16 @@ New Features
   to the variable-aperture defaults (multiples of the measured FWHM, not
   pixels), with a warning (see issue ``#654``). A file written by a
   newer stellarphot raises ``NewerFormatError``. [#656]
++ ``transform_to_catalog()`` now reports how good each image's fit was, in six
```
mwcraig: Another changelog entry that is way too long

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -314,6 +314,59 @@ def _generate_observed_table(
     return observed.group_by("file")
 
 
+def _noisy_observations(ra, dec, instrumental, sigma, seed, mag_error=None, **kwargs):
```
mwcraig: Why a completely new function instead of adding a new argument to the other function that controls the amount of error addded?

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -685,6 +759,179 @@ def test_transform_to_catalog_weights_by_inverse_error(mocker):
...
+    # One fit per image, so this too is repeated down every row.
+    assert (reported == reported[0]).all()
+    assert reported[0] == pytest.approx(expected_redchi, rel=0.3)
+
+
+def _break_the_covariance(mocker, how):
+    """
+    Make the fit come back with a covariance matrix that cannot be used.
+
+    Neither state is reachable from data. ``lmfit.minimize`` only leaves
```
mwcraig: "Neither state is reachable from data" means there is no reason at all to test it.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -685,6 +759,179 @@ def test_transform_to_catalog_weights_by_inverse_error(mocker):
...
+    mocker.patch.object(
+        magnitude_transforms.lmfit,
+        "minimize",
+        side_effect=minimize_then_spoil_covariance,
+    )
+
+
+@pytest.mark.parametrize("how", ["missing", "not_finite"])
+def test_transform_to_catalog_nans_uncertainties_without_a_covariance(mocker, how):
+    # A fit can converge and still leave nothing to work the uncertainties out
```
mwcraig: "A fit can converge and still leave nothing...." no, that cannot happen with real data.

mwcraig (commented):
I do not see why we are coding around an artifical case that you acknowledge cannot happen with real data.

stellarphot/utils/magnitude_transforms.py
mwcraig: This paragraph is going away entirely rather than moving: per the summary thread, the `covar` check it justifies is coming out of this PR, and `_coefficient_uncertainties` shrinks to reading `stderr` off the fit result. If any pointer survives it will be one line in a Notes section, as you ask.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed that the standard error is never exactly zero on real data. The plan (summary thread) is to key on nothing at all: when lmfit fails to compute uncertainties it leaves `stderr` as `None`, which this code already maps to NaN, so reading `stderr` directly gives identical output everywhere reachable — no check on `covar` *or* `errorbars`.

For the record, `errorbars` specifically was avoided for a reason beyond the perfect-fit case: checking it would not have removed any code (the `None` → NaN mapping is still needed), and the flag answers a different question than "is the covariance usable" — lmfit clears it whenever any varied term's `stderr > 0.0` comparison fails (`minimizer.py` L822), yet leaves it `True` when the covariance is NaN off the diagonal. That distinction only matters once the full matrix is consumed, which is #674's problem, so the verdict logic moves to that PR.

stellarphot/utils/magnitude_transforms.py
mwcraig: Not the perfect-fit case, but yes, artificial. A perfect fit still *has* a covariance (all zeros) and reports `*_error = 0.0`. This sentence describes `covar is None`, which needs `np.linalg.inv` to raise on an exactly singular design — and the underdetermined guard (condition number > 3e3, vs ~1e16 to actually break the inversion) rejects those images before any results are kept. Unreachable today, and the sentence is deleted along with the branch it describes (summary thread).

stellarphot/utils/magnitude_transforms.py
mwcraig: What lmfit does with `scale_covar=True`: after computing `covar = inv(JᵀJ)` from the weighted problem, it multiplies the whole matrix by `redchi` before taking `stderr = sqrt(diag)`. That is equivalent to saying "trust the *pattern* of the supplied errors but not their absolute scale, and set the scale from the fit's own residuals" — i.e. rescale so the reduced chi-square would come out 1.

Two consequences:

1. **Robustness.** If `mag_error` is quoted 10× too small, `redchi` comes out ≈100 and the ×redchi covariance scaling exactly cancels the underestimate — the reported `a_error`…`z_error` land where the *observed* scatter says they should, unchanged. (The redchi test in this PR is that exact scenario.)
2. **The price.** `fit_redchi` and the `*_error` columns are then computed from the same residual scatter. An image whose stars scatter 2× what they claim reports `redchi ≈ 4` *and* 2×-inflated coefficient errors — one measurement shown twice, so a caller must not treat "the errors look small" and "redchi looks fine" as two independent confirmations.

It is the same convention as `scipy.optimize.curve_fit`'s default (`absolute_sigma=False`). The alternative, `scale_covar=False`, believes the quoted errors absolutely: `*_error` and `fit_redchi` would then be independent diagnostics, but a mis-quoted `mag_error` would corrupt the reported uncertainties directly.

CHANGES.rst
mwcraig: Agreed. Proposed replacement, which also fixes Copilot's (correct) point that `redchi` is per degree of freedom rather than a mean:

```rst
+ ``transform_to_catalog()`` now reports how well each image's transform
  fit, in six new columns repeated down the image's rows like the
  coefficients themselves: ``a_error`` through ``z_error``, the uncertainty
  of each coefficient, and ``fit_redchi``, the summed squared residuals per
  degree of freedom -- in units of the supplied errors, or in mag² when no
  ``obs_error_column`` is given. [#677]
```

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: No strong reason — will merge it in. The only intent the wrapper carried was the coupled default: `mag_error` defaults to the injected `sigma`, so reporting *truthful* errors is the default and a test that wants lying errors has to say so explicitly. That survives fine as `noise_sigma=0.0, seed=None` arguments on `_generate_observed_table`, with the same default coupling when `noise_sigma` is nonzero.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — with the branch gone from this PR (summary thread), both mocked tests go with it. They reappear in the #674 follow-up, where the covariance matrix is actually consumed and a NaN matrix would silently propagate into confident wrong `mag_cal_error` values. There the check is load-bearing and reachable through its consumer rather than only through a mock.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Right — same resolution as the `_break_the_covariance` thread: the branch this test exercises comes out of the PR, and the test is deleted with it.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -513,11 +575,39 @@ def transform_to_catalog(
...
+    zero and is therefore known exactly, so its uncertainty is exactly zero
+    rather than a small number. The uncertainties are NaN, while the
+    coefficients themselves are still reported, for an image whose fit
+    converged but left no usable covariance behind.
+
+    ``fit_redchi`` is a reduced chi-square only when ``obs_error_column`` is
+    given, because only then is anything dividing the residuals: it is the
+    mean squared residual in units of the errors that were supplied, and a
+    value near one says the model misses the stars by about as much as they
+    claim to be uncertain. Without an error column the fit is unweighted and
```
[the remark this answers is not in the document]
mwcraig: Correct — confirmed against lmfit 1.3.4 (`minimizer.py` line 325: `redchi = chisqr / max(1, nfree)`). The docstring will say "the summed squared residuals per degree of freedom (stars fit minus terms varied)" instead of "mean squared residual".

On reachability of the extreme case: fits with fewer usable stars than varied terms are rejected before fitting, so `nfree >= 1` always, but a minimally determined fit (e.g. 4 stars, 3 varied terms, `nfree == 1`) can pass the guards, and there the distinction is large, as you say.

CHANGES.rst
```diff
@@ -18,6 +18,16 @@ New Features
...
+  new columns repeated down every row of that image alongside the coefficients
+  themselves: ``a_error`` through ``z_error``, the uncertainty of each
+  coefficient, and ``fit_redchi``. A term that is not being fit is held at
+  exactly zero and its uncertainty is exactly zero; the uncertainties are NaN,
+  while the coefficients are still reported, for a fit that converged but left
+  no usable covariance behind. ``fit_redchi`` is a reduced chi-square only
+  when ``obs_error_column`` is given -- unweighted it is the mean squared
+  residual in mag squared -- and, because ``lmfit`` scales the covariance by
+  it, it and the ``*_error`` columns are not independent diagnostics. [#677]
```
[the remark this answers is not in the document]
mwcraig: Agreed. This entry is also being shortened at the maintainer's direction; the proposed replacement (posted in the maintainer's thread on this entry) says "summed squared residuals per degree of freedom".

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -685,6 +759,179 @@ def test_transform_to_catalog_weights_by_inverse_error(mocker):
...
+        assert np.isfinite(reported).all(), term
+        assert (reported > 0).all(), term
+        # One fit per image, so every row of the image carries the same value.
+        assert (reported == reported[0]).all(), term
+
+    # A term that is not fit is held at exactly zero and is therefore known
+    # exactly. Its uncertainty is exactly zero -- not NaN, and not a small
+    # number, which is the same exact comparison
+    # test_transform_to_catalog_fixes_unvaried_terms makes of the value.
+    assert (np.asarray(result["b_error"]) == 0.0).all()
```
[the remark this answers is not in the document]
mwcraig: Declined — this cuts the opposite way from the maintainer's review, and the `errorbars` special-casing is being removed from the PR entirely: `_coefficient_uncertainties` will read `stderr` straight off the fit result, so no `errorbars`-dependent branch will remain to exercise. A varied term with `stderr == 0.0` is then a value passing through, not a code path.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -685,6 +759,179 @@ def test_transform_to_catalog_weights_by_inverse_error(mocker):
...
+    mocker, error_scale, expected_redchi
+):
+    # See issue #677.
+    n_stars = 100
+    sigma = 0.02
+
+    catalog, ra, dec, instrumental = _generate_fake_catalog(n_stars)
+    observed = _noisy_observations(
+        ra, dec, instrumental, sigma, seed=13579, mag_error=error_scale * sigma
+    )
```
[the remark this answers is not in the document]
mwcraig: Accepted — a noisy unweighted case will be added asserting `fit_redchi ≈ σ²` (the unweighted residual sum of squares per degree of freedom, in mag²), which pins the documented "same column, different units" behavior instead of only checking that the call completes.

stellarphot/utils/magnitude_transforms.py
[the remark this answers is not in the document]
mwcraig: Why aren't we just using fit_result.redchi instead of rolling our own?

stellarphot/utils/magnitude_transforms.py
mwcraig: Open an issue to remind us to discuss this further. For now proceed with your plan.

mwcraig: Opened #690 for that discussion — it lays out the two conventions (`scale_covar` on vs. off) and notes that the #674 follow-up inherits whichever one is chosen, since the propagated `mag_cal_error` comes from this covariance.

stellarphot/utils/magnitude_transforms.py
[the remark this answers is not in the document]
mwcraig: We are — the column is `fit_result.redchi` verbatim (`magnitude_transforms.py` line 875: `fit_redchis[rows] = fit_result.redchi`); nothing is computed by hand. This thread was only ever about the *docstring's description* of that number: it called `redchi` a "mean squared residual", and Copilot's point is that lmfit divides by `nfree = N − n_varied`, not by `N`, so the wording will change to "per degree of freedom". The value itself is untouched.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -286,8 +293,24 @@ def _generate_observed_table(
...
+        Uncertainty to report for each instrumental magnitude. Defaults to
+        ``noise_sigma`` when noise is being added, which is the truthful
+        value -- passing anything else alongside ``noise_sigma`` is how a
+        test says the reported errors are wrong -- and to 0.01 otherwise.
+
+    noise_sigma : float, optional
+        Standard deviation of Gaussian noise added to the instrumental
+        magnitudes. The default of zero adds none, so the observations fit a
+        catalog from `_generate_fake_catalog` essentially exactly. That is
+        what lets the recovery tests demand 1e-6 of the coefficients, but it
```
mwcraig: This eems like an overly wordy docstring...

mwcraig: Trimmed in 0eba95a — the ``noise_sigma`` description is now three lines (recovery tests leave it at zero; diagnostic tests need data that misses the model by a known amount), and the ``mag_error`` description lost its parenthetical.

PR #687, opened 2026-08-11
Propagate the transform's own uncertainty into mag_cal_error

mwcraig:
Closes #674.

Stacked on #682, #683 and #684 — please merge those first; this branch contains
their commits.

`mag_cal_error` was the star's instrumental error scaled by `1 + a` (or `a`
without `fit_diff`) and nothing else. That counts the star's own measurement
error and ignores the uncertainty of the transform it was calibrated *through*,
so the number was known to be too small — and it is the number that goes to the
AAVSO.

It now carries both parts, correlations between the terms included, worked out
per star.

### lmfit does the propagation

`lmfit.Parameters.create_uvars(covar)` returns `uncertainties` values built with
`correlated_values`, so they carry the full covariance.
`calibrated_from_instrumental` — the same model function the fit itself used —
evaluates on them **completely unmodified**.

The nice part: making `mag_inst` itself a ufloat means the `1 + a + 2·b·m`
sensitivity is **never written down**. `uncertainties` derives it from the fact
that `star_mag` appears twice in the expression, once inside the model and once
in the `fit_diff` add-back. So the whole `sensitivity` block disappears rather
than being rewritten, and adding a term to the model keeps the propagation
correct by construction — where a hand-built design matrix would be a second
copy of the model's derivatives, quietly going stale.

The hand-coded `J Σ Jᵀ` numpy version was written and checked against this one
first: they agree to **6.6e-17** over 20 000 Monte Carlo trials. It is recorded
in the commit message as the documented fallback, because the ufloat route costs
~24 µs/star (12 ms at a realistic 500 stars/image, invisible beside the catalog
cross-match, but it is the one thing that would push a future enormous field
back to numpy).

### What the numbers do

50 stars, σ = 0.02, a ≈ 0.023:

| | value |
| --- | --- |
| old `mag_cal_error` (a constant) | 0.020467 |
| new, min / median / max | 0.020802 / 0.021382 / 0.022275 |
| brightest star, end of the range | 0.020467 → **0.021752** (+6.3%) |
| a star in the middle | 0.020467 → **0.021039** (+2.8%) |

The ends-vs-middle excess is 1.35 and stable to three figures across seeds —
that is a fit predicting best at the centroid of the stars it was fit to, and it
is what proves the number is per-star rather than a constant.

Against the diagonal-only version computable from the `*_error` columns, the
reported error is **0.66–0.71×**: `a` and `z` are strongly anticorrelated, so
the off-diagonal terms cut it by ~1.4×. There is a test asserting that
difference, because `create_uvars` degrading to the uncorrelated answer is
exactly the silent failure to guard against.

Monte Carlo, 300 noise realizations, a star not in the fit: `std(mag_cal)` =
0.024379 against a reported 0.023589, **ratio 1.034**. The old formula gave
0.02040 for the same case — understating by ~20%.

### When the covariance is unusable

That image's `mag_cal_error` is NaN and a warning names it. `mag_cal` itself is
still produced. There is deliberately no fallback to the instrumental term
alone: a knowingly-too-small uncertainty is the bug being fixed. This leans on
#684's `covar is None or not finite` verdict — `create_uvars` fails *silently*
three ways on a bad covariance (values 4.2–4.6× too large, or zero, or a finite
uncorrelated answer), never by raising.

### Catalog magnitude uncertainty is deliberately excluded

A star's catalog entry decides whether it was matched but never enters the
calculation of its calibrated magnitude, so there is nothing there to propagate.
What the catalog *does* contribute is its scatter about the transform across the
field — and that is already in the answer, because `scale_covar` rescales the
covariance by `fit_redchi`. That is the right home for it: a property of the
calibration shared by every star in the image. The catalog's systematic tie to
the standard system (~0.02 mag for APASS DR9) is identical for every star in
every image, and in a per-star column would be actively misleading, since
averaging would appear to beat it down by √N.

### Note on `uncertainties` in `pyproject.toml`

Not a new dependency — it is an unconditional install requirement of lmfit and
`uncertainties 3.2.3` is already resolved into every environment in `pixi.lock`.
The declaration is honesty about a direct import, the same argument that added
the explicit `scipy` entry in #676.

---

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -163,6 +163,19 @@ Bug Fixes
   nearest catalog star's values however far away it was. [#678]
 + ``transform_to_catalog()`` no longer requires a star to have a catalog
   color unless a color term (``"c"`` or ``"d"``) is actually being fit. [#681]
++ ``mag_cal_error`` from ``transform_to_catalog()`` is now a real propagation.
+  It used to be the star's own measurement error scaled by ``1 + a``, as though
+  the transform the star was calibrated through were known exactly, and so was
+  too small -- on a fit to a few dozen stars with 0.02 mag scatter, by around
+  20%. It now carries the uncertainty of the fit as well, correlations between
+  the terms included, and varies from star to star because a fit predicts best
+  at the centroid of the stars it was fit to. Reported errors will therefore
+  grow, which matters because these numbers are submitted to the AAVSO. For an
+  image whose fit left no usable covariance behind, ``mag_cal_error`` is now
+  NaN with a warning rather than a value known to be too small; the calibrated
+  magnitudes are still produced. The uncertainty of the catalog magnitudes is
+  deliberately excluded -- see the ``transform_to_catalog()`` docstring for
+  why. [#674]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -163,19 +163,12 @@ Bug Fixes
   nearest catalog star's values however far away it was. [#678]
 + ``transform_to_catalog()`` no longer requires a star to have a catalog
   color unless a color term (``"c"`` or ``"d"``) is actually being fit. [#681]
-+ ``mag_cal_error`` from ``transform_to_catalog()`` is now a real propagation.
-  It used to be the star's own measurement error scaled by ``1 + a``, as though
-  the transform the star was calibrated through were known exactly, and so was
-  too small -- on a fit to a few dozen stars with 0.02 mag scatter, by around
-  20%. It now carries the uncertainty of the fit as well, correlations between
-  the terms included, and varies from star to star because a fit predicts best
-  at the centroid of the stars it was fit to. Reported errors will therefore
-  grow, which matters because these numbers are submitted to the AAVSO. For an
-  image whose fit left no usable covariance behind, ``mag_cal_error`` is now
-  NaN with a warning rather than a value known to be too small; the calibrated
-  magnitudes are still produced. The uncertainty of the catalog magnitudes is
-  deliberately excluded -- see the ``transform_to_catalog()`` docstring for
-  why. [#674]
++ ``mag_cal_error`` from ``transform_to_catalog()`` now includes the
+  uncertainty of the fitted transform, correlations included, instead of
+  only the star's scaled measurement error, so it varies per star and is
+  somewhat larger. When a fit leaves no usable covariance it is NaN, with
+  a warning. Catalog magnitude uncertainty is still excluded -- see the
+  docstring for why. [#674]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,8 +25,9 @@ dependencies = [
                    # to speed up its internal numpy operations when present
     "ccdproc",
     "lightkurve",
-    "lmfit",  # fits the magnitude transform in stellarphot.utils; also the
-              # fitter behind the optional transit fitting
+    "lmfit >=1.2.2",  # fits the magnitude transform in stellarphot.utils; also
+                      # the fitter behind the optional transit fitting; 1.2.2
+                      # introduced Parameters.create_uvars()
     "matplotlib",
     "pandas",
     "photutils >=2,<3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,10 @@ dependencies = [
                 # astroquery.XMatch gripes on import if it is not installed
     "scipy",  # imported directly in stellarphot.photometry; arrives
               # transitively via astropy's dependencies otherwise
+    "uncertainties",  # imported directly in stellarphot.utils to propagate the
+                      # magnitude transform's covariance into each star's
+                      # calibrated error; an unconditional requirement of lmfit
+                      # otherwise
 ]
 
 [project.optional-dependencies]
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -6,6 +6,7 @@ import numpy as np
 from astropy import units as u
 from astropy.coordinates import SkyCoord
 from astropy.utils.exceptions import AstropyUserWarning
+from uncertainties import unumpy
 
 from ..catalogs import apass_dr9, refcat2
 from .magnitude_system_transforms import transform_apass_bands, transform_refcat2_bands
@@ -212,6 +213,145 @@ def _underdetermined_reason(fit_result, vary):
         )
 
     return None
+
+
+def _coefficient_uncertainties(fit_result):
+    """
+    Uncertainty of each coefficient of a fit, when the fit can supply them.
+
+    This is the single place the question "does this fit have a usable
+    covariance?" is answered, so that everything derived from the covariance
+    agrees about whether there is one to derive anything from.
+
+    Note that ``fit_result.errorbars`` is deliberately *not* what is checked.
+    `lmfit` clears that flag whenever any varied term comes out with a
+    standard error of exactly zero, and that is what a perfect fit gives:
+    data that follows the model to the last bit has a covariance that really
+    is zero, and zero is the right answer there rather than a missing one.
+    What makes the uncertainties unavailable is ``covar`` being absent
+    altogether, or having non-finite entries -- an inverse that came back with
+    NaN in it looks like a matrix and is not one, and anything that reads it
+    without checking reports a confident wrong number instead of failing.
+
+    Parameters
+    ----------
+
+    fit_result : `lmfit.minimizer.MinimizerResult`
+        Result of fitting one image.
+
+    Returns
+    -------
+    uncertainties : dict
+        Standard error of each term of the transform model, keyed by term
+        name. A term that was not varied was held at exactly zero and is
+        therefore known exactly, so its uncertainty is ``0.0``. Every value is
+        NaN if the covariance cannot be used.
+
+    covar : `numpy.ndarray` or None
+        The covariance matrix, or `None` if it cannot be used. Anything that
+        needs the correlations between the terms as well as their individual
+        uncertainties should take the matrix from here rather than asking the
+        fit result again, so that there is only ever one verdict.
+    """
+    covar = fit_result.covar
+
+    if covar is None or not np.all(np.isfinite(covar)):
+        return {name: np.nan for name in _COEFF_NAMES}, None
+
+    return {
+        name: (
+            np.nan
+            if fit_result.params[name].stderr is None
+            else float(fit_result.params[name].stderr)
+        )
+        for name in _COEFF_NAMES
+    }, covar
+
+
+def _calibrated_with_uncertainty(fit_result, covar, mag_inst, color, errors, fit_diff):
+    """
+    Calibrated magnitude of every star, and how uncertain each one is.
+
+    The uncertainty has two parts: the star's own measurement error, and the
+    uncertainty of the transform it was calibrated through. Neither one is
+    written down here. `lmfit` turns the fit into `uncertainties` values that
+    carry the whole covariance matrix, including the correlations between the
+    terms, and `calibrated_from_instrumental` -- the same model function the
+    fit itself used -- then evaluates on them unchanged.
+
+    Doing it this way means the sensitivity of the calibrated magnitude to the
+    instrumental one is never spelled out. Making ``mag_inst`` itself an
+    uncertain value is enough: it appears twice in the expression, once inside
+    the model and once in the ``fit_diff`` add-back, and `uncertainties`
+    works out the ``1 + a + 2*b*mag_inst`` that follows from that. Add a term
+    to the model and the propagation stays right by construction, where a
+    hand-written derivative would be a second copy of the model quietly going
+    stale.
+
+    Note that this is deliberately not exact for a star that was itself in the
+    fit: that star's own noise helped move the coefficients, and treating its
+    measurement error and the transform's uncertainty as independent
+    double-counts a little of it. Measured against a Monte Carlo the reported
+    error is about 10% too large in that case, shrinking as one over the
+    number of stars fit. Erring high is the right direction and modelling the
+    correlation is not worth what it would cost.
+
+    Parameters
+    ----------
+
+    fit_result : `lmfit.minimizer.MinimizerResult`
+        Result of fitting one image.
+
+    covar : `numpy.ndarray` or None
+        Covariance matrix of the fit, as vetted by
+        `_coefficient_uncertainties`, or `None` if it cannot be used. Nothing
+        is propagated when it is `None`: the calibrated magnitudes are still
+        worked out, but there is no honest uncertainty to report for them.
+
+    mag_inst, color : `numpy.ndarray`
+        Instrumental magnitude and color of every star in the image, not just
+        the ones the fit used.
+
+    errors : `numpy.ndarray` or None
+        Uncertainty of each instrumental magnitude, or `None` if none was
+        supplied. Must already have been checked: `uncertainties` refuses
+        outright to build a value whose standard deviation is negative. NaN it
+        accepts, and carries through to a NaN uncertainty beside a perfectly
+        good calibrated magnitude.
+
+    fit_diff : bool
+        Whether the fit was of the difference between the catalog and
+        instrumental magnitudes, in which case the instrumental magnitude is
+        added back here.
+
+    Returns
+    -------
+    calibrated : `numpy.ndarray`
+        Calibrated magnitude of each star.
+
+    uncertainty : `numpy.ndarray` or None
+        Uncertainty of each calibrated magnitude, or `None` when there was
+        nothing to propagate.
+    """
+    if covar is None or errors is None:
+        calibrated = calibrated_from_instrumental(
+            (mag_inst, color),
+            *(fit_result.params[name].value for name in _COEFF_NAMES),
+        )
+        if fit_diff:
+            calibrated = calibrated + mag_inst
+        return calibrated, None
+
+    star_mag = unumpy.uarray(mag_inst, errors)
+    uvars = fit_result.params.create_uvars(covar)
+
+    calibrated = calibrated_from_instrumental(
+        (star_mag, color), *(uvars[name] for name in _COEFF_NAMES)
+    )
+    if fit_diff:
+        calibrated = calibrated + star_mag
+
+    return unumpy.nominal_values(calibrated), unumpy.std_devs(calibrated)
 
 
 def _check_known_terms(terms, argument_name):
@@ -550,12 +690,38 @@ def transform_to_catalog(
     a ``fit_redchi`` near four and coefficient uncertainties twice as large,
     and that is one observation rather than two agreeing ones.
 
-    The values in ``mag_cal_error`` are the instrumental errors scaled by how
-    much the calibrated magnitude moves when the instrumental one does, which
-    is ``1 + a`` when ``fit_diff`` is ``True`` and ``a`` when it is ``False``.
-    That is not a propagation of the uncertainty in the fit itself, and so
-    understates the true uncertainty; the ``*_error`` columns are where that
-    uncertainty can be read off in the meantime.
+    ``mag_cal_error`` combines two things: the star's own measurement error,
+    and the uncertainty of the transform it was calibrated through, correlations
+    between the terms included. The second part is the larger of the two on a
+    fit to a modest number of stars, and it is worked out per star rather than
+    added on as a constant, because a fit predicts best at the centroid of the
+    stars it was fit to and worst at the ends of their range. It is NaN for an
+    image whose fit left no usable covariance behind: an uncertainty that
+    cannot be worked out is reported as no number rather than as the
+    measurement error on its own, which would be a confident value known to be
+    too small.
+
+    The uncertainty of the *catalog* magnitudes is deliberately not part of it.
+    A star's catalog entry decides whether it was matched, but never enters the
+    calculation of its calibrated magnitude, so there is nothing there to
+    propagate. What the catalog does contribute is its scatter about the
+    transform across the field, and that is already in the answer: ``lmfit``
+    scales the covariance by ``fit_redchi``, so catalog scatter inflates every
+    coefficient uncertainty and with it every calibrated error. That is the
+    right home for it, a property of the calibration shared by every star in
+    the image. The catalog's systematic tie to the standard system -- about
+    0.02 mag for APASS DR9 -- is identical for every star in every image, and
+    in a per-star column would be actively misleading, since averaging these
+    magnitudes would appear to beat it down by the square root of the number of
+    stars.
+
+    One consequence of that scaling is worth knowing. The transform half of
+    ``mag_cal_error`` corrects itself when the quoted instrumental errors are
+    systematically wrong, because the covariance it comes from was scaled by
+    the scatter actually observed; the measurement half does not, because it is
+    those quoted errors. So an image with a ``fit_redchi`` far from one is
+    reporting calibrated errors whose two halves disagree about how much to
+    trust the input, which is the reason to read that column.
     """
     if obs_error_column is None:
         warnings.warn(
@@ -777,18 +943,16 @@ def transform_to_catalog(
 
         values = {name: fit_result.params[name].value for name in _COEFF_NAMES}
 
-        # How well the fit pinned each term down. lmfit's stderr is exactly
-        # 0.0 for a term that was not varied -- held at zero, so known
-        # exactly -- and None in the states where no uncertainty could be
-        # worked out at all.
-        uncertainties = {
-            name: (
-                np.nan
-                if fit_result.params[name].stderr is None
-                else float(fit_result.params[name].stderr)
-            )
-            for name in _COEFF_NAMES
-        }
+        # How well the fit pinned each term down, and how well the model it
+        # landed on describes the stars. Both are answers about the image
+        # rather than about any one star, and both are reported even when
+        # they are the only thing that would tell a caller the coefficients
+        # below are not worth applying.
+        # The covariance matrix comes back too, and ``covar is None`` is
+        # this function's single verdict on whether there is a usable one:
+        # the calibrated errors below are propagated through it, or are
+        # not reported at all.
+        uncertainties, covar = _coefficient_uncertainties(fit_result)
 
         # The expected ranges are a check on the answer, not a constraint on
         # the fit, so a value outside its range is reported and kept.
@@ -801,14 +965,32 @@ def transform_to_catalog(
         if out_of_range:
             logger.warning(f"Fit for {file[0]}: " + "; ".join(out_of_range))
 
+        if obs_error_column is None:
+            star_errors = None
+        else:
+            # An error that is not a positive, finite number is kept out of
+            # the fit above, and must not come back out as a calibrated
+            # error either: the AAVSO writer turns only non-finite errors
+            # into "na", so a zero would be submitted as a real
+            # uncertainty. Done here, before the propagation rather than
+            # after it, because `uncertainties` refuses outright to build a
+            # value with a negative standard deviation.
+            star_errors = np.where(np.isfinite(errors) & (errors > 0), errors, np.nan)
+
+            if covar is None:
+                logger.warning(
+                    f"Fit for {file[0]} left no usable covariance behind, so "
+                    "the uncertainty of the transform cannot be propagated "
+                    "and mag_cal_error is NaN for this image. Reporting the "
+                    "measurement error on its own would understate it."
+                )
+
         # Calculate calibrated magnitudes for every star in the image, not
-        # just the ones the fit used.
-        model_coefficients = [values[name] for name in _COEFF_NAMES]
-        cal_mag = calibrated_from_instrumental(
-            (mag_inst, model_color), *model_coefficients
+        # just the ones the fit used, propagating the uncertainty of the
+        # fit into them when there is one to propagate.
+        cal_mag, cal_error = _calibrated_with_uncertainty(
+            fit_result, covar, mag_inst, model_color, star_errors, fit_diff
         )
-        if fit_diff:
-            cal_mag = cal_mag + mag_inst
 
         # One cut for everything that comes from the match: a star matched
         # closely enough to be given a calibrated magnitude keeps the
@@ -829,23 +1011,7 @@ def transform_to_catalog(
         cat_mags[rows] = np.where(matched, cat_mag, np.nan)
         cat_colors[rows] = np.where(matched, color, np.nan)
 
-        if obs_error_column is not None:
-            # How much the calibrated magnitude moves when the instrumental
-            # one does. With fit_diff the instrumental magnitude is added
-            # back to the model result, so that sensitivity is 1 + a;
-            # without it the model is the calibrated magnitude outright and
-            # a itself is the factor -- it fits to about 1 rather than
-            # about 0. Using the same factor for both would double the
-            # reported uncertainty in one of the two modes.
-            sensitivity = 1 + values["a"] if fit_diff else values["a"]
-            cal_error = sensitivity * errors
-
-            # An error that is not a positive, finite number is kept out of
-            # the fit above, and must not come back out as a calibrated
-            # error either: the AAVSO writer turns only non-finite errors
-            # into "na", so a zero would be submitted as a real uncertainty.
-            cal_error[~(np.isfinite(errors) & (errors > 0))] = np.nan
-
+        if cal_error is not None:
             # A star with no calibrated magnitude has no calibrated error
             # either, whatever the reason it has no magnitude. That is
             # stronger than repeating the distance cut -- it also covers a
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -7,7 +7,8 @@ import numpy as np
 from astropy import units as u
 from astropy.coordinates import SkyCoord
 from astropy.utils.exceptions import AstropyUserWarning
-from uncertainties import unumpy
+from numpy.linalg import LinAlgError
+from uncertainties import correlated_values, unumpy
 
 from ..catalogs import apass_dr9, refcat2
 from .magnitude_system_transforms import transform_apass_bands, transform_refcat2_bands
@@ -220,20 +221,6 @@ def _coefficient_uncertainties(fit_result):
     """
     Uncertainty of each coefficient of a fit, when the fit can supply them.
 
-    This is the single place the question "does this fit have a usable
-    covariance?" is answered, so that everything derived from the covariance
-    agrees about whether there is one to derive anything from.
-
-    Note that ``fit_result.errorbars`` is deliberately *not* what is checked.
-    `lmfit` clears that flag whenever any varied term comes out with a
-    standard error of exactly zero, and that is what a perfect fit gives:
-    data that follows the model to the last bit has a covariance that really
-    is zero, and zero is the right answer there rather than a missing one.
-    What makes the uncertainties unavailable is ``covar`` being absent
-    altogether, or having non-finite entries -- an inverse that came back with
-    NaN in it looks like a matrix and is not one, and anything that reads it
-    without checking reports a confident wrong number instead of failing.
-
     Parameters
     ----------
 
@@ -253,10 +240,33 @@ def _coefficient_uncertainties(fit_result):
         needs the correlations between the terms as well as their individual
         uncertainties should take the matrix from here rather than asking the
         fit result again, so that there is only ever one verdict.
+
+    Notes
+    -----
+
+    This is the single place the question "does this fit have a usable
+    covariance?" is answered, so that everything derived from the covariance
+    agrees about whether there is one to derive anything from.
+
+    ``fit_result.errorbars`` is deliberately *not* what is checked. This code
+    consumes the covariance matrix itself, so the matrix is what it checks;
+    ``errorbars`` is a summary flag `lmfit` derives from that same object, and
+    checking the artifact actually used is one less indirection.
     """
     covar = fit_result.covar
 
     if covar is None or not np.all(np.isfinite(covar)):
+        return {name: np.nan for name in _COEFF_NAMES}, None
+
+    # lmfit's own `create_uvars` wraps this same decomposition in
+    # `except (LinAlgError, ValueError): pass` and falls back silently to
+    # uncorrelated, stderr-based ufloats. Not known to happen with real data;
+    # checked only so a covariance that fails to decompose (e.g. a tiny
+    # negative eigenvalue from round-off) is never turned into nonsense.
+    varied = [name for name in _COEFF_NAMES if fit_result.params[name].vary]
+    try:
+        correlated_values([fit_result.params[name].value for name in varied], covar)
+    except (LinAlgError, ValueError):
         return {name: np.nan for name in _COEFF_NAMES}, None
 
     return {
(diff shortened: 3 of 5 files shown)


CHANGES.rst
```diff
@@ -163,6 +163,19 @@ Bug Fixes
   nearest catalog star's values however far away it was. [#678]
 + ``transform_to_catalog()`` no longer requires a star to have a catalog
   color unless a color term (``"c"`` or ``"d"``) is actually being fit. [#681]
++ ``mag_cal_error`` from ``transform_to_catalog()`` is now a real propagation.
```
mwcraig: This is way, way too wordy. In the future be more concise when writing changelog entriees.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -214,6 +216,177 @@ def _underdetermined_reason(fit_result, vary):
...
 
+def _coefficient_uncertainties(fit_result):
+    """
+    Uncertainty of each coefficient of a fit, when the fit can supply them.
+
+    This is the single place the question "does this fit have a usable
+    covariance?" is answered, so that everything derived from the covariance
+    agrees about whether there is one to derive anything from.
+
+    Note that ``fit_result.errorbars`` is deliberately *not* what is checked.
```
mwcraig: The fact that the docstring starts with the word "Note" is a strong hint that it should be in the notes section. The paragraph that starts on 223 should move there too.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -214,6 +216,177 @@ def _underdetermined_reason(fit_result, vary):
...
+    `lmfit` clears that flag whenever any varied term comes out with a
+    standard error of exactly zero, and that is what a perfect fit gives:
```
mwcraig: But the standard error will never be zero for a real fit....

stellarphot/utils/magnitude_transforms.py
```diff
@@ -214,6 +216,177 @@ def _underdetermined_reason(fit_result, vary):
...
+    with warnings.catch_warnings(), np.errstate(invalid="ignore"):
+        warnings.filterwarnings("ignore", "invalid value encountered", RuntimeWarning)
+        yield
+
+
+def _calibrated_with_uncertainty(fit_result, covar, mag_inst, color, errors, fit_diff):
+    """
+    Calibrated magnitude of every star, and how uncertain each one is.
+
+    The uncertainty has two parts: the star's own measurement error, and the
```
mwcraig: This extended description belongs in notes and can be shortened.

stellarphot/utils/magnitude_transforms.py
mwcraig, line 701: Can you shrink this section?

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -624,6 +624,68 @@ def _fit_a_catalog(mocker, n_stars=20, vary=None, **catalog_coefficients):
     return result, catalog, instrumental
 
 
+def _fit_a_noisy_catalog(
```
mwcraig: Can this be combined with `_fit_a_catalog`

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -867,6 +921,79 @@ def test_transform_to_catalog_reports_unweighted_fit_statistic(mocker):
     assert reported[0] == pytest.approx(sigma**2, rel=0.3)
 
 
+def _break_the_covariance(mocker, how):
+    """
+    Make the fit come back with a covariance matrix that cannot be used.
+
+    Neither state is reachable from data. ``lmfit.minimize`` only leaves
```
mwcraig: "Neither state is reachable from data." -- so why is it showing up in the tests? We don't need to test for situations that cannot be reached by data.

pyproject.toml
```diff
@@ -40,6 +40,10 @@ dependencies = [
                 # astroquery.XMatch gripes on import if it is not installed
     "scipy",  # imported directly in stellarphot.photometry; arrives
               # transitively via astropy's dependencies otherwise
+    "uncertainties",  # imported directly in stellarphot.utils to propagate the
```
mwcraig, line 44: While touching the dependency list: the new code calls `Parameters.create_uvars()` (`magnitude_transforms.py:374`), which only exists in **lmfit >= 1.2.2**, but the `lmfit` entry above has no minimum version. An environment where something else pins an older lmfit (it is unpinned and pure Python, so it resolves fine) will run the fit successfully and then crash with `AttributeError: 'Parameters' object has no attribute 'create_uvars'` — a regression from before this PR, when that environment worked. Suggest pinning `lmfit>=1.2.2` here.

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/magnitude_transforms.py
```diff
@@ -214,6 +216,177 @@ def _underdetermined_reason(fit_result, vary):
...
+        with _nan_is_an_expected_input():
+            calibrated = calibrated_from_instrumental(
+                (mag_inst, color),
+                *(fit_result.params[name].value for name in _COEFF_NAMES),
+            )
+            if fit_diff:
+                calibrated = calibrated + mag_inst
+        return calibrated, None
+
+    uvars = fit_result.params.create_uvars(covar)
```
mwcraig, line 380: The "single verdict" from `_coefficient_uncertainties` can be silently overruled here: lmfit's `create_uvars` wraps `uncertainties.correlated_values(vbest, covar)` in `except (LinAlgError, ValueError): pass` and falls back to **uncorrelated** stderr-based ufloats. A finite covariance with a numerically negative diagonal entry (pinv round-off on a marginal but condition-number-passing fit) sails through the all-finite check, then fails decomposition inside `create_uvars` — and the propagation quietly degrades to diagonal-only, which the whole-covariance test in this PR measures as 1.4–1.8x too large, or produces NaN stderr with no warning. That is exactly the confident-wrong-number outcome the covar-unusable path is designed to prevent.

Suggested fix in the arbiter: have `_coefficient_uncertainties` attempt `uncertainties.correlated_values` itself (or at least check `np.diag(covar) >= 0`) and route failure to the existing covar-unusable path, so this call can never hit lmfit's silent fallback.

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/magnitude_transforms.py
```diff
@@ -214,6 +216,177 @@ def _underdetermined_reason(fit_result, vary):
...
+    -------
+    calibrated : `numpy.ndarray`
+        Calibrated magnitude of each star.
+
+    uncertainty : `numpy.ndarray` or None
+        Uncertainty of each calibrated magnitude, or `None` when there was
+        nothing to propagate.
+    """
+    if covar is None or errors is None:
+        with _nan_is_an_expected_input():
```
mwcraig: This fallback branch evaluates the model on plain floats, but wrapping it in `_nan_is_an_expected_input()` also applies `np.errstate(invalid="ignore")`. On main this exact computation surfaced an `invalid value encountered` RuntimeWarning for genuinely bad photometry — e.g. a star with zero net counts gives `mag_inst = +inf`, and the held-at-zero `b` term computes `inf * 0`. That diagnostic pointed at the bad input; now it is silently swallowed and the user just gets NaN `mag_cal`. The context manager's own docstring says "a genuinely invalid operation on a real number is not what this covers" — suggest keeping the suppression only around the object-array propagation path and letting this plain-float evaluation warn.

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/magnitude_transforms.py
```diff
@@ -550,12 +723,38 @@ def transform_to_catalog(
...
+    fit to a modest number of stars, and it is worked out per star rather than
+    added on as a constant, because a fit predicts best at the centroid of the
+    stars it was fit to and worst at the ends of their range. It is NaN for an
+    image whose fit left no usable covariance behind: an uncertainty that
+    cannot be worked out is reported as no number rather than as the
+    measurement error on its own, which would be a confident value known to be
+    too small.
+
+    The uncertainty of the *catalog* magnitudes is deliberately not part of it.
+    A star's catalog entry decides whether it was matched, but never enters the
```
mwcraig: This docstring is not quite right, and it papers over a real omitted error term: a star's matched catalog entry *does* enter the calculation of its own calibrated magnitude — the model applies `c * color + d * color**2` using that star's catalog color. The star's catalog-color uncertainty (0.05–0.1 mag for faint APASS/refcat2 stars) is therefore a real contribution of roughly `c * sigma_color` (order 0.005–0.02 mag) that `mag_cal_error` omits, and the redchi scaling only absorbs the *fit* stars' scatter, not the target star's own color error. Leaving the term unpropagated may be a fine trade for now, but the docstring should say the color-error term is knowingly omitted rather than claiming there is nothing to propagate.

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/magnitude_transforms.py
```diff
@@ -801,14 +998,32 @@ def transform_to_catalog(
...
+            # An error that is not a positive, finite number is kept out of
+            # the fit above, and must not come back out as a calibrated
+            # error either: the AAVSO writer turns only non-finite errors
+            # into "na", so a zero would be submitted as a real
+            # uncertainty. Done here, before the propagation rather than
+            # after it, because `uncertainties` refuses outright to build a
+            # value with a negative standard deviation.
+            star_errors = np.where(np.isfinite(errors) & (errors > 0), errors, np.nan)
+
+            if covar is None:
```
mwcraig: This warning only fires inside the `obs_error_column is not None` branch, but the coefficient `*_error` columns are NaNed for a broken covariance regardless of weighting. An unweighted fit whose covariance is unusable gets all-NaN `a_error`/`c_error`/`z_error` with no log line explaining why. Consider hoisting the covar-is-None warning out of the error-column branch and wording it to cover the coefficient errors too (with the `mag_cal_error` sentence kept conditional).

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1207,8 +1511,13 @@ def test_transform_to_catalog_fit_diff_agrees(mocker):
     np.testing.assert_allclose(
```
mwcraig, line 1418: Flagging what this loosening gives up: main enforced that the two `fit_diff` modes agree to `atol=1e-8`; a future regression that makes them disagree at, say, the 5e-8 level (a sensitivity term applied in one mode but not the other, scaled small on this fixture) now passes. The measured 4e-10 with a factor-25 margin supports the new bound, so this reads as a deliberate trade — but a brief comment here recording the measured value would keep the next reader from tightening or loosening it blind.

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -867,6 +921,79 @@ def test_transform_to_catalog_reports_unweighted_fit_statistic(mocker):
...
+
+    mocker.patch.object(
+        magnitude_transforms.lmfit,
+        "minimize",
+        side_effect=minimize_then_spoil_covariance,
+    )
+
+
+@pytest.mark.parametrize("how", ["missing", "not_finite"])
+def test_transform_to_catalog_nans_uncertainties_without_a_covariance(
```
mwcraig: This test and `test_transform_to_catalog_nans_calibrated_error_without_a_covariance` (line ~1319) are the same scenario split in two: both parametrize `how` over `missing`/`not_finite`, both call `_break_the_covariance(mocker, how)` then `_fit_a_noisy_catalog(..., a=0.02, c=0.15)` (differing only in an arbitrary seed), and each asserts a disjoint slice of one outcome — coefficient `*_error` NaN + finite coefficients here, `mag_cal_error` NaN + warning there — held together by cross-referencing comments. Merging them into one test with both assertion groups halves the full 50-star fit pipeline runs (4 → 2) with identical coverage, and removes the risk of the twins drifting apart.

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: `_fit_a_noisy_catalog` is a near-verbatim copy of `_fit_a_catalog` — the only difference is passing `noise_sigma=sigma, seed=seed` to `_generate_observed_table`; the rest, including ~50 lines of parallel docstring, is duplicated. The next change to how observations are built has to be made twice, and a fix applied to one helper leaves half the suite on a different fixture path. Suggest adding `sigma=0.0, seed=None` parameters to `_fit_a_catalog` and deleting this helper (callers pass `sigma=0.02` explicitly).

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1032,6 +1166,176 @@ def test_transform_to_catalog_error_is_scaled_instrumental_error(mocker):
     )
 
 
+def test_transform_to_catalog_error_includes_the_transform_uncertainty(mocker):
```
mwcraig: This test, `test_transform_to_catalog_reports_coefficient_uncertainties` (~line 799), and `test_transform_to_catalog_error_uses_the_whole_covariance` (~line 1204) all run the byte-identical fit — `_fit_a_noisy_catalog(mocker, a=0.02, c=0.15)` with the default `n_stars=50, sigma=0.02, seed=20260811` — so the full catalog-match-plus-lmfit pipeline runs three times to produce the same table. Computing it once (module-scoped fixture, or one test with the three assertion blocks) cuts that runtime to a third and leaves one place to reason about the shared scenario.

*(Comment written by Claude at Matts direction.)*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1032,6 +1166,176 @@ def test_transform_to_catalog_error_is_scaled_instrumental_error(mocker):
...
+    # How much the calibrated magnitude moves when the instrumental one does
+    # is 1 + a + 2*b*mag_inst, not 1 + a, whenever the quadratic term is being
+    # fit -- and it is different for every star. Nothing writes that
+    # expression down: it falls out of the instrumental magnitude appearing
+    # twice in the model, once inside it and once in the fit_diff add-back,
+    # which is exactly why it needs a test of its own. On this catalog the fit
+    # is perfect, so the transform term is negligible and the sensitivity is
+    # the whole answer. See issue #674.
+    a, b = 0.02, 0.01
+    mag_error = 0.01
```
mwcraig, line 1222: `mag_error = 0.01` here silently restates the implicit default inside `_generate_observed_table` — nothing connects this local constant to the helper value it must match, so if the default ever changes the expected values are computed from a stale number. Suggest passing `mag_error` explicitly through the helper (here and in the perfect-fit test at ~line 1160 that set the pattern).

*(Comment written by Claude at Matts direction.)*

CHANGES.rst
mwcraig: Not yet fixed — the entry on the branch is still the long version. Agreed, and noted as standing guidance for future entries. Proposed replacement:

```rst
+ ``mag_cal_error`` from ``transform_to_catalog()`` now includes the
  uncertainty of the fitted transform, correlations included, instead of
  only the star's scaled measurement error, so it varies per star and is
  somewhat larger. When a fit leaves no usable covariance it is NaN, with
  a warning. Catalog magnitude uncertainty is still excluded -- see the
  docstring for why. [#674]
```

Say the word and I'll apply it on the branch.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed, and not yet fixed — on the branch this is all still in the extended summary with no ``Notes`` section. The plan: keep the one-line summary, move the "single place the question is answered" paragraph and the ``errorbars`` paragraph into a numpydoc ``Notes`` section, with the ``errorbars`` paragraph reworded per the discussion in the other thread on this docstring. Happy to make the edit.

stellarphot/utils/magnitude_transforms.py
mwcraig: You're right about real data, and the docstring leads with the wrong justification. Checked against installed lmfit (1.3.4) source: the flag clears via `result.errorbars = result.errorbars and (par.stderr > 0.0)` (minimizer.py:822), so stderr exactly `0.0` is what it takes — which noise-free synthetic data produces, but a real noisy fit essentially never will.

The load-bearing reason to check `covar` directly is the *other* direction: for `leastsq`, lmfit sets `errorbars = (_cov is not None)` and never looks at the entries. A covariance carrying NaN — the #677 shape, which `create_uvars` silently turns into confident wrong numbers — sails through with `errorbars=True`. The code consumes the matrix itself, so the matrix is what gets checked; `errorbars` is a proxy that is both too strict (zero stderr) and too loose (non-finite entries).

Suggest rewording to lead with the NaN direction and demote the perfect-fit case to a parenthetical about the noiseless test fixtures. Happy to make that edit.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed, not yet done — still all in the extended summary on the branch. Plan: cut the summary to the two-part statement (measurement error + transform uncertainty, correlations included), and move the mechanism (ufloats carry the covariance, the sensitivity never being written down, the deliberate slight over-count for stars in the fit) into a trimmed ``Notes`` section. Happy to make the edit.

stellarphot/utils/magnitude_transforms.py
mwcraig: Not yet shrunk — the section is still ~24 lines on the branch (this thread only shows as outdated because it was anchored to an intermediate commit's diff). Will compress the two paragraphs to roughly half.

One caveat before shrinking: the second paragraph's claim that a star's catalog entry "never enters the calculation" is factually wrong whenever a color term is fit — the model applies `c * color + d * color**2` using that star's catalog color (see the thread at line 738). The shrink should land together with that correction rather than preserving the sentence. Happy to do both in one edit.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Yes — and not yet done. The bodies differ only in passing `noise_sigma=sigma, seed=seed` to `_generate_observed_table`, plus defaults (`n_stars` 50 vs 20); the ~50-line docstring is near-duplicated. Plan: add `sigma=0.0, seed=None` (and a `mag_error` passthrough, per the thread at line ~1300) to `_fit_a_catalog`, delete `_fit_a_noisy_catalog`, and have the noisy callers pass `n_stars=50, sigma=0.02` and their seed explicitly. Same merge was requested in the other thread on this line — one commit covers both. Happy to implement.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Pushing back on the premise, with one concession: the *docstring* is what's wrong here, not the tests.

The `"missing"` state is reachable from data. The helper's claim that lmfit only loses `covar` "when inverting raises" overstates — for `leastsq`, lmfit leaves `result.covar` as `None` whenever scipy returns `cov_x=None` (minimizer.py:1712–1720 simply skips the assignment and appends "Could not estimate error-bars."), and scipy does that for an exactly rank-deficient Jacobian, no exception involved. That is constructible from data: if every matched star has the same catalog color (catalog magnitudes are rounded, small fields match few stars), the `c` column is an exact multiple of the `z` column and the Jacobian is singular. The upstream degeneracy check guards the star *count*, not the rank.

The `"not_finite"` state I can't point to a data trigger for. It's there because a NaN-carrying covariance is the one shape `create_uvars` converts into confidently wrong numbers without raising (#677/#684), and the guard plus its test is the cost of never finding out via an AAVSO submission.

So: keep the tests, fix the helper docstring to say the mock is for determinism (provoking exact rank deficiency through the public API would be a brittle fixture) rather than claiming the states are fictional. Can drop the `"not_finite"` parametrization if you'd rather trust lmfit's post-processing. Happy to make either edit.

pyproject.toml
mwcraig: Status check: still applies, not yet addressed — `lmfit` is still unpinned on the branch. Re-verified: `create_uvars` is present in the installed lmfit 1.3.4, and the upstream 1.2.2 release notes list it as new in that release. The fix is one line: `"lmfit>=1.2.2",`. Happy to push it.

stellarphot/utils/magnitude_transforms.py
mwcraig: Status check: confirmed against installed lmfit 1.3.4 source, and not yet addressed. `parameter.py` has exactly the silent fallback described:

```python
if covar is not None:
    try:
        corr_uvars = correlated_values(vbest, covar)
        for name, cuv in zip(vnames, corr_uvars):
            uvars[name] = cuv
    except (LinAlgError, ValueError):
        pass
```

and `_coefficient_uncertainties` on the branch still checks only `covar is None or not np.all(np.isfinite(covar))`, so a finite but non-decomposable matrix reaches the fallback and the propagation quietly degrades to diagonal-only. The fix stands as proposed: attempt `uncertainties.correlated_values` inside the arbiter and route failure to the existing unusable-covariance path. Happy to implement, with a `_break_the_covariance` case to pin it.

stellarphot/utils/magnitude_transforms.py
mwcraig: Status check: still applies, not yet addressed. On the branch the `covar is None or errors is None` branch wraps the plain-float evaluation in `_nan_is_an_expected_input()`. The suppression buys nothing on that path — a NaN *input* propagates through float arithmetic without tripping the `invalid` flag; only creating NaN from non-NaN does (e.g. `inf * 0` from a star with zero net counts and a held-at-zero term) — so its only effect there is swallowing the genuine bad-photometry diagnostic that main surfaced. Fix: drop the context manager from this branch and keep it around the object-array propagation, matching what the manager's own docstring says it covers. Happy to implement.

stellarphot/utils/magnitude_transforms.py
mwcraig: Status check: confirmed against the code, not yet addressed. `color` is the matched star's catalog color (`cat["color"][cat_idx]`, built from catalog magnitudes) and the model applies `c * color + d * color**2` with it per star — so the docstring's "never enters the calculation" is wrong whenever a color term is fit, and that star's catalog-color error is a real, knowingly omitted term. Suggest folding the correction into the shrink Matt asked for on this section (thread at line ~695): state the omission plainly instead of arguing there is nothing to propagate. Happy to write that edit.

stellarphot/utils/magnitude_transforms.py
mwcraig: Status check: still applies, not yet addressed. On the branch the `if covar is None:` warning sits inside the `obs_error_column is not None` else-branch, and the coefficient `*_error` columns are initialized to NaN and simply never filled when the covariance is unusable — so an unweighted fit gets all-NaN `a_error`/`c_error`/`z_error` with no log line. Fix as proposed: hoist the warning out of the error-column branch, word it around the coefficient uncertainties, and keep the `mag_cal_error` sentence conditional on weighting. Happy to implement.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Status check: already satisfied — no change needed. The branch has exactly the requested comment directly above the loosened assertion:

```python
    # Looser than the magnitudes above, and deliberately so: the two modes fit
    # different numbers, so their covariances differ in the last few bits --
    # measured at 1.5e-11 -- and the propagation carries that into the errors
    # at about 4e-10. That clears 1e-8 by only a factor of 25, close enough to
    # the edge to be worth not waiting for.
```

The `mag_cal` assertion above it keeps main's `atol=1e-6` unchanged; only the error tolerance moved (1e-8 → 1e-7), and it's the one the comment documents. This thread can be resolved as-is.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Status check: confirmed, not yet addressed. Both tests still parametrize `how` over `missing`/`not_finite`, break the covariance the same way, and run the same `a=0.02, c=0.15` fit — the only difference is the seed (246810 vs the default 20260811), which nothing in either test depends on. Merging into one test with both assertion groups (coefficient `*_error` NaN + finite coefficients + `mag_cal_error` NaN + both warning checks) halves the full-pipeline runs from 4 to 2 with identical coverage. Happy to implement.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Status check: confirmed, not yet addressed — and Matt asked for the same merge in the other thread on this line, so one fix closes both. Plan: `_fit_a_catalog` grows `sigma=0.0, seed=None` (and a `mag_error=None` passthrough, per the line-~1300 thread), `_fit_a_noisy_catalog` is deleted, and noisy callers pass `n_stars=50, sigma=0.02` and their seed explicitly. Happy to implement.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Status check: confirmed, not yet addressed. All three tests run the byte-identical fit — `a=0.02, c=0.15` with `_fit_a_noisy_catalog` defaults `n_stars=50, sigma=0.02, seed=20260811` (two of them restate the defaults locally). pytest-mock's `module_mocker` makes a module-scoped fixture straightforward for the shared healthy-fit result (the broken-covariance tests keep their own function-scoped mocks); alternatively, one test with three assertion blocks. Either cuts the three pipeline runs to one. Happy to implement.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Status check: confirmed, not yet addressed. In both this test and the perfect-fit test at ~line 1160, the local `mag_error = 0.01` is never passed to the helper — it silently matches the `0.01` fallback inside `_generate_observed_table` (`mag_error = noise_sigma if noise_sigma else 0.01`). Fix folds naturally into the helper merge requested at line 627: give `_fit_a_catalog` a `mag_error` passthrough and have both tests hand it their local constant. Happy to implement.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: "That is constructible from data: if every matched star has the same catalog color (catalog magnitudes are rounded, small fields match few stars), the c column is an exact multiple of the z column and the Jacobian is singular. " -- that does not happen with real data matched to unique stars, and stellarphot ensures that its photometry never happens on stars 1 arcsec apart, and sources in APASS and refcat2 are more widely spaced than that, so there will never be several matches to the same star at the 1 arcsec limit for inclusion in the fitting set.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -214,6 +216,177 @@ def _underdetermined_reason(fit_result, vary):
     return None
 
 
+def _coefficient_uncertainties(fit_result):
```
mwcraig, line 220: Why is this a function?

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Conceded on the data. To clarify what I meant first: not several matches to the same catalog star — distinct stars whose *color values* coincide exactly after catalog rounding. But for the Jacobian to be singular, *every* star in the fit would need the identical rounded color, and with mmag rounding and any real color spread across a field that's effectively probability zero. So you're right: no realistic catalog produces `covar=None` here.

What's left is a contract argument, not a data one: lmfit's own `leastsq` path has a branch that leaves `covar` as `None` ("Could not estimate error-bars."), so the attribute is documented as nullable regardless of what our data can trigger — and the guard's `covar is None` arm is also what keeps the check itself from crashing (`np.isfinite(None)` raises). Two lines of defense against an API-documented state, plus a mock test so the branch isn't shipped untested.

So the concrete proposal for this thread: reword the helper docstring to say "not reachable from realistic catalog data; handled because lmfit documents that ``covar`` may be None, and arranged by mock because data cannot arrange it" — and I'll keep or drop the `"not_finite"` case at your call. Happy to push that edit.

stellarphot/utils/magnitude_transforms.py
mwcraig: It has one call site, so the honest answer is organization, not reuse:

1. **It fixes a contract between two consumers.** Its return bundles the `*_error` values with the covariance-or-None verdict, and that same `covar` object is then handed to `_calibrated_with_uncertainty`, whose docstring specifies it as "the covariance from `_coefficient_uncertainties`, or `None` if it cannot be used". Inlined, the usability decision and the two things that branch on it become three separate stretches of the loop body that have to stay in agreement by hand.
2. **The loop body it would join is already the longest stretch of a very long function.** The name plus docstring keeps the `errorbars`-vs-`covar` rationale (~12 lines of why) out of the loop, where it would otherwise sit as a comment block.
3. **It matches the file's existing pattern**: `_underdetermined_reason` directly above is the same shape — a single-call-site, per-fit question factored out and named.

That said, the body is only ~10 lines; if you'd rather see it inlined with the rationale compressed into a comment, that works too. Say which and I'll do it.

stellarphot/utils/magnitude_transforms.py
mwcraig: "A covariance carrying NaN — the #677 shape, which create_uvars silently turns into confident wrong numbers — sails through with errorbars=True. " -- can this happen on real data?

stellarphot/utils/magnitude_transforms.py
mwcraig: Explain how this can happen in real data

stellarphot/utils/magnitude_transforms.py
mwcraig: Why not propagate the color error as parat of this PR?

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Let's just not worry about it -- if this happens on someone's real data they will report the bug and we'll figure it out then.

stellarphot/utils/magnitude_transforms.py
mwcraig: I'm ok with keeping it for keeping the code complexity down -- note that another of your scomments indicates this is not actually the single source of truth though.

stellarphot/utils/magnitude_transforms.py
mwcraig: Honestly: no — not that I can demonstrate. Every real-data path I can construct through `leastsq` returns either a finite covariance or `None`; the NaN-carrying shape came out of #684's fault injection, not out of data. So the docstring shouldn't lean on it. The defensible core is just: this code consumes the matrix itself, so the matrix is what it checks — `errorbars` is a summary flag lmfit derives from the same object, and checking the artifact you actually use is one less indirection to reason about. I'll cut both corner-case justifications (the perfect-fit story and the NaN story) when moving this into ``Notes``.

stellarphot/utils/magnitude_transforms.py
mwcraig: The mechanism: a covariance that is all-finite but numerically not positive definite. The realistic road there is an ill-conditioned fit — matched stars spanning a very narrow color range with both `c` and `z` varied makes the columns nearly collinear, and round-off in forming/inverting the normal equations can leave a tiny negative eigenvalue. `correlated_values` then fails to decompose it, and lmfit's `except: pass` quietly hands back uncorrelated values.

The honest caveat: I cannot point to a recorded real-data case, and whether a given marginal fit trips it is round-off luck — so under the rule you set in the `_break_the_covariance` thread, this can be dropped too. The one asymmetry worth naming before it is: when `covar` is `None` the failure is loud (NaN column plus a warning), but when this one fires the failure is a confidently wrong, diagonal-only number in an AAVSO submission — the same silent-wrong class this PR exists to fix — and the insurance is a single try/except in `_coefficient_uncertainties`. Your call; happy either way.

stellarphot/utils/magnitude_transforms.py
mwcraig: Two real blockers, one scope call:

1. **The numbers aren't in the room.** `transform_to_catalog` never receives per-star catalog magnitude errors — nothing in the function reads any catalog error column today. Getting them here means plumbing error columns through every supported catalog's column conventions, which is exactly the surface #680 (real catalogs and bands) is reshaping right now; doing it in this PR would collide with that work.
2. **The correlation question needs an answer first.** For stars that were in the fit, their color errors already moved the coefficients; adding the same error per star as if independent double-counts it. The docstring already flags this trade for the measurement term, where it's ~10% and tolerated — for faint-star color errors (0.05–0.1 mag) the double-count would be bigger, so "treat as independent" needs to be justified or corrected, not just assumed.

The mechanics, by contrast, are genuinely one line once the numbers exist — `unumpy.uarray(color, color_error)` and the existing machinery propagates it — which is the argument for doing it as a small follow-up on top of this PR rather than inside it. Happy to file the follow-up issue with that sketch.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed. Concretely that means: delete `_break_the_covariance` and the two tests built on it. In `_coefficient_uncertainties` I'd keep the `covar is None` arm — without it the code crashes on lmfit's documented `None` instead of reporting the bug you'd want reported — and the `isfinite` arm can go or stay at your preference (it's six words). Codecov will flag the now-untested `None` branch; that's the accepted cost of the decision. Will fold this into the fix batch.

stellarphot/utils/magnitude_transforms.py
mwcraig: Correct — as written, `create_uvars` can quietly re-answer the question downstream (the line-374 thread), so the "single place" claim is aspirational rather than true. Two consistent ways to close the gap: add the `correlated_values` attempt to this function (makes the claim true), or soften the sentence to what it actually is — the one place the *lmfit result* is judged. Given your call in the `_break_the_covariance` thread, softening is the consistent choice unless you want the try/except; I'll do whichever alongside the ``Notes`` rework.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -550,12 +723,38 @@ def transform_to_catalog(
...
-    much the calibrated magnitude moves when the instrumental one does, which
-    is ``1 + a`` when ``fit_diff`` is ``True`` and ``a`` when it is ``False``.
-    That is not a propagation of the uncertainty in the fit itself, and so
-    understates the true uncertainty; the ``*_error`` columns are where that
-    uncertainty can be read off in the meantime.
+    ``mag_cal_error`` combines two things: the star's own measurement error,
+    and the uncertainty of the transform it was calibrated through, correlations
+    between the terms included. The second part is the larger of the two on a
+    fit to a modest number of stars, and it is worked out per star rather than
+    added on as a constant, because a fit predicts best at the centroid of the
```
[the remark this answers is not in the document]
mwcraig: Confirmed against both of the PR's own fixtures — the sentence overstates. In the 50-star example the transform component is ~0.0062 against a ~0.0205 measurement component, and even in the 12-star Monte Carlo the test's own comment puts the transform share at about a fifth of the total. The measurement term dominates in every case this PR measured. Will reword to describe the transform term as a significant contribution that grows as the number of fitted stars shrinks — and the part the old formula omitted entirely — folding it into the shrink of this section already requested at line ~695.

CHANGES.rst
```diff
@@ -163,6 +163,19 @@ Bug Fixes
...
+  It used to be the star's own measurement error scaled by ``1 + a``, as though
+  the transform the star was calibrated through were known exactly, and so was
+  too small -- on a fit to a few dozen stars with 0.02 mag scatter, by around
+  20%. It now carries the uncertainty of the fit as well, correlations between
```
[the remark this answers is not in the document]
mwcraig: Confirmed — the pairing mixes two fixtures. The ~20% understatement comes from the 12-star Monte Carlo (`n_stars = 12` in `test_transform_to_catalog_error_matches_a_monte_carlo`), while the 50-star example — which is what "a few dozen" reads as — grows the median error by only ~4.5%. The concise rewrite already proposed in the thread at line ~166 drops the quantitative claim entirely, which resolves this; if a number is kept, it should be stated as "up to ~20% on a 12-star fit, a few percent at 50".

stellarphot/utils/magnitude_transforms.py
mwcraig: I can live with the single try/except -- make sure to document it as an unlikely-to-occur thing that we are checking just to ensure we don't produce nonsense.

stellarphot/utils/magnitude_transforms.py
mwcraig: File the follow-up issue

stellarphot/utils/magnitude_transforms.py
mwcraig: I OKed the try/except

CHANGES.rst
mwcraig: Applied in 0d77a2b — the entry is now the six-line version proposed above.

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 0d77a2b — the one-line summary stays and both paragraphs now live in a numpydoc ``Notes`` section, with the ``errorbars`` paragraph cut down per the thread below.

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 0d77a2b — both corner-case justifications are gone. The ``Notes`` now say only that this code consumes the covariance matrix itself, so the matrix is what it checks; ``errorbars`` is a summary flag lmfit derives from the same object.

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 0d77a2b — the summary is cut to the two-part statement, and the mechanism (coefficients as ufloats carrying the covariance, ``mag_inst`` itself a ufloat so the sensitivity is never written down, the deliberate slight over-count for in-fit stars) is now a short ``Notes`` section.

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 0d77a2b — the section is about half its former length, with the two factual corrections from the other threads on it folded in (the catalog color does enter when a color term is fit; "significant" rather than the larger of the two).

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Done in 0d77a2b — `_fit_a_noisy_catalog` is deleted; `_fit_a_catalog` grew `sigma=0.0, seed=None` (and a `mag_error` passthrough) with one merged docstring, and the noisy callers pass `n_stars=50, sigma=0.02` and their seed explicitly.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Done in 0d77a2b — `_break_the_covariance` and both tests built on it are deleted. The `covar is None` arm (and the six-word isfinite check) stay in `_coefficient_uncertainties` so lmfit's documented `None` reports cleanly instead of crashing; that branch is now untested by design.

pyproject.toml
mwcraig: Pinned in 0d77a2b: `lmfit >=1.2.2` with a comment noting `create_uvars()` arrived there, and `pixi.lock` refreshed — the lock diff is just the recorded pin, no package resolution changed.

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 0d77a2b — `_coefficient_uncertainties` now attempts `uncertainties.correlated_values` itself and treats `LinAlgError`/`ValueError` as an unusable covariance, with a comment documenting it as not known to occur with real data, checked only so a covariance that fails to decompose is never turned into nonsense. Per the decision in the `_break_the_covariance` thread there is no mock test for it.

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 0d77a2b — the context manager is gone from the plain-float branch (a short comment notes NaN inputs never tripped the invalid flag there), and it remains around the object-array propagation path.

stellarphot/utils/magnitude_transforms.py
mwcraig: Follow-up filed as #691, and the docstring correction landed in 0d77a2b — it now states the roughly ``c * sigma_color`` term is knowingly omitted and points at the issue.

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 0d77a2b — the warning is hoisted out of the error-column branch, worded around the coefficient `*_error` columns being NaN, with the `mag_cal_error` sentence appended only when `obs_error_column` was given.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Overtaken by the decision in the thread at line ~928: both tests and `_break_the_covariance` are deleted in 0d77a2b, which resolves the duplication too.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Done in 0d77a2b together with Matt's thread on this line — one merged helper, one docstring.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Done in 0d77a2b — the three tests now share a module-scoped `_noisy_fit_result` fixture (via `module_mocker`) that runs the fit once; verified none of them mutate the table.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Done in 0d77a2b — `_fit_a_catalog` grew a `mag_error` passthrough and both this test and the perfect-fit test pass their local constant explicitly.

stellarphot/utils/magnitude_transforms.py
mwcraig: The try/except landed in 0d77a2b, so the "single place" claim is now true as written — `create_uvars` can no longer re-answer the question downstream. The function is otherwise kept as-is.

CHANGES.rst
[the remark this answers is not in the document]
mwcraig: Resolved in 0d77a2b — the rewritten changelog entry drops the quantitative claim entirely.

stellarphot/utils/magnitude_transforms.py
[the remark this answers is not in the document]
mwcraig: Reworded in 0d77a2b — the docstring now describes the transform term as a significant contribution that grows as the number of fitted stars shrinks, rather than the larger of the two.

PR #688, opened 2026-08-11
Calibrate against the observed band, weighted by both errors

mwcraig:
Closes #680.

Stacked on #682, #683, #684 and #687 — please merge those first; this branch
contains their commits.

Three things #680 asked for.

### (a) A mismatched passband pair is always a mistake, so raise

Calibrating an SR observation against a Cousins R catalog magnitude is not a
judgement call, it is a bug — and nothing stopped it. `transform_to_catalog` now
raises `ValueError` naming both bands, with `R`/`RC`/`Rc` and `I`/`IC`/`Ic`
treated as the same band, so the check is not over-eager about spelling.

The check sits **immediately after** the existing "No rows with passband" check.
That ordering is deliberate: it keeps `..._passband_not_in_table_raises` getting
the error it expects, and validating the table before commenting on the
arguments is the better order anyway.

`cat_filter` and `cat_color` now default to `None` rather than to `"R"` and
`("R", "I")`. Leaving the literal would have turned every call that omits
`cat_filter` for a non-R band into a hard error — a gratuitous break.
`cat_filter` resolves to `obs_filter`, and `cat_color` to that band's
conventional pair (U→U−B, B/V→B−V, R/I→R−I, SG→SG−SR, SR/SI→SR−SI, …). A band
with no conventional pair raises and asks for `cat_color`, rather than dying on
a `KeyError` for a missing column.

One existing test was doing the mistaken thing: `..._accepts_photometry_data`
calibrated `obs_filter="SR"` against `cat_filter="R"`. Sloan r′ and Cousins R
genuinely are different bands, so that is fixed rather than suppressed.

### (b) Weighting, not a magnitude cut

#676 removed `faintest_mag_for_transform` and left nothing in its place. The
replacement is to weight the fit by `1/√(obs_error² + cat_error²)` — a faint
star with a large catalog error contributes less, without a cliff edge at an
arbitrary magnitude.

The catch, and the reason this warns: `CatalogData.passband_columns` emits
`mag_error_<band>` only for a catalog's *native* passbands.
`transform_apass_bands` / `transform_refcat2_bands` add Johnson–Cousins `mag_R`
and `mag_I` afterwards and return no errors. So the combined weighting works for
APASS's B, V, SG, SR, SI and **not** for R and I — the default and the shipped
notebook's main use. That path falls back to the observed errors alone and warns
once per call naming the band, citing #685, which is the follow-up for teaching
the band transformers to propagate errors.

A missing or non-positive catalog error is excluded from the fit, the same way a
bad observed error already was, rather than becoming an infinite weight.

### (c) End-to-end against real catalogs

`remote_data` tests against APASS DR9 and refcat2 — the first time this function
has been exercised against a catalog it did not build itself. They run only in
the single `tox -e coverage -- --remote-data=any` job.

The point of them is the masked-row check: a real catalog has rows with no
magnitude, and those must come back NaN in `mag_cal` and `mag_cat` — **and not
as any of the usual fill values** (`-999`, `1e20`, `0.0`), because `np.isnan` on
a filled column is exactly the check that would silently pass.

Field choice mattered more than it looks. `transform_to_catalog` hardcodes
`radius=1 * u.degree` and `from_vizier` uses `row_limit=-1` (filed as #686), so
the test pulls a full degree-radius cone however small a field it asks for.
Hence the north galactic pole, where that cone is survivable, with a comment
saying so.

### Notes for review

- **These two tests are unrun.** There is no network in the environment they
  were written in. The body was dry-run against a synthetic masked catalog, so
  the array logic, masked-row assertions and `pytest.warns` all execute — what
  is unverified is the real query, the real band transform, and the field's
  actual star counts. Worth eyeballing the first live run.
- The masked-row block is guarded by `if no_catalog_mag.any()`, so it cannot
  fail if the field has no masked rows — but that also means the headline check
  silently does nothing in that case.
- `ValueError` is in `SERVER_DOWN_ERRORS`, so the tests deliberately narrow the
  exception set around the `transform_to_catalog` call; a real failure cannot be
  xfailed away.
- The synthetic test catalog now carries `mag_error_R`/`mag_error_I` by default.
  That was the honest fix for ~60 tests rather than blanket `pytest.warns` — a
  real catalog does supply errors for bands it measured — and no numeric
  expectation moved, because `scale_covar=True` makes a uniform catalog error
  invisible in every reported uncertainty.
- **Behaviour change for existing users:** passing `cat_filter` for a different
  band than `obs_filter` is now a hard error. The shipped notebook passes the
  same band for both, so it is unaffected.

---

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -169,24 +169,12 @@ Bug Fixes
   somewhat larger. When a fit leaves no usable covariance it is NaN, with
   a warning. Catalog magnitude uncertainty is still excluded -- see the
   docstring for why. [#674]
-+ ``transform_to_catalog()`` now raises when ``cat_filter`` names a different
-  band from ``obs_filter``. Calibrating observations in one band against
-  catalog magnitudes in another makes the color difference between the two
-  bands part of the fit, which then reports it as a transform coefficient and
-  applies it to every star, so a mismatched pair is always a mistake -- two
-  spellings of one band, such as ``"R"`` and ``"RC"``, are not one. Along with
-  that, ``cat_filter`` and ``cat_color`` now default to `None` and are
-  resolved to ``obs_filter`` and the color conventionally used with it,
-  instead of to the literals ``"R"`` and ``("R", "I")``, which would have made
-  every call that omitted them for another band an error. The fit is also now
-  weighted by the observed errors and the catalog's own errors combined in
-  quadrature wherever the catalog has a ``mag_error_<cat_filter>`` column, and
-  a star whose catalog error is missing or not positive is left out of the
-  fit, as one with an unusable observed error already was. Where the column is
-  absent -- which is the case for the Johnson-Cousins R and I, transformed
-  from a catalog's native bands by a transform that does not yet propagate
-  errors -- the fit is weighted by the observed errors alone, exactly as
-  before, and logs one message naming the band. See issue ``#685``. [#680]
++ ``transform_to_catalog()`` now warns loudly when ``cat_filter`` explicitly
+  names a band different from ``obs_filter`` instead of silently folding the
+  color difference into the fit. Passband spellings are canonicalized
+  (``"Rc"`` and ``"R"`` are one band), ``cat_filter``/``cat_color`` default
+  from ``obs_filter``, and the fit is weighted by observed and catalog errors
+  in quadrature where available, with the weighting recorded in ``.meta``. [#680]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -169,6 +169,24 @@ Bug Fixes
   somewhat larger. When a fit leaves no usable covariance it is NaN, with
   a warning. Catalog magnitude uncertainty is still excluded -- see the
   docstring for why. [#674]
++ ``transform_to_catalog()`` now raises when ``cat_filter`` names a different
+  band from ``obs_filter``. Calibrating observations in one band against
+  catalog magnitudes in another makes the color difference between the two
+  bands part of the fit, which then reports it as a transform coefficient and
+  applies it to every star, so a mismatched pair is always a mistake -- two
+  spellings of one band, such as ``"R"`` and ``"RC"``, are not one. Along with
+  that, ``cat_filter`` and ``cat_color`` now default to `None` and are
+  resolved to ``obs_filter`` and the color conventionally used with it,
+  instead of to the literals ``"R"`` and ``("R", "I")``, which would have made
+  every call that omitted them for another band an error. The fit is also now
+  weighted by the observed errors and the catalog's own errors combined in
+  quadrature wherever the catalog has a ``mag_error_<cat_filter>`` column, and
+  a star whose catalog error is missing or not positive is left out of the
+  fit, as one with an unusable observed error already was. Where the column is
+  absent -- which is the case for the Johnson-Cousins R and I, transformed
+  from a catalog's native bands by a transform that does not yet propagate
+  errors -- the fit is weighted by the observed errors alone, exactly as
+  before, and logs one message naming the band. See issue ``#685``. [#680]
 
 2.1.2 (2026-07-19)
 -------------------
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -58,6 +58,36 @@ _FIT_REDCHI_COLUMN = "fit_redchi"
 _FIT_MATCH_ARCSEC = 1.0
 _CAL_MATCH_ARCSEC = 1.5
 
+# Passband names that mean the same physical band. Cousins R and I are each
+# written several ways, and `transform_refcat2_bands` really does add ``mag_R``
+# and ``mag_RC`` as copies of one column, so matching R observations against a
+# catalog's RC is one band spelled two ways rather than the mistake the check
+# that uses this exists to catch.
+_SAME_BAND = (frozenset({"R", "RC", "Rc"}), frozenset({"I", "IC", "Ic"}))
+
+# The color conventionally used with each passband, for a caller who names a
+# band but no color. Which color goes with a band is a convention rather than
+# something derivable, so a band that is not here has to be asked about rather
+# than guessed at. Every spelling of Cousins R and I is a key of its own,
+# because the band is looked up as the caller wrote it and is never rewritten
+# into one preferred spelling.
+_CONVENTIONAL_COLOR = {
+    "U": ("U", "B"),
+    "B": ("B", "V"),
+    "V": ("B", "V"),
+    "R": ("R", "I"),
+    "RC": ("R", "I"),
+    "Rc": ("R", "I"),
+    "I": ("R", "I"),
+    "IC": ("R", "I"),
+    "Ic": ("R", "I"),
+    "SU": ("SU", "SG"),
+    "SG": ("SG", "SR"),
+    "SR": ("SR", "SI"),
+    "SI": ("SR", "SI"),
+    "SZ": ("SI", "SZ"),
+}
+
 # How strongly the terms of the fit may be correlated with each other before
 # their individual values stop meaning anything. See `_underdetermined_reason`,
 # which explains how this was arrived at.
@@ -393,6 +423,28 @@ def _calibrated_with_uncertainty(fit_result, covar, mag_inst, color, errors, fit
         return unumpy.nominal_values(calibrated), unumpy.std_devs(calibrated)
 
 
+def _is_same_band(one_filter, another_filter):
+    """
+    Whether two passband names name the same physical band.
+
+    Parameters
+    ----------
+
+    one_filter, another_filter : str
+        Passband names to compare.
+
+    Returns
+    -------
+    bool
+        `True` if the two names are the same, or are two spellings of one
+        band; see `_SAME_BAND`.
+    """
+    if one_filter == another_filter:
+        return True
+
+    return any(one_filter in band and another_filter in band for band in _SAME_BAND)
+
+
 def _check_known_terms(terms, argument_name):
     """
     Raise if any of ``terms`` is not a coefficient of the transform model.
@@ -593,8 +645,8 @@ def transform_to_catalog(
     obs_mag_col="mag_inst",
     obs_error_column=None,
     cat_name="apass_dr9",
-    cat_filter="R",
-    cat_color=("R", "I"),
+    cat_filter=None,
+    cat_color=None,
     vary=_DEFAULT_VARY,
     expected=None,
     in_place=True,
@@ -629,9 +681,10 @@ def transform_to_catalog(
 
     obs_error_column : str, optional
         Name of the column in ``observed_mags_grouped`` that contains the error in
-        the magnitude. The fit is weighted by these errors, so leaving this
-        out is rarely what you want and warns. Stars whose error is not a
-        positive, finite number are left out of the fit.
+        the magnitude. The fit is weighted by these errors, combined with the
+        catalog's own errors where it has them, so leaving this out is rarely
+        what you want and warns. Stars whose error is not a positive, finite
+        number are left out of the fit.
 
     cat_name : str, optional
         Name of the catalog to calibrate against, either ``"apass_dr9"`` or
@@ -641,12 +694,22 @@ def transform_to_catalog(
         Name of the passband in the catalog that should be matched to the
         instrumental magnitudes, e.g. ``"R"`` or ``"SR"``. This is a passband
         name, not a column name: the column used is ``mag_<cat_filter>``.
+        Defaults to ``obs_filter``, and naming a different band raises, since
+        calibrating observations in one band against catalog magnitudes in
+        another folds the color difference between the two bands into the fit,
+        where it comes back as a transform coefficient and is applied to every
+        star. Two spellings of one band -- ``"R"`` and ``"RC"``, say -- are not
+        a different band.
 
     cat_color : tuple of two strings, optional
         Names of the two passbands whose difference is the color. The color is
         calculated in the order the passbands are given, so ``("R", "I")``
         means the ``mag_R`` column minus the ``mag_I`` column. As with
         ``cat_filter``, these are passband names rather than column names.
+        Defaults to the color conventionally used with ``cat_filter``, which
+        is ``("R", "I")`` for R and I, ``("B", "V")`` for B and V, and the
+        neighboring Sloan band for the Sloan ones. A band with no convention
+        recorded raises rather than being guessed at.
 
     vary : sequence of str, optional
         Which terms of the transform model to fit. Any term not named here is
@@ -718,6 +781,26 @@ def transform_to_catalog(
     freedom in **mag squared** -- the same name and a completely different
     scale, so values from a weighted and an unweighted fit must not be
     compared with each other.
+
+    The fit is weighted by the observed errors and the catalog's own errors
+    combined in quadrature, ``1 / sqrt(obs_error**2 + cat_error**2)``, wherever
+    the catalog has a ``mag_error_<cat_filter>`` column; by the observed errors
+    alone, with a log message naming the band, wherever it does not. A star
+    whose catalog error is missing or is not positive is left out of the fit,
+    exactly as one whose observed error is. Which bands have an error depends on the
+    catalog rather than on this function: `~stellarphot.CatalogData` builds
+    ``mag_error_<band>`` for the bands a catalog measured itself -- B, V, SG,
+    SR and SI for APASS DR9, the Sloan bands for refcat2 -- while the
+    Johnson-Cousins R and I are added afterwards by
+    `~stellarphot.utils.magnitude_system_transforms.transform_apass_bands` and
+    `~stellarphot.utils.magnitude_system_transforms.transform_refcat2_bands`,
+    which return magnitudes and no errors. So the fallback is the normal case
+    for R and I, the bands most often calibrated here, until issue #685
+    teaches those transforms to propagate errors. When the fallback is in use,
+    ``fit_redchi`` is the column to watch: an image whose stars scatter about
+    the transform by more than they claim to be uncertain reports a
+    ``fit_redchi`` well above one, and where the catalog's uncertainty is the
+    reason, weighting cannot know that but the scatter still shows up there.
 
     The reported uncertainties are scaled by ``fit_redchi``, because `lmfit`'s
     ``scale_covar`` is left on. They therefore describe the scatter actually
@@ -808,6 +891,32 @@ def transform_to_catalog(
             f"{sorted(set(observed_mags_grouped['passband']))}."
         )
 
+    # After the check above, not before it: a caller who got the observed
+    # passband name wrong should hear about the name rather than about the
+    # mismatch that follows from it, and checking the table before commenting
+    # on the arguments is the better order in any case.
+    cat_filter = obs_filter if cat_filter is None else cat_filter
+    if not _is_same_band(obs_filter, cat_filter):
+        raise ValueError(
+            f"Observed passband {obs_filter!r} and catalog passband "
+            f"{cat_filter!r} are different bands. Calibrating one against the "
+            "other makes the color difference between the two bands part of "
+            "the fit, which then reports it as a transform coefficient and "
+            "applies it to every star in the image. Leave cat_filter out to "
+            "use the observed passband, or transform the observations to the "
+            "catalog's band first."
+        )
+
+    if cat_color is None:
+        try:
+            cat_color = _CONVENTIONAL_COLOR[cat_filter]
+        except KeyError:
+            raise ValueError(
+                "No color is conventionally used with passband "
+                f"{cat_filter!r}, so cat_color must be given explicitly. The "
+                f"passbands with a default are {sorted(_CONVENTIONAL_COLOR)}."
+            ) from None
+
     # Output is built at the length of the whole table, not of the rows in
     # this passband, and is written by row index below. Rows that are never
     # written keep the NaN they start with.
@@ -854,6 +963,34 @@ def transform_to_catalog(
     # this table, and the invariant documented just above is what makes
     # writing by row index legal.
     result = observed_mags_grouped if in_place else observed_mags_grouped.copy()
+
+    # What is fit is the difference between an observed magnitude and a catalog
+    # one, so both of them contribute to how well each star constrains the fit,
+    # and the catalog says which of its stars it knows poorly. Weighting by the
+    # two errors combined is what replaced the magnitude cut this function used
+    # to make: a faint star is down-weighted by its errors rather than dropped
+    # at a limit drawn across the whole field.
+    #
+    # The column is decided once here rather than per image, since it depends
+    # only on the catalog and the band, and so is the log message when there
+    # is none -- it is a fact about the call rather than about any one image.
+    cat_error_column = f"mag_error_{cat_filter}"
+    if obs_error_column is None:
+        # The fit is unweighted at the caller's request, and was warned about
+        # above. Weighting it by the catalog errors alone would make it a
+        # weighted fit nobody asked for, and would quietly change what
+        # ``fit_redchi`` means.
+        cat_error_column = None
+    elif cat_error_column not in cat.colnames:
+        logger.warning(
+            f"The {cat_name} catalog has no {cat_error_column} column, so the "
+            "fit is weighted by the errors of the observations alone and how "
+            "well the catalog knows each star is not accounted for. Expect "
+            "this for the Johnson-Cousins R and I bands, which are transformed "
+            "from the catalog's native bands by a transform that does not yet "
+            "propagate errors; see issue #685."
+        )
+        cat_error_column = None
 
     for group_number, (file, one_image_all_bands) in enumerate(
         zip(
@@ -908,6 +1045,15 @@ def transform_to_catalog(
             # A non-positive or non-finite error gives a meaningless weight.
             good_dat = good_dat & np.isfinite(errors) & (errors > 0)
 
+        if cat_error_column is not None:
+            cat_error = _to_float_array(cat[cat_error_column][cat_idx])
+            # The same rule as for the observed errors just above, and for
+            # the same reasons: a catalog error of zero is an infinite
+            # weight, and a missing one makes the weight NaN, which
+            # poisons every residual in the image rather than only this
+            # star's.
+            good_cat = good_cat & np.isfinite(cat_error) & (cat_error > 0)
+
         # Both halves have to be good for the *same* star. Checking each on
         # its own is not enough: the two sets can be disjoint, which leaves
         # nothing to take the median of below.
@@ -932,7 +1078,16 @@ def transform_to_catalog(
         fit_color = model_color[good]
         offset = fit_mag if fit_diff else 0.0
         fit_data = cat_mag[good] - offset
-        weights = 1.0 / errors[good] if obs_error_column is not None else 1.0
+
+        if obs_error_column is None:
+            weights = 1.0
+        elif cat_error_column is None:
+            weights = 1.0 / errors[good]
+        else:
+            # In quadrature, because the residual being fit is the
+            # difference between a measured magnitude and a catalog one and
+            # is uncertain by both of them.
+            weights = 1.0 / np.hypot(errors[good], cat_error[good])
 
         # Fewer stars than terms has to be caught before the fit rather than
         # after it: with nothing left to fit, lmfit takes the square root of
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -11,6 +11,7 @@ from numpy.linalg import LinAlgError
 from uncertainties import correlated_values, unumpy
 
 from ..catalogs import apass_dr9, refcat2
+from ..settings.aavso_models import AAVSOFilters
 from .magnitude_system_transforms import transform_apass_bands, transform_refcat2_bands
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


mwcraig (commented):
Critical review of the rebased branch (head 6be4621). Ten findings, most-severe first, as inline comments: the three production concerns are the no-override same-band error, the silent exclusion of stars with zero/masked catalog errors, and `_CONVENTIONAL_COLOR` advertising bands that end in bare KeyErrors; the most important test finding is the vacuous negligible-catalog-error comparison (`in_place=True`).

*This review was written by Claude at Matt's direction.*

stellarphot/utils/magnitude_transforms.py
```diff
@@ -808,6 +891,32 @@ def transform_to_catalog(
             f"{sorted(set(observed_mags_grouped['passband']))}."
         )
 
+    # After the check above, not before it: a caller who got the observed
+    # passband name wrong should hear about the name rather than about the
+    # mismatch that follows from it, and checking the table before commenting
+    # on the arguments is the better order in any case.
+    cat_filter = obs_filter if cat_filter is None else cat_filter
+    if not _is_same_band(obs_filter, cat_filter):
```
mwcraig: **Correctness:** This ValueError has no escape hatch, which hard-blocks previously working cross-band calibrations — e.g. AAVSO CV/TG workflows (clear/DSLR data reduced to V) or instrumental spellings like `rp` against `SR`.

Concretely: a user with unfiltered data who previously called `transform_to_catalog(obs, 'CV', cat_filter='V')` now hits this error (`_SAME_BAND` covers only the R/I spellings). Following the error's advice to omit `cat_filter` defaults it to `'CV'`, which then fails on the missing conventional color or dies on `cat['mag_CV']` — so CV data cannot be calibrated at all. Consider an explicit override parameter for intentional cross-band transformation.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -908,6 +1045,15 @@ def transform_to_catalog(
...
             good_dat = good_dat & np.isfinite(errors) & (errors > 0)
 
+        if cat_error_column is not None:
+            cat_error = _to_float_array(cat[cat_error_column][cat_idx])
+            # The same rule as for the observed errors just above, and for
+            # the same reasons: a catalog error of zero is an infinite
+            # weight, and a missing one makes the weight NaN, which
+            # poisons every residual in the image rather than only this
+            # star's.
+            good_cat = good_cat & np.isfinite(cat_error) & (cat_error > 0)
```
mwcraig: **Correctness:** Stars whose catalog error is masked, NaN, or zero are silently excluded from the fit with no per-image count or message.

APASS DR9 reports zero/blank `e_mag` for single-observation stars, so on a sparse field calibrating B or V, stars that participated in the fit pre-PR now fail `np.isfinite(cat_error) & (cat_error > 0)`; an image can drop below the stars-vs-terms threshold and return NaN coefficients with only the generic 'underdetermined' warning — nothing tells the user that catalog-error filtering removed the stars. (The diagnostic at ~line 985 covers the column being entirely absent, which is a different case.)

stellarphot/utils/magnitude_transforms.py
```diff
@@ -58,6 +58,36 @@
...
+# that uses this exists to catch.
+_SAME_BAND = (frozenset({"R", "RC", "Rc"}), frozenset({"I", "IC", "Ic"}))
+
+# The color conventionally used with each passband, for a caller who names a
+# band but no color. Which color goes with a band is a convention rather than
+# something derivable, so a band that is not here has to be asked about rather
+# than guessed at. Every spelling of Cousins R and I is a key of its own,
+# because the band is looked up as the caller wrote it and is never rewritten
+# into one preferred spelling.
+_CONVENTIONAL_COLOR = {
```
mwcraig, line 70: **Correctness:** `_CONVENTIONAL_COLOR` (and the ValueError below listing 'passbands with a default') advertises bands the hardcoded catalog fetches can never supply — `U`, `SU`, and the `Rc`/`Ic` spellings — which then die in a bare KeyError.

E.g. `transform_to_catalog(observed, 'U')` or `(observed, 'Rc')`: the same-band check passes, `cat_color` resolves, the Vizier query runs, then `cat['mag_U']` / `cat['mag_Rc']` raises KeyError — the band transformers emit `mag_RC`/`mag_R` but never `mag_Rc`, and neither catalog branch requests U or SU. The user is told these bands 'have a default' by the function's own error message, then gets an unexplained KeyError after a network round trip.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -755,38 +827,265 @@ def test_transform_to_catalog_omitted_quadratic_term_degrades_gracefully(mocker)
...
+    sigma = 0.02
+
+    catalog, ra, dec, instrumental = _generate_fake_catalog(
+        n_stars, a=0.02, c=0.15, cat_error=1e-30
+    )
+    observed = _generate_observed_table(
+        ra, dec, instrumental, noise_sigma=sigma, seed=13579
+    )
+
+    negligible = _run_transform_to_catalog(mocker, catalog, observed)
```
mwcraig: **Test bug:** This test is vacuous: `transform_to_catalog` defaults to `in_place=True`, so both runs return the same `Table` object and every column is compared with itself.

`negligible = _run_transform_to_catalog(mocker, catalog, observed)` *is* `observed`; after `del catalog['mag_error_R']` the second call overwrites and returns the same object as `absent`. `assert_array_equal` then passes bit-for-bit no matter what the two weighting paths compute — change the fallback to `weights = 2.0/errors[good]` and this still passes, despite the comment claiming the equality 'is only true if the two paths do the same arithmetic'. Fix: pass `in_place=False` or run the second call on a fresh copy of the observations.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2176,3 +2563,172 @@ def test_transform_to_catalog_non_numeric_existing_column_raises(mocker):
...
+    fetch, passbands, transformer = _REMOTE_CATALOGS[cat_name]
+
+    try:
+        nearby = fetch(
+            _REMOTE_FIELD_CENTER,
+            radius=_REMOTE_OBSERVED_RADIUS,
+            clip_by_frame=False,
+            padding=0,
+        )
+    except SERVER_DOWN_ERRORS as e:
```
mwcraig: **Test robustness:** This direct catalog fetch is guarded by `except SERVER_DOWN_ERRORS`, which includes `ValueError` — the exact trap `_QUERY_DOWN_ERRORS` (defined above) exists to avoid for the call under test.

`apass_dr9`/`refcat2` and `CatalogData.from_vizier` raise `ValueError` for argument-validation and catalog-construction problems, so any such regression makes the fetch xfail as 'Vizier is down or misbehaving' instead of failing — permanently hidden in the only CI job that runs remote data. The fetch guard should use `_QUERY_DOWN_ERRORS` too.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -755,38 +827,265 @@ def test_transform_to_catalog_omitted_quadratic_term_degrades_gracefully(mocker)
...
+
+    with caplog.at_level(logging.WARNING, logger=_MAGNITUDE_TRANSFORMS_LOGGER):
+        result = _run_transform_to_catalog(mocker, catalog, observed)
+
+    about_the_band = [
+        r for r in caplog.records if "mag_error_R" in r.message and "#685" in r.message
+    ]
+    # One message per call, not one per image: the column is missing from the
+    # catalog, which is a fact about the call rather than about an image.
+    assert len(about_the_band) == 1
```
mwcraig, line 1024: **Test coverage:** The 'one message per call, not one per image' assertion is made against a single-image table, so it cannot distinguish per-call from per-image logging. An implementation that moved the warning inside the per-image loop would emit exactly one record here and `assert len(about_the_band) == 1` would still pass. Pinning the claimed property needs a two-image table (the file's multi-image helpers exist for this).

CHANGES.rst
```diff
@@ -169,6 +169,24 @@ Bug Fixes
   somewhat larger. When a fit leaves no usable covariance it is NaN, with
   a warning. Catalog magnitude uncertainty is still excluded -- see the
   docstring for why. [#674]
++ ``transform_to_catalog()`` now raises when ``cat_filter`` names a different
```
mwcraig: **Convention:** This entry is ~19 lines; per the feedback on #687, CHANGES.rst entries should be roughly 3–6 lines: what changed, the user-visible effect, and the issue number. The color-difference rationale, quadrature formula, and fallback/logging details belong in the PR description or the docstring.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -58,6 +58,36 @@
 _FIT_MATCH_ARCSEC = 1.0
 _CAL_MATCH_ARCSEC = 1.5
 
+# Passband names that mean the same physical band. Cousins R and I are each
+# written several ways, and `transform_refcat2_bands` really does add ``mag_R``
+# and ``mag_RC`` as copies of one column, so matching R observations against a
+# catalog's RC is one band spelled two ways rather than the mistake the check
+# that uses this exists to catch.
+_SAME_BAND = (frozenset({"R", "RC", "Rc"}), frozenset({"I", "IC", "Ic"}))
```
mwcraig: **Duplication:** `_SAME_BAND` (and the Rc/RC/IC/Ic rows of `_CONVENTIONAL_COLOR`) is a third private copy of band-alias knowledge that already exists in `settings/aavso_models.py` (`AAVSOFilters.Rc = 'R'`) and in `magnitude_system_transforms.py` (`mag_R` aliased to `mag_RC`, with a comment saying the duplication should end). When that acknowledged cleanup lands (or a new alias appears), the copies will drift and the same-band check will start rejecting or accepting spellings inconsistently with the rest of the package. Consider deriving this from, or at least cross-referencing, `AAVSOFilters`.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2176,3 +2563,172 @@ def test_transform_to_catalog_non_numeric_existing_column_raises(mocker):
...
+            radius=_REMOTE_OBSERVED_RADIUS,
+            clip_by_frame=False,
+            padding=0,
+        )
+    except SERVER_DOWN_ERRORS as e:
+        pytest.xfail(f"Vizier is down or misbehaving: {e}")
+
+    stars = nearby.passband_columns(passbands=passbands, transformer=transformer)
+
+    catalog_mag = np.ma.filled(np.ma.asarray(stars["mag_R"], dtype=float), np.nan)
```
mwcraig: **Reuse:** This inlines `np.ma.filled(np.ma.asarray(col, dtype=float), np.nan)` (here and again ~13 lines below), which is exactly the body of the existing `_to_float_array` helper in `magnitude_transforms.py`. Importing and reusing the helper keeps the test measuring the same conversion the production code performs (including the integer-MaskedColumn edge case its docstring documents).

stellarphot/utils/magnitude_transforms.py
```diff
@@ -855,6 +964,34 @@ def transform_to_catalog(
...
+    # one, so both of them contribute to how well each star constrains the fit,
+    # and the catalog says which of its stars it knows poorly. Weighting by the
+    # two errors combined is what replaced the magnitude cut this function used
+    # to make: a faint star is down-weighted by its errors rather than dropped
+    # at a limit drawn across the whole field.
+    #
+    # The column is decided once here rather than per image, since it depends
+    # only on the catalog and the band, and so is the log message when there
+    # is none -- it is a fact about the call rather than about any one image.
+    cat_error_column = f"mag_error_{cat_filter}"
```
mwcraig, line 1042: **Observability:** Whether the fit was weighted by combined errors, observed errors alone, or unweighted is decided per call but recorded nowhere in the output — even though the docstring itself warns that values from differently weighted fits must not be compared. A user calibrating SR (quadrature weights) and R (observed-only fallback) from the same night gets `fit_redchi` and `*_error` columns on different scales with nothing in the table distinguishing them; the only trace is a transient log message. A weighting-mode entry in the table metadata (or a column) would make the outputs self-describing.

stellarphot/utils/magnitude_transforms.py
mwcraig: Isn't there something in AAVSOPassbands that already captures this? Also, by the AAVSO Passbands convention, passbands are always uppercase.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -58,6 +58,36 @@
...
 _CAL_MATCH_ARCSEC = 1.5
 
+# Passband names that mean the same physical band. Cousins R and I are each
+# written several ways, and `transform_refcat2_bands` really does add ``mag_R``
+# and ``mag_RC`` as copies of one column, so matching R observations against a
+# catalog's RC is one band spelled two ways rather than the mistake the check
+# that uses this exists to catch.
+_SAME_BAND = (frozenset({"R", "RC", "Rc"}), frozenset({"I", "IC", "Ic"}))
+
+# The color conventionally used with each passband, for a caller who names a
```
mwcraig: I can see some utility to this -- except for the multiple entries for R/RC/Rc.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -393,6 +423,28 @@ def _calibrated_with_uncertainty(fit_result, covar, mag_inst, color, errors, fit
         return unumpy.nominal_values(calibrated), unumpy.std_devs(calibrated)
 
 
+def _is_same_band(one_filter, another_filter):
```
mwcraig: This feels a lot like overkill...

stellarphot/utils/magnitude_transforms.py
```diff
@@ -641,12 +694,22 @@ def transform_to_catalog(
         Name of the passband in the catalog that should be matched to the
         instrumental magnitudes, e.g. ``"R"`` or ``"SR"``. This is a passband
         name, not a column name: the column used is ``mag_<cat_filter>``.
+        Defaults to ``obs_filter``, and naming a different band raises, since
```
mwcraig: If naming a different band raises, then why have this argument at all?

stellarphot/utils/magnitude_transforms.py
```diff
@@ -808,6 +891,32 @@ def transform_to_catalog(
...
+            f"Observed passband {obs_filter!r} and catalog passband "
+            f"{cat_filter!r} are different bands. Calibrating one against the "
+            "other makes the color difference between the two bands part of "
+            "the fit, which then reports it as a transform coefficient and "
+            "applies it to every star in the image. Leave cat_filter out to "
+            "use the observed passband, or transform the observations to the "
+            "catalog's band first."
+        )
+
+    if cat_color is None:
```
mwcraig: Leaving out cat_color is not an error if the reequested fit contains no color terms.

CHANGES.rst
mwcraig: This would be a good novel, but it is a bad changelog entry

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -257,9 +287,51 @@ def _generate_fake_catalog(
...
+        errors = np.broadcast_to(
+            np.asarray(cat_error, dtype=float), np.shape(cat_r)
+        ).copy()
+        catalog["mag_error_R"] = errors
+        catalog["mag_error_I"] = errors.copy()
+
     return catalog, ra, dec, instrumental
 
 
+def _catalog_bands_renamed(catalog, **new_from_old):
```
mwcraig: Why is this test duplicating functionality (passband maps) that exists elsewhere?

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -755,38 +827,265 @@ def test_transform_to_catalog_omitted_quadratic_term_degrades_gracefully(mocker)
     )
 
 
+def _one_star_with_a_large_error(mocker, n_stars=20, cat_error=_FAKE_CATALOG_ERROR):
```
mwcraig, line 836: Why is this and the function that follows a separate function? Is it used multiple times?

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -755,38 +827,265 @@ def test_transform_to_catalog_omitted_quadratic_term_degrades_gracefully(mocker)
...
+    # catalog, which is a fact about the call rather than about an image.
+    assert len(about_the_band) == 1
+
+    np.testing.assert_allclose(result["mag_cal"], catalog["mag_R"], rtol=0, atol=1e-6)
+
+
+def test_transform_to_catalog_negligible_catalog_error_changes_nothing(mocker):
+    # The other side of the fallback: adding catalog errors to the weighting
+    # must not have moved the answer for anyone whose catalog errors are too
+    # small to matter. A catalog error 27 orders of magnitude below the
```
mwcraig: 27 orders of magnitude? That is a ridiculous test, may as well drop it.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed valid. `_SAME_BAND` covers only the R/I spellings, so `transform_to_catalog(obs, "CV", cat_filter="V")` raises here, and following the error's advice to omit `cat_filter` dead-ends anyway: `"CV"` has no entry in `_CONVENTIONAL_COLOR` and no catalog fetch produces a `mag_CV` column. Under the pre-PR defaults that workflow worked, so #680's premise that a mismatched pair is *always* a mistake has a real counterexample in unfiltered/DSLR data reduced to V.

Two coherent ways out, and the choice interacts with @mwcraig's question below about whether `cat_filter` should exist at all:

1. **Explicit mismatch warns instead of raising.** The default can never mismatch (it *is* `obs_filter`), so the raise would only ever fire on a deliberately typed `cat_filter` — treat that as intent, log a loud warning about the color-term contamination, and proceed. This keeps CV/TG calibration possible.
2. **Keep the raise and drop `cat_filter` entirely** (deprecation cycle), which forecloses cross-band calibration on purpose.

Happy to implement either on the branch once the direction is picked.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed, and there is a cleaner fix than adding a log message. The code comment's justification ("a catalog error of zero is an infinite weight") does not actually hold on this path: the weight is `1.0 / np.hypot(errors[good], cat_error[good])`, and `errors > 0` is already enforced by `good_dat`, so `hypot(obs_err, 0.0)` is just `obs_err` — finite.

So instead of excluding the star, treat a masked/NaN/non-positive catalog error as `0.0` in the quadrature: "the catalog does not know its own uncertainty for this star, weight it by the observed error alone." That keeps APASS's zero-`e_mag` single-observation stars in the fit exactly as they were pre-PR, removes this silent-exclusion class entirely, and needs no new diagnostics. Only the NaN case needs the substitution to keep the weight from poisoning the image.

Happy to implement on the branch.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed. The apass branch fetches passbands `B, V, R, I, SR, SG, SI` and refcat2 fetches `B, V, R, I`; the transformers emit `mag_RC`/`mag_IC` (uppercase) as aliases but never `mag_Rc`/`mag_Ic`, and nothing produces `mag_U` or `mag_SU`. So `cat_filter="U"` or `"Rc"` passes every argument check, spends the Vizier round trip, and dies in a bare `KeyError` — after the function's own error message listed those bands as having a default.

Fix folds into the canonicalization discussed on the `_SAME_BAND` thread: uppercase the incoming name, map spellings through `AAVSOFilters`, and trim `_CONVENTIONAL_COLOR` to (or validate `cat_filter` against) the bands the selected catalog can actually supply, before any network call.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Confirmed. `transform_to_catalog` defaults to `in_place=True` and `_run_transform_to_catalog` does not override it, so `negligible`, `absent`, and `observed` are all the same `Table` object and every `assert_array_equal` compares a column with itself. Will fix by running each call on its own copy of the observations (see also the thread at the 1e-30 comment above about whether this test survives at all).

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Confirmed — `SERVER_DOWN_ERRORS` in `conftest.py` includes `ValueError` (for the GAIA aperture service), and `apass_dr9`/`refcat2`/`from_vizier` raise `ValueError` for argument-validation and catalog-construction failures, so a real regression in the fetch would xfail as "Vizier is down" in the one CI job that runs remote data. The guard around the direct fetch will use `_QUERY_DOWN_ERRORS`, same as the call under test. Copilot flagged the same thing.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Confirmed — `_generate_observed_table` builds a single image, so one warning record is consistent with both per-call and per-image emission. Will rebuild the table from two images via `_combine_observed_tables` so `assert len(about_the_band) == 1` actually pins the per-call property.

CHANGES.rst
mwcraig: Agreed — the entry runs ~19 lines against the 3–6 line convention from #687. Will trim to roughly: mismatch raises, `cat_filter`/`cat_color` now default from `obs_filter`, fit weighted by observed and catalog errors in quadrature where available, with the issue numbers. The rationale and fallback details already live in the docstring and PR description.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed — `AAVSOFilters.Rc = "R"` already exists in `settings/aavso_models.py`, and `magnitude_system_transforms.py` carries the `mag_R`/`mag_RC` aliasing with a comment saying the duplication should end. Canonicalizing passband names through `AAVSOFilters` (uppercased, per the AAVSO convention @mwcraig noted on the other thread) collapses `_SAME_BAND` into plain equality and removes the triplicated `Rc`/`RC` rows from `_CONVENTIONAL_COLOR`. That is the direction the fixes on this file's other threads converge on.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Confirmed — the inlined expression is character-for-character the body of `_to_float_array`, including the integer-`MaskedColumn` handling its docstring documents. Will import and reuse the helper in both places.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — the docstring warns against comparing differently weighted fits while the output carries no trace of which weighting ran. Proposal: record it in the result's metadata keyed by passband (e.g. `result.meta["transform_weighting"]["SR"] = "combined"` / `"observed"` / `"unweighted"`), since the function is called once per passband on the same table and a single flat key would be overwritten by the next call. Happy to implement.

stellarphot/utils/magnitude_transforms.py
mwcraig: Yes — `AAVSOFilters` in `settings/aavso_models.py` already maps `Rc → "R"`, and `magnitude_system_transforms.py` has the `mag_R`/`mag_RC` aliasing with a comment saying the duplication should end. The right fix is to canonicalize incoming passband names through `AAVSOFilters`, uppercasing per the AAVSO convention you note, so `"Rc"`/`"Ic"` stop existing as distinct spellings anywhere in this module. That also fixes the bug in the review thread below where `cat_filter="Rc"` passes every check and then dies in a `KeyError` after the Vizier query, because no catalog ever grows a `mag_Rc` column.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — with names canonicalized through `AAVSOFilters` first (per your comment at `_SAME_BAND` above), this table shrinks to one entry per physical band and the triplicated `R`/`RC`/`Rc` rows disappear.

stellarphot/utils/magnitude_transforms.py
mwcraig: It is — and it goes away entirely once spellings are canonicalized through `AAVSOFilters` at the top of `transform_to_catalog`: the same-band check becomes plain `==` and both this function and `_SAME_BAND` get deleted.

stellarphot/utils/magnitude_transforms.py
mwcraig: Fair question. After this PR the argument has only two remaining uses: choosing a spelling of the same band (which canonicalization makes moot) and deliberate cross-band calibration (which the new check forbids). So the coherent options are:

1. **Drop `cat_filter`** (with a deprecation cycle — it has been in the signature for years) and always calibrate against `obs_filter`. Simplest, but it forecloses the unfiltered/DSLR workflows raised in the review thread above (CV or TG data calibrated against catalog V), which the old defaults supported.
2. **Keep it as the explicit escape hatch**: defaulted, it is always `obs_filter`; explicitly passed with a different band, it warns loudly about the color-term contamination instead of raising.

Option 2 answers both threads at once, but it softens what #680 asked for, so it is your call. Happy to implement either.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — `fitting_color` is already computed a few lines up, so the default-resolution (and this raise) should be gated on it. When neither `"c"` nor `"d"` is in `vary` and the band has no conventional color, the fit needs no color at all; `color_cat` can simply come back NaN instead of the call refusing to run. Will fix so the raise only fires when a color term is actually being fit and no `cat_color` was given or derivable.

CHANGES.rst
mwcraig: Agreed — will cut it to 3–6 lines (mismatch raises, new `None` defaults resolved from `obs_filter`, quadrature weighting, issue numbers) per the #687 convention. The rest already lives in the docstring and PR description.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: It isn't quite a passband map — no band relationships or transform coefficients are encoded here. It copies the synthetic catalog's `mag_R`/`mag_I` columns verbatim under other names (`SR`/`SI`, `TG`, `RC`, …) so the one fixture can pose as a catalog in any band; a fit against the renamed band must recover exactly what it recovers against the original, which is what the tests using it assert. The real passband machinery (`passband_columns` + `transform_apass_bands`/`transform_refcat2_bands`) is exactly what the unit tests mock out, and the remote-data tests at the bottom of the file exercise the real thing. It has five call sites. If the name suggests band knowledge it doesn't have, renaming it (e.g. `_catalog_posing_as_bands`) might help — say the word.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Yes — both are shared: `test_transform_to_catalog_weights_by_inverse_error` and `test_transform_to_catalog_weights_by_inverse_error_without_catalog_errors` (the next two tests in the file) each call `_one_star_with_a_large_error` and `_assert_the_bad_star_was_ignored`, differing only in `cat_error=None`. The factoring is what keeps the fallback path asserting the identical outcome rather than a re-derived one.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: The size is deliberate rather than decorative: `1e-30` is so far below the observed errors that `np.hypot(err, 1e-30) == err` to the last bit in float64, which is what makes exact bit-for-bit equality between the two weighting code paths a legitimate expectation — a *plausible* small catalog error would force tolerance-based comparison, which cannot distinguish the quadrature path from the fallback at all.

That said, the review thread below found the comparison is currently vacuous anyway (`in_place=True` means it compares a table with itself), so as written it earns nothing. Two options: fix it by running each call on a fresh copy, or drop it and let the quadrature test above carry the weighting coverage. Your call — happy to do either.

stellarphot/utils/magnitude_transforms.py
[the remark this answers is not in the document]
mwcraig: Confirmed — same conclusion as the review threads on `_SAME_BAND`/`_CONVENTIONAL_COLOR`: the transformers emit `mag_R`/`mag_RC` and `mag_I`/`mag_IC` but never `mag_Rc`/`mag_Ic`, so the lowercase spellings pass the same-band check and then `KeyError` after the Vizier round trip. The fix is to canonicalize passband names through `AAVSOFilters` (uppercased) before any column lookup, which removes the lowercase spellings from this module entirely.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2176,3 +2563,172 @@ def test_transform_to_catalog_non_numeric_existing_column_raises(mocker):
...
+
+    try:
+        nearby = fetch(
+            _REMOTE_FIELD_CENTER,
+            radius=_REMOTE_OBSERVED_RADIUS,
+            clip_by_frame=False,
+            padding=0,
+        )
+    except SERVER_DOWN_ERRORS as e:
+        pytest.xfail(f"Vizier is down or misbehaving: {e}")
```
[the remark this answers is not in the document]
mwcraig: Confirmed — the same issue was flagged in the review thread on this guard. The direct fetch will use `_QUERY_DOWN_ERRORS` (the tuple defined just above that strips `ValueError`) so an argument-validation or catalog-construction regression fails instead of xfailing as a server outage.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2176,3 +2563,172 @@ def test_transform_to_catalog_non_numeric_existing_column_raises(mocker):
...
+    if no_catalog_mag.any():
+        for column in ("mag_cal", "mag_cat"):
+            # np.ma.getdata rather than the column itself: `np.isnan` of a
+            # masked array is masked in turn, and `.all()` of that is True
+            # whatever the numbers underneath are -- exactly the check that
+            # passes when it should not.
+            missing = np.ma.getdata(result[column])[no_catalog_mag]
+            assert np.isnan(missing).all(), column
+            for fill_value in (-999, 1e20, 0.0):
+                assert not (missing == fill_value).any(), (column, fill_value)
```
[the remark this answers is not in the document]
mwcraig: Agreed — the PR description flags this same gap ("the headline check silently does nothing in that case"). Will add `assert no_catalog_mag.any()` after the field selection, with a message noting that failure means either the band transform stopped producing masked entries (the regression this test exists for) or the field needs re-choosing, and then run the fill-value assertions unconditionally. Both causes are worth failing loudly on rather than passing silently.

stellarphot/utils/magnitude_transforms.py
mwcraig: 1

stellarphot/utils/magnitude_transforms.py
mwcraig: 2 is the way to go

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: I like that new name better

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Decided above to run on fresh copies. Claridy the explanatory text in the comment

PR #692, opened 2026-08-13
Propagate catalog errors through the band transforms

mwcraig:
Closes #685

`transform_apass_bands()` and `transform_refcat2_bands()` produced `mag_R`/`mag_I` (plus `mag_B`/`mag_V` for refcat2) with no matching `mag_error_` columns, so `transform_to_catalog()`'s combined-error weighting fell back to the observed errors alone for exactly the bands most calibrations use. Both transformers now propagate the catalog's native band errors, and each transform's published rms residual is included in quadrature as the noise floor of the transform itself.

How the errors are computed:

- **RefCat2 (Pan-STARRS1 → JC)**: each JC band is `native + P(g − r)`, so the partials are `P′(c)` through the color plus one for the native band. `PanStarrs1ToJohnsonCousins.__call__` grew an optional `from_magnitude_errors` argument and returns `(mags, errors)` when it is given; the residual comes from the `residual` field already stored in `PS1_to_JC.json`.
- **APASS (Jester)**: the transforms are linear in g, r, i, so the errors go through the squared coefficients. The coefficient dict moved from inside `filter_transform()` to module level so both users share one source, alongside the tabulated rms values. The shipped `R` and `I` relations are compositions of the published ones (`R = V − (V−R)`, `I = R − (Rc−Ic)`), so their residuals combine in quadrature (≈0.032 and ≈0.033 mag).
- **APASS with `apply_sdssdr7_transform=True`**: the matrix step makes the intermediate SDSS g, r, i errors *correlated*, so the two stages cannot be propagated separately — the Jacobians are composed instead (Jester row through the matrix). A finite-difference test pins this down; a naive two-stage propagation fails it.

A table without the native error columns gets magnitudes only, no invented errors — that keeps `transform_to_catalog()`'s observed-errors-only fallback (and its warning) for catalogs that genuinely have none.

Also fixed in passing: `transform_refcat2_bands()` broke on a one-row table because `np.squeeze` flattened the transform output before the column indexing; `np.atleast_2d` restores the expected shape.

Tests: propagated errors are checked against hand-written published coefficients/polynomials and against finite-difference partials of the magnitude transforms, zero input error must yield exactly the residual, the refcat2 `SZ` error must not leak into any band, and the remote-data catalog tests now assert the error columns appear.

https://claude.ai/code/session_01UhAvKVUNXqqKYwzwr3La9d

### DIFF
--- a/stellarphot/utils/magnitude_system_transforms.py
+++ b/stellarphot/utils/magnitude_system_transforms.py
@@ -188,8 +188,11 @@ class PanStarrs1ToJohnsonCousinsMixin:
     """
 
     def __call__(
-        self, from_magnitudes: np.typing.ArrayLike, ps1_band_for_V: str = "gp1"
-    ) -> np.ndarray:
+        self,
+        from_magnitudes: np.typing.ArrayLike,
+        ps1_band_for_V: str = "gp1",
+        from_magnitude_errors: np.typing.ArrayLike | None = None,
+    ) -> np.ndarray | tuple[np.ndarray, np.ndarray]:
         """
         Transform Pan-STARRS1 magnitudes to Johnson-Cousins magnitudes.
 
@@ -200,6 +203,15 @@ class PanStarrs1ToJohnsonCousinsMixin:
             or (n, 6), where n is the number of magnitudes in the Pan-STARRS1 system
             you wish to transform to Johnson-Cousins system.
 
+        ps1_band_for_V : str, optional
+            Pan-STARRS1 band to use as the base magnitude for the V band.
+
+        from_magnitude_errors : np.ArrayLike, optional
+            Errors in the Pan-STARRS1 magnitudes, same shape as
+            ``from_magnitudes``. When given, the return value is the tuple
+            ``(magnitudes, errors)``, where the errors combine the input
+            errors propagated through each transform with the published
+            residual of that transform, in quadrature.
         """
         # Convert to numpy array
         from_magnitudes = np.asarray(from_magnitudes)
@@ -223,9 +235,14 @@ class PanStarrs1ToJohnsonCousinsMixin:
         )
         if len(from_magnitudes.shape) == 1:
             from_magnitudes = from_magnitudes[np.newaxis, :]
+        if from_magnitude_errors is not None:
+            from_magnitude_errors = np.asarray(from_magnitude_errors).reshape(
+                from_magnitudes.shape
+            )
         to_mags = np.zeros(
             from_magnitudes.shape[:-1] + (len(self.to_system.passbands),)
         )
+        to_errors = np.zeros_like(to_mags)
         for index, jc_band in enumerate(self.to_system.passbands):
             to_band_name = jc_band
             match to_band_name:
@@ -246,12 +263,29 @@ class PanStarrs1ToJohnsonCousinsMixin:
                         f"transformation from Pan-STARRS1 to Johnson-Cousins."
                     )
 
+            transform = self.transform_information[(from_band_name, to_band_name)]
             to_mags[..., index] = (
-                self.transform_information[(from_band_name, to_band_name)].polynomial(
-                    color
-                )
+                transform.polynomial(color)
                 + from_magnitudes[..., from_band_index[from_band_name]]
             )
+
+            if from_magnitude_errors is not None:
+                # The transform is base magnitude plus polynomial in the
+                # g - r color, so the partial with respect to each input
+                # band is the polynomial's derivative through the color
+                # (+1 for gp1, -1 for rp1) plus one for the base band.
+                dpoly_dcolor = transform.polynomial.deriv()(color)
+                partials = np.zeros_like(from_magnitudes)
+                partials[..., from_band_index["gp1"]] += dpoly_dcolor
+                partials[..., from_band_index["rp1"]] -= dpoly_dcolor
+                partials[..., from_band_index[from_band_name]] += 1.0
+                to_errors[..., index] = np.sqrt(
+                    np.sum((partials * from_magnitude_errors) ** 2, axis=-1)
+                    + transform.residual**2
+                )
+
+        if from_magnitude_errors is not None:
+            return np.squeeze(to_mags), np.squeeze(to_errors)
         return np.squeeze(to_mags)
 
 
@@ -335,7 +369,7 @@ def transform_apass_bands(table, apply_sdssdr7_transform: bool = False):
         Table of catalog information. This is assumed to be in "passband" format
     """
     # Putting this here to avoid a circular import
-    from .magnitude_transforms import filter_transform
+    from .magnitude_transforms import _JESTER_RMS, _JESTER_TRANSFORMS, filter_transform
 
     # Set up for input to Jester transform
     use_columns = dict(
@@ -343,6 +377,24 @@ def transform_apass_bands(table, apply_sdssdr7_transform: bool = False):
         r="mag_SR",
         i="mag_SI",
     )
+
+    # Errors are propagated only when the catalog supplies them for every
+    # native band the transform uses.
+    error_columns = dict(
+        g="mag_error_SG",
+        r="mag_error_SR",
+        i="mag_error_SI",
+    )
+    have_errors = all(column in table.colnames for column in error_columns.values())
+    if have_errors:
+        error_values = {band: table[column] for band, column in error_columns.items()}
+        # Effective linear coefficients of the native g, r and i in each
+        # output band; without the SDSS DR7 step these are just the Jester
+        # coefficients.
+        effective_coeffs = {
+            band: np.asarray(_JESTER_TRANSFORMS[band][:3], dtype=float)
+            for band in ("R", "I")
+        }
 
     # The Jester transformation really for transformation from the ugriz on the
     # 2.5m SDSS telescope to the Johnson-Cousins system.
@@ -379,6 +431,20 @@ def transform_apass_bands(table, apply_sdssdr7_transform: bool = False):
             i="mag_SI_tmp",
         )
 
+        if have_errors:
+            # The matrix step makes the SDSS g, r and i magnitudes
+            # correlated -- each is a mix of the same native bands -- so the
+            # errors cannot be propagated through it and Jester separately.
+            # Compose the two linear maps instead: the effective coefficient
+            # of each native band is the Jester row through the matrix.
+            matrix = transform.transform_information.array
+            for band in ("R", "I"):
+                jester_row = np.zeros(matrix.shape[0])
+                # Rows 1, 2 and 3 of the matrix output are g, r and i, the
+                # same indexing used for sdss_mags above.
+                jester_row[1:4] = _JESTER_TRANSFORMS[band][:3]
+                effective_coeffs[band] = (jester_row @ matrix)[1:4]
+
     table["mag_RC"] = filter_transform(
         table,
         output_filter="R",
@@ -392,9 +458,25 @@ def transform_apass_bands(table, apply_sdssdr7_transform: bool = False):
         transform="jester",
     )
 
+    if have_errors:
+        # The transforms are linear in the native g, r and i, so the errors
+        # propagate through the squared coefficients, and the rms residual
+        # of the Jester transform adds in quadrature.
+        for band, native in (("R", "RC"), ("I", "IC")):
+            coeff_g, coeff_r, coeff_i = effective_coeffs[band]
+            table[f"mag_error_{native}"] = np.sqrt(
+                (coeff_g * error_values["g"]) ** 2
+                + (coeff_r * error_values["r"]) ** 2
+                + (coeff_i * error_values["i"]) ** 2
+                + _JESTER_RMS[band] ** 2
+            )
+
     # Yes, this is dumb. You fix it if you want it less dumb.
     table["mag_R"] = table["mag_RC"]
     table["mag_I"] = table["mag_IC"]
+    if have_errors:
+        table["mag_error_R"] = table["mag_error_RC"]
+        table["mag_error_I"] = table["mag_error_IC"]
     # The dumbness has ended for now
 
     # Remove the temporary columns if they were created
@@ -430,20 +512,53 @@ def transform_refcat2_bands(table):
         ]
     )
 
-    transformed_data = transform(
-        transform_input.T,
-        ps1_band_for_V="gp1",
-    )
+    error_columns = ["mag_error_SG", "mag_error_SR", "mag_error_SI", "mag_error_SZ"]
+    have_errors = all(column in table.colnames for column in error_columns)
+    if have_errors:
+        zeros = np.zeros(num_rows)
+        error_input = np.array(
+            [
+                table["mag_error_SG"],
+                table["mag_error_SR"],
+                table["mag_error_SI"],
+                table["mag_error_SZ"],
+                zeros,  # Placeholder for the y_p1 band
+                zeros,  # Placeholder for the w_p1 band
+            ]
+        )
+        transformed_data, transformed_errors = transform(
+            transform_input.T,
+            ps1_band_for_V="gp1",
+            from_magnitude_errors=error_input.T,
+        )
+        transformed_errors = np.atleast_2d(transformed_errors)
+    else:
+        transformed_data = transform(
+            transform_input.T,
+            ps1_band_for_V="gp1",
+        )
+
+    # A one-row table comes back squeezed to a 1-d array
+    transformed_data = np.atleast_2d(transformed_data)
 
     table["mag_B"] = transformed_data[:, 0]
     table["mag_V"] = transformed_data[:, 1]
     table["mag_RC"] = transformed_data[:, 2]
     table["mag_IC"] = transformed_data[:, 3]
 
+    if have_errors:
+        table["mag_error_B"] = transformed_errors[:, 0]
+        table["mag_error_V"] = transformed_errors[:, 1]
+        table["mag_error_RC"] = transformed_errors[:, 2]
+        table["mag_error_IC"] = transformed_errors[:, 3]
+
     # Yes, this is dumb. You fix it if you want it less dumb.
     # We should only have one name for RC and IC, not two...
     table["mag_R"] = table["mag_RC"]
     table["mag_I"] = table["mag_IC"]
+    if have_errors:
+        table["mag_error_R"] = table["mag_error_RC"]
+        table["mag_error_I"] = table["mag_error_IC"]
     # The dumbness has ended for now
 
     return table
--- a/stellarphot/utils/magnitude_system_transforms.py
+++ b/stellarphot/utils/magnitude_system_transforms.py
@@ -198,15 +198,16 @@ class PanStarrs1ToJohnsonCousinsMixin:
 
         Parameters
         ----------
-        from_magnitudes : np.ArrayLike
+        from_magnitudes : np.typing.ArrayLike
             Pan-STARRS1 magnitudes to transform. The shape should either be (6,)
             or (n, 6), where n is the number of magnitudes in the Pan-STARRS1 system
-            you wish to transform to Johnson-Cousins system.
+            you wish to transform to Johnson-Cousins system. The result has the
+            shape (4,) or (n, 4) to match.
 
         ps1_band_for_V : str, optional
             Pan-STARRS1 band to use as the base magnitude for the V band.
 
-        from_magnitude_errors : np.ArrayLike, optional
+        from_magnitude_errors : np.typing.ArrayLike, optional
             Errors in the Pan-STARRS1 magnitudes, same shape as
             ``from_magnitudes``. When given, the return value is the tuple
             ``(magnitudes, errors)``, where the errors combine the input
@@ -233,12 +234,19 @@ class PanStarrs1ToJohnsonCousinsMixin:
             from_magnitudes[..., from_band_index["gp1"]]
             - from_magnitudes[..., from_band_index["rp1"]]
         )
-        if len(from_magnitudes.shape) == 1:
+        input_was_1d = len(from_magnitudes.shape) == 1
+        if input_was_1d:
             from_magnitudes = from_magnitudes[np.newaxis, :]
         if from_magnitude_errors is not None:
-            from_magnitude_errors = np.asarray(from_magnitude_errors).reshape(
-                from_magnitudes.shape
-            )
+            from_magnitude_errors = np.asarray(from_magnitude_errors)
+            if len(from_magnitude_errors.shape) == 1:
+                from_magnitude_errors = from_magnitude_errors[np.newaxis, :]
+            if from_magnitude_errors.shape != from_magnitudes.shape:
+                raise ValueError(
+                    "from_magnitude_errors must have the same shape as "
+                    f"from_magnitudes; got {from_magnitude_errors.shape} and "
+                    f"{from_magnitudes.shape}."
+                )
         to_mags = np.zeros(
             from_magnitudes.shape[:-1] + (len(self.to_system.passbands),)
         )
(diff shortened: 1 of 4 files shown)


stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -430,20 +512,53 @@ def transform_refcat2_bands(table):
         ]
     )
 
-    transformed_data = transform(
-        transform_input.T,
-        ps1_band_for_V="gp1",
-    )
+    error_columns = ["mag_error_SG", "mag_error_SR", "mag_error_SI", "mag_error_SZ"]
+    have_errors = all(column in table.colnames for column in error_columns)
```
mwcraig, line 557: **Bug: a catalog missing only `mag_error_SZ` silently gets no propagated errors at all.** The z partials are exactly zero for every output band — this PR's own test asserts the SZ error must not leak into any band — yet `have_errors` requires it. A table with `mag_error_SG/SR/SI` but no `mag_error_SZ` falls into the no-errors branch: no `mag_error_B/V/R/I` columns, no warning, and `transform_to_catalog` silently degrades to observed-only weighting for exactly the bands this PR set out to fix. Requiring only the three bands that actually contribute (and feeding zeros for SZ) would fix this.

mwcraig (commented):

Critical review of the error-propagation changes. The core math held up under scrutiny: the SDSS DR7 Jacobian composition was verified against the shipped 6×6 matrix, the Jester coefficients and the composed R/I residuals match the published relations, and 0-row/1-row handling works. What follows are the issues that survived verification — two of the bugs were reproduced by actually executing the code. Seven are anchored as inline comments; three findings fall outside the diff so they're listed here:

**`stellarphot/utils/magnitude_transforms.py:1070` — stale warning text (pre-existing lines, made wrong by this PR).** The missing-error warning still says the R/I transforms "do not yet propagate errors; see issue #685", which this PR makes false. After merge, that branch firing for R or I means error columns were genuinely lost upstream — a real bug to investigate — but the warning tells the user it's expected and normal.

**`stellarphot/utils/magnitude_transforms.py:830` — stale docstring.** `transform_to_catalog`'s docstring still states that `transform_apass_bands`/`transform_refcat2_bands` return magnitudes and no errors, and that the observed-only fallback "is the normal case for R and I … until issue #685". That now permanently describes the pre-PR behavior.

**`stellarphot/utils/magnitude_transforms.py:1063` — all-NaN errors mislabeled as 'combined' (plausible, not reproduced).** `weighting_mode` is recorded as `combined` whenever the `mag_error` column exists, and this PR now creates such columns even when every propagated value is NaN (e.g. a Vizier response lacking `e_*mag` pivots to all-NaN error columns). The quadrature propagates NaN through, `meta['transform_weighting']` says `combined`, and the warning is silenced — while the per-star NaN fallback makes the actual fit observed-error-only. Pre-PR this case honestly recorded `observed`.

Minor nits not worth inline comments: `to_errors` is allocated even on the no-errors path; the "Yes, this is dumb" RC→R alias block is now duplicated in both transformers; and the test file's `_JESTER_R_RMS`/`_JESTER_I_RMS` copy the production quadrature expression instead of pinning independent literals, so a bug in the composition would pass its own test.

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -223,9 +235,14 @@ def __call__(
         )
         if len(from_magnitudes.shape) == 1:
             from_magnitudes = from_magnitudes[np.newaxis, :]
+        if from_magnitude_errors is not None:
+            from_magnitude_errors = np.asarray(from_magnitude_errors).reshape(
```
mwcraig: **Bug (reproduced): `reshape` silently scrambles a wrong-shaped errors array instead of raising.** `np.asarray(from_magnitude_errors).reshape(from_magnitudes.shape)` accepts any same-*size* array. Passing errors transposed as `(6, 3)` against mags of `(3, 6)` succeeds and reinterprets the values in C order — in a test run, correct errors `[0.796, 0.238, 0.074, 0.081]` silently became `[0.849, 0.354, 0.573, 0.512]` with no exception, despite the docstring promising the same shape as `from_magnitudes`. This should validate the shape and raise (allowing only the 1-d→2-d promotion that mirrors the magnitudes handling above).

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -246,12 +263,29 @@ def __call__(
...
                 + from_magnitudes[..., from_band_index[from_band_name]]
             )
+
+            if from_magnitude_errors is not None:
+                # The transform is base magnitude plus polynomial in the
+                # g - r color, so the partial with respect to each input
+                # band is the polynomial's derivative through the color
+                # (+1 for gp1, -1 for rp1) plus one for the base band.
+                dpoly_dcolor = transform.polynomial.deriv()(color)
+                partials = np.zeros_like(from_magnitudes)
```
mwcraig: **Bug (reproduced): integer-dtype magnitudes crash the error path but not the magnitudes-only path.** `np.zeros_like(from_magnitudes)` inherits the input dtype, so `PanStarrs1ToJohnsonCousins.load()(np.array([20, 19, 19, 18, 18, 17]), from_magnitude_errors=np.full(6, 0.01))` raises `UFuncTypeError` ("Cannot cast ufunc 'add' output from float64 to int64") at the `+= dpoly_dcolor` line, while the identical call without errors succeeds. `np.zeros(from_magnitudes.shape)` fixes it.

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -246,12 +263,29 @@ def __call__(
...
+                partials[..., from_band_index["gp1"]] += dpoly_dcolor
+                partials[..., from_band_index["rp1"]] -= dpoly_dcolor
+                partials[..., from_band_index[from_band_name]] += 1.0
+                to_errors[..., index] = np.sqrt(
+                    np.sum((partials * from_magnitude_errors) ** 2, axis=-1)
+                    + transform.residual**2
+                )
+
+        if from_magnitude_errors is not None:
+            return np.squeeze(to_mags), np.squeeze(to_errors)
```
mwcraig: **Simplification: fix the shape contract here instead of re-inflating at call sites.** The unconditional `np.squeeze` is what forced the two `np.atleast_2d` patches in `transform_refcat2_bands`, and the new errors return path doubles the number of sites needing that dance — the one-row `IndexError` this PR fixes will recur at the next caller. The function already knows whether the input was 1-d (the `np.newaxis` promotion above); squeezing only in that case restores "output shape matches input shape" and lets the call sites drop `np.atleast_2d` entirely.

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -379,6 +431,20 @@ def transform_apass_bands(table, apply_sdssdr7_transform: bool = False):
             i="mag_SI_tmp",
         )
 
+        if have_errors:
+            # The matrix step makes the SDSS g, r and i magnitudes
+            # correlated -- each is a mix of the same native bands -- so the
+            # errors cannot be propagated through it and Jester separately.
+            # Compose the two linear maps instead: the effective coefficient
+            # of each native band is the Jester row through the matrix.
+            matrix = transform.transform_information.array
```
mwcraig, line 473: **Docs mismatch: on the `apply_sdssdr7_transform=True` path, only the Jester rms enters the quadrature.** The USNO′→SDSS DR7 matrix step contributes no residual term (the shipped JSON has no `residual` field), while CHANGES.rst says "each transform's published rms residual is included in quadrature". At minimum this comment and the CHANGES entry should state that the USNO step's residual is unavailable and omitted, so the two-step errors are understood as a floor.

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -392,9 +458,25 @@ def transform_apass_bands(table, apply_sdssdr7_transform: bool = False):
         transform="jester",
     )
 
+    if have_errors:
```
mwcraig, line 498: **Robustness: nothing couples this error math to the transform choice above.** This block hardcodes the Jester linear coefficients while the transform is selected by a free string passed to `filter_transform` twice (lines 452 and 458). `filter_transform` equally supports `"ivezic"`, which is a polynomial in color rather than linear in g/r/i — a one-word edit to the string would change the magnitudes while `mag_error_RC/IC` silently stayed Jester-based, and `transform_to_catalog` would weight fits with wrong errors. A shared `transform_name` variable plus a guard comment (or assertion) that the error propagation is Jester/linear-only would close the gap.

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -430,20 +512,53 @@ def transform_refcat2_bands(table):
...
+            [
+                table["mag_error_SG"],
+                table["mag_error_SR"],
+                table["mag_error_SI"],
+                table["mag_error_SZ"],
+                zeros,  # Placeholder for the y_p1 band
+                zeros,  # Placeholder for the w_p1 band
+            ]
+        )
+        transformed_data, transformed_errors = transform(
```
mwcraig: **Simplification: the `transform(...)` call is duplicated across the `have_errors` branches**, differing only in the errors kwarg, with `np.atleast_2d` applied asymmetrically (errors inside the branch, data after it). A future edit applied to one branch only — a different `ps1_band_for_V`, a new kwarg — would make the two paths transform magnitudes differently. One call passing `from_magnitude_errors=error_input.T if have_errors else None`, then unpacking, keeps them in lockstep.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: From magnitudes will never be integers, though, fyi...

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -379,6 +431,20 @@ def transform_apass_bands(table, apply_sdssdr7_transform: bool = False):
...
         )
 
+        if have_errors:
+            # The matrix step makes the SDSS g, r and i magnitudes
+            # correlated -- each is a mix of the same native bands -- so the
+            # errors cannot be propagated through it and Jester separately.
+            # Compose the two linear maps instead: the effective coefficient
+            # of each native band is the Jester row through the matrix.
+            matrix = transform.transform_information.array
+            for band in ("R", "I"):
```
mwcraig: Can this not be expressed as a matrix multiplication?

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -430,20 +512,53 @@ def transform_refcat2_bands(table):
...
+        error_input = np.array(
+            [
+                table["mag_error_SG"],
+                table["mag_error_SR"],
+                table["mag_error_SI"],
+                table["mag_error_SZ"],
+                zeros,  # Placeholder for the y_p1 band
+                zeros,  # Placeholder for the w_p1 band
+            ]
+        )
```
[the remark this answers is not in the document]
mwcraig: Confirmed — the z partial is exactly zero, so requiring `mag_error_SZ` gates propagation on a column that can never affect the answer. Checked directly:

```python
m  = np.array([[20., 19., 19., 18., 18., 17.]])
e0 = np.array([[0.02, 0.02, 0.02, 0.0, 0., 0.]])
e1 = np.array([[0.02, 0.02, 0.02, 5.0, 0., 0.]])  # absurd z error
np.allclose(t(m, from_magnitude_errors=e0)[1],
            t(m, from_magnitude_errors=e1)[1])   # -> True
```

The reason is structural, not numerical: in `__call__` the only partials ever written are `gp1`, `rp1`, and the base band (`gp1`/`rp1`/`ip1`), so index 3 stays zero for every output band.

One scoping note on the practical exposure: for catalogs built by `stellarphot`'s own `refcat2` reader this is latent rather than active — Vizier's `J/ApJ/867/105/refcat2` supplies `e_zmag` alongside `e_gmag`/`e_rmag`/`e_imag`, so all four error columns arrive together. It bites hand-built or column-trimmed tables, which `transform_refcat2_bands()` accepts just as readily. Worth fixing regardless, since a silent downgrade to observed-only weighting is exactly the failure this PR set out to remove.

Same point as @mwcraig's comment on line 516; noting it here so the threads don't diverge.

stellarphot/utils/magnitude_system_transforms.py
```diff
@@ -200,6 +203,15 @@ def __call__(
...
             you wish to transform to Johnson-Cousins system.
 
+        ps1_band_for_V : str, optional
+            Pan-STARRS1 band to use as the base magnitude for the V band.
+
+        from_magnitude_errors : np.ArrayLike, optional
+            Errors in the Pan-STARRS1 magnitudes, same shape as
+            ``from_magnitudes``. When given, the return value is the tuple
+            ``(magnitudes, errors)``, where the errors combine the input
+            errors propagated through each transform with the published
```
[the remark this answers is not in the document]
mwcraig: Correct on the facts — `np.ArrayLike` does not exist; the annotation is `np.typing.ArrayLike`.

Worth noting the scope, though: this is a pre-existing convention in the file rather than something the new parameter introduced. All three docstring occurrences are the same:

- line 201 `from_magnitudes : np.ArrayLike` (pre-existing)
- line 209 `from_magnitude_errors : np.ArrayLike, optional` (new)
- line 303 `from_magnitudes : np.ArrayLike` in `MatrixTransformMixin.__call__` (pre-existing, untouched by this PR)

So fixing only the new line would leave the file inconsistent in the other direction. Happy to change all three to `np.typing.ArrayLike` in one pass.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Confirmed. The z error genuinely cannot reach any output — an absurd z error changes nothing:

```python
m  = np.array([[20., 19., 19., 18., 18., 17.]])
e0 = np.array([[0.02, 0.02, 0.02, 0.0, 0., 0.]])
e1 = np.array([[0.02, 0.02, 0.02, 5.0, 0., 0.]])
np.allclose(t(m, from_magnitude_errors=e0)[1],
            t(m, from_magnitude_errors=e1)[1])   # -> True
```

Structurally: `__call__` only ever writes partials for `gp1`, `rp1`, and the base band, so index 3 is zero for B, V, Rc and Ic alike.

Suggested change at line 515:

```python
# The z partials are identically zero for every output band, so the z
# error is not required -- feed zeros for it when the column is absent.
error_columns = ["mag_error_SG", "mag_error_SR", "mag_error_SI"]
have_errors = all(column in table.colnames for column in error_columns)
if have_errors:
    zeros = np.zeros(num_rows)
    z_error = table["mag_error_SZ"] if "mag_error_SZ" in table.colnames else zeros
```

with `z_error` used in place of `table["mag_error_SZ"]` in `error_input`.

One calibration on how often this fires today: catalogs from `stellarphot`'s own refcat2 reader always have all four, since Vizier's `J/ApJ/867/105/refcat2` ships `e_zmag` with `e_gmag`/`e_rmag`/`e_imag`. So it is latent for that path and live for hand-built or trimmed tables, which the function accepts equally.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Reproduced. `reshape` accepts any same-size array and silently reinterprets it in C order:

```python
mags = np.array([[20., 19., 19., 18., 18., 17.],
                 [21., 20., 20., 19., 19., 18.],
                 [22., 21., 21., 20., 20., 19.]])
errs = np.random.default_rng(42).uniform(0.01, 0.1, size=(3, 6))

t(mags, from_magnitude_errors=errs)[1][0]        # [0.1368 0.0479 0.0611 0.0904]
t(mags, from_magnitude_errors=errs.T.copy())[1][0]  # [0.142  0.0572 0.0937 0.0726]
```

No exception either way, and the docstring right above promises "same shape as `from_magnitudes`".

There is also a precedent for the fix inside this same file: `MatrixTransformMixin.__call__` validates its input shape and raises `"expected 5 passband rows plus a final row of ones (6 rows total)"` (with a test at `test_transform_wrong_shape_raises`). Doing the same here keeps the two mixins consistent:

```python
if from_magnitude_errors is not None:
    from_magnitude_errors = np.asarray(from_magnitude_errors)
    if from_magnitude_errors.ndim == 1:
        from_magnitude_errors = from_magnitude_errors[np.newaxis, :]
    if from_magnitude_errors.shape != from_magnitudes.shape:
        raise ValueError(
            "from_magnitude_errors must have the same shape as from_magnitudes; "
            f"got {from_magnitude_errors.shape} and {from_magnitudes.shape}."
        )
```

That preserves the 1-d → 2-d promotion that mirrors the magnitudes handling and rejects everything else.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Agreed on the practical point, and the follow-up is the right calibration: neither in-tree caller can produce integer magnitudes. `transform_refcat2_bands` builds `transform_input` from catalog magnitude columns, which are floats, so this is a hypothetical rather than a live failure.

For the record, the asymmetry is real when someone does hand it integers:

```python
t = PanStarrs1ToJohnsonCousins.load()
imags = np.array([20, 19, 19, 18, 18, 17])
t(imags)                                          # fine -> [20.802 19.48 18.726 18.48]
t(imags, from_magnitude_errors=np.full(6, 0.01))  # UFuncTypeError
```

`np.zeros_like(from_magnitudes)` inherits `int64`, so `partials[...] += dpoly_dcolor` fails, while the magnitudes-only path never allocates from the input dtype.

Since the fix is `np.zeros_like(from_magnitudes)` → `np.zeros(from_magnitudes.shape)` — same line count, no behaviour change for float input — the cost of taking it is about zero, so I'd lean toward making the error path as dtype-forgiving as the magnitudes path. Entirely your call, though; happy to leave it if you'd rather not carry a change for a case that cannot arise.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Agreed, and the existing tests already pin the contract the change would restore, so it can be made without loosening coverage:

- 1-d input → `(4,)` (`test_...` at line 194), and `(4,)` errors (line 285)
- 3-row input → `(3, 4)` (line 202), 2-row errors → `(2, 4)` (line 271)

No test asserts a squeezed result for a `(1, 6)` input — that case is exactly the one currently forcing `np.atleast_2d` at lines 534 and 542. So:

```python
squeeze = np.squeeze if input_was_1d else (lambda x: x)
```

with `input_was_1d` captured where the `np.newaxis` promotion happens, keeps every asserted shape identical and lets both `np.atleast_2d` calls in `transform_refcat2_bands` go away.

Worth flagging one consequence to decide deliberately: this *is* a return-shape change for any external caller that passes a `(1, 6)` array and relies on getting `(4,)` back. `PanStarrs1ToJohnsonCousins` is in `__all__`, so that is a public surface, even if there is no in-tree caller doing it. Given the alternative is the atleast_2d dance spreading to every future call site, the shape-contract fix looks like the better trade — but it belongs in CHANGES.rst if taken.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Confirmed — `USNO_SDSS_to_SDSS_DR7.json` has no `residual` field anywhere in the file, so on the `apply_sdssdr7_transform=True` path the only noise-floor term entering the quadrature is `_JESTER_RMS[band]`. The scatter of the USNO′→SDSS DR7 relation itself is simply not represented, which makes those two-step errors a lower bound rather than a complete budget.

That is not fixable without a published rms for the USNO step (the SDSS DR1 photometric-equations page the JSON references gives the coefficients but no scatter), so stating it is the right response rather than inventing a number.

Suggested comment amendment at line 435:

```python
# The matrix step makes the SDSS g, r and i magnitudes correlated -- each
# is a mix of the same native bands -- so the errors cannot be propagated
# through it and Jester separately. Compose the two linear maps instead:
# the effective coefficient of each native band is the Jester row through
# the matrix. Note the USNO'->SDSS DR7 step has no published rms residual,
# so only the Jester residual enters the quadrature below and the result
# is a floor on the true uncertainty.
```

and the CHANGES.rst line reworded from "each transform's published rms residual is included in quadrature" to something like "each transform's published rms residual is included in quadrature where one exists; the USNO′→SDSS DR7 matrix step has none, so errors on that path are a floor."

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Verified the premise: `filter_transform` does accept both, `supported_transforms = ["jester", "ivezic"]`, and the `ivezic` branch is structurally different — `transform_poly = np.poly1d(transform_ivezic[output_filter])` evaluated on a color, added to a base magnitude. Nothing like the linear-in-g,r,i form the error block assumes. So switching the string would leave `mag_error_RC`/`mag_error_IC` describing a transform no longer being applied, and the failure would be silent: plausible-looking errors, wrong values, fed straight into `transform_to_catalog`'s weighting.

Concrete shape for the fix:

```python
# The error propagation below assumes the transform is linear in the
# native g, r and i. That is true of "jester" and NOT of "ivezic",
# which is a polynomial in colour -- do not change this string without
# reworking the error propagation.
transform_name = "jester"
...
table["mag_RC"] = filter_transform(table, output_filter="R", **use_columns, transform=transform_name)
table["mag_IC"] = filter_transform(table, output_filter="I", **use_columns, transform=transform_name)
...
if have_errors:
    if transform_name != "jester":
        raise NotImplementedError(
            "Error propagation is implemented only for the linear Jester "
            f"transform, not {transform_name!r}."
        )
```

The single variable removes the two-places-to-edit trap and the guard turns a future one-word edit into a loud failure instead of quiet wrong errors.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Agreed — the two branches differ only in the errors kwarg, yet each repeats `transform_input.T` and `ps1_band_for_V="gp1"`, so a change to either of those has two places to land and no test would catch it landing in one.

The single-call version:

```python
error_input = None
if have_errors:
    zeros = np.zeros(num_rows)
    error_input = np.array([
        table["mag_error_SG"],
        table["mag_error_SR"],
        table["mag_error_SI"],
        z_error,
        zeros,  # Placeholder for the y_p1 band
        zeros,  # Placeholder for the w_p1 band
    ]).T

result = transform(
    transform_input.T,
    ps1_band_for_V="gp1",
    from_magnitude_errors=error_input,
)
transformed_data, transformed_errors = result if have_errors else (result, None)
```

That also removes the asymmetric `np.atleast_2d` — both arrays get it in one place afterwards — or drops it entirely if the squeeze contract is fixed in `__call__` per the other thread. Those two changes compose: fix the squeeze and this block becomes a single call with no reshaping at all.

stellarphot/utils/magnitude_system_transforms.py
mwcraig: Yes — it is just a 2×6 Jester matrix times the transform matrix. Verified the vectorized form reproduces the loop exactly:

```python
matrix = USNOPrimeToSDSSDR7.load().transform_information.array  # (6, 6)

jester = np.zeros((2, matrix.shape[0]))
jester[0, 1:4] = _JESTER_TRANSFORMS["R"][:3]
jester[1, 1:4] = _JESTER_TRANSFORMS["I"][:3]
effective = (jester @ matrix)[:, 1:4]
```

gives `R = [0.4346, -0.49741, 1.06281]` and `I = [0.4346, -1.49141, 2.05681]`, identical (`np.allclose`) to the current per-band loop.

Dropped into place:

```python
if have_errors:
    # ... comment as now ...
    matrix = transform.transform_information.array
    # Rows 1, 2 and 3 of the matrix output are g, r and i, the same
    # indexing used for sdss_mags above.
    jester = np.zeros((2, matrix.shape[0]))
    for row, band in enumerate(("R", "I")):
        jester[row, 1:4] = _JESTER_TRANSFORMS[band][:3]
    effective_coeffs = dict(zip(("R", "I"), (jester @ matrix)[:, 1:4], strict=True))
```

Honest assessment of whether it is an improvement: it makes the composition-of-two-linear-maps structure visible as one matrix product rather than something the reader reassembles from a loop, which is the actual argument for it. It does not get shorter — the padding of the Jester rows into the 6-wide space still needs a loop or an explicit literal, and it is only two output bands. So: real readability gain on the maths, roughly neutral on line count. Worth doing if you want the comment's "compose the two linear maps" claim to be literally what the code says.

stellarphot/utils/magnitude_system_transforms.py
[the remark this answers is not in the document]
mwcraig: Change all three

PR #695, opened 2026-08-13
Floor the transform fit sigma and report how each fit was weighted

mwcraig:
Fixes #694.

A catalog error of exactly `0.0` meant "this catalog does not know its own uncertainty for this star", so the star was weighted by its observed error alone. On APASS DR9 that assumption fails badly: it reports `0.000` for most of its B errors, and it does so for the *faint* stars, so the fit ends up weighting the worst-measured stars most heavily. In the V fit from the issue a single star — catalog error zero, observed error 0.00068 mag — held **88.5% of the weight of a fit over 422 stars**, and nothing in the output said so.

## What this does

**1. Floors the fit's *weighting* at `min_fit_sigma`, a new keyword defaulting to 0.01 mag** (0 disables it; negative is rejected). A least-squares weight goes as one over the variance, so a star claiming a sigma far below that does not merely count for a little more — it can run the fit. The floor applies only to the weights: `fit_redchi`, `fit_excess_scatter` and `mag_cal_error` are all measured against the errors as quoted, so an image with badly under-quoted errors still raises the alarm those columns exist for.

**2. Adds three diagnostic columns beside `fit_redchi`**, since a reduced chi-square is a ratio and an image whose errors are wrong in the right way reports a healthy-looking one:

| column | what it catches | reads on the issue's data |
|---|---|---|
| `fit_cat_error_missing_frac` | catalog didn't know its own errors → `fit_redchi` is not comparable across bands | 0.95 (B) vs 0.002 (V) |
| `fit_max_weight_share` | one star running the fit | 0.885 (V) |
| `fit_excess_scatter` | real scatter no weighting can fix (flat-field gradient, catalog photometry) | 0.086 / 0.022 / 0.037 mag |

**3. Documents the 1.0 arcsec fit-selection cut** (`_FIT_MATCH_ARCSEC`), which the docstring never mentioned — only the 1.5 arcsec cut for the match-derived columns. The fit is therefore not reproducible from the output table by selecting on `mag_cat` alone.

**4. Updates the APASS transform notebook.** Its "How well did each image fit?" section told the reader to judge an image by `fit_redchi`, which is exactly the reading the issue showed to be insufficient.

## Two judgment calls worth reviewing

**The floor is applied to both weighted paths, not just the combined one.** The issue's suggested patch touches only the `combined` branch. But `weights = 1.0 / errors[good]` — the path taken when the catalog has no error column for the band, i.e. Johnson-Cousins R and I per #685 — has the identical single-star-domination defect, and the justification for 0.01 (a single ground-based measurement tied to a survey catalog is not credible below about 0.01 mag) does not depend on whether the catalog supplied errors. Flooring only one branch would leave the bug alive for the bands most often calibrated here. Easy to narrow if you disagree.

**Excess scatter is reported, not folded into the weights** — per the issue. A fit that absorbs its own excess scatter has a reduced chi-square of one by construction, which destroys the only diagnostic that surfaced any of this.

## What this does *not* fix

**B.** Its zero-error stars have observed errors around 0.031, well above any sane floor, so the floor barely moves it — their problem is a missing sigma, not a tiny one. `fit_cat_error_missing_frac` reading 0.95 is what makes that visible instead of silent. As the issue says, the real answer there is likely a better catalog.

Also untouched, and flagged in the issue as overlapping #674/#691: `mag_cal_error` still understates the true scatter by 3–5×.

## Behavior change

Two existing tests that asserted two runs were identical across *every* output column now exclude `fit_cat_error_missing_frac` and assert on it explicitly instead — a catalog error of `1e-30` is usable and one of `0.0` is not, and telling those apart is the whole point of that column.

## After review (59ac2f9)

The review round changed the design in one important way: the first version floored the sigma that *also* fed `fit_redchi` and `fit_excess_scatter`, which capped the redchi alarm and could invert it (quoted errors of 0.002 against a true scatter of 0.008 reported redchi ≈ 0.6 and excess ≈ 0 — an affirmative all-clear for exactly the case the diagnostics exist to catch). Now only the weights are floored and both statistics are measured against the errors as quoted; `test_transform_to_catalog_floor_caps_the_reported_redchi` inverted into `test_transform_to_catalog_floor_preserves_the_redchi_alarm`. Also from review: the `min_fit_sigma` keyword, a fix for a rare `brentq` crash when redchi lands within a few ULPs of 1.0, `_fit_diagnostics` taking the usability mask and weights instead of re-deriving them, a `_FIT_DIAGNOSTIC_COLUMNS` tuple, doc fixes ("closer than 1.0 arcsec"; the floor leaves `mag_cal_error` untouched), and the #694 narrative consolidated into the issue with one-line pointers in code.

## Testing

Full suite: **1007 passed, 49 skipped**. 17 new tests across both rounds, written first and confirmed failing — including `fit_redchi` coming out at ~10⁴ where the sigma-floored version reported 4.

---


stellarphot/utils/magnitude_transforms.py
```diff
@@ -1178,14 +1394,29 @@ def transform_to_catalog(
...
+                sigma = np.hypot(errors[good], cat_error[good])
+
+            # Bounded below so that no one star can take the fit over; see
+            # `_MIN_FIT_SIGMA` and issue #694. The floor goes on the total the
+            # fit divides by rather than on either error separately, because
+            # it is the total that decides the star's weight -- and a star
+            # whose catalog error is the unusable zero has a total equal to
+            # its observed error alone, which is exactly the case that needs
+            # bounding.
+            sigma = np.maximum(sigma, _MIN_FIT_SIGMA)
```
mwcraig: **Correctness: flooring the sigma that also feeds `fit_redchi` and `fit_excess_scatter` can invert the under-quoted-errors alarm, not just cap it.**

Take an image whose pipeline quotes 0.002 mag errors while the true scatter about the transform is 0.008 mag. Pre-PR the fit reported `fit_redchi` ≈ 15 (verified numerically) — a clear alarm. Post-PR the sigma is floored to 0.01, so `fit_redchi` ≈ 0.60 and `_excess_scatter`'s `redchi <= 1.0` branch returns exactly 0.0: the new diagnostic affirmatively reports that the errors were adequate, for exactly the case it exists to catch.

The docstring documents only the cap ("can no longer report an arbitrarily large `fit_redchi`"), not this below-1 inversion, and the pre-existing Notes sentence ("the residuals are in units of the errors that were supplied") is now false below the floor.

Suggested fix: floor only the *weights* the fit uses, and compute the reported `fit_redchi`/`fit_excess_scatter` against the unfloored sigmas — that keeps the leverage bound while preserving the alarm.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -270,6 +294,133 @@ def _underdetermined_reason(fit_result, vary):
...
+        return np.sum(residual**2 / (sigma**2 + excess**2)) / fit_result.nfree - 1.0
+
+    # A bracket rather than a guess. The function falls monotonically from a
+    # positive value at zero -- the branch above ruled out the alternative --
+    # and at this upper bound every ``sigma**2 + excess**2`` is at least
+    # ``excess**2``, so the sum is at most ``nfree`` and the function is at
+    # most zero. So a root lies between them.
+    upper = np.sqrt(np.sum(residual**2) / fit_result.nfree)
+
+    return float(brentq(reduced_chi_square_less_one, 0.0, upper))
```
mwcraig, line 443: **Correctness: `brentq` can raise `ValueError` and crash the whole `transform_to_catalog` call when `fit_result.redchi` lands within a few ULPs above 1.0.**

The gate above tests lmfit's `redchi = sum(r**2)/nfree`, but the bracket endpoint `f(0)` recomputes it as `sum((r*sigma)**2/sigma**2)/nfree`, which differs by up to ~4e-16 (verified numerically). A fit with `redchi = 1 + 1e-16` passes the gate while `f(0)` evaluates ≤ 0; with `f(upper)` also negative, `brentq` raises `f(a) and f(b) must have different signs` and the call aborts for a perfectly healthy image instead of reporting ~0 excess scatter.

Rare, but it's an unhandled crash. Either test `f(0)` directly for the gate, or catch the `ValueError` and return 0.0.

stellarphot/utils/magnitude_transforms.py
mwcraig: **The credibility floor is applied to the fit weights but not to `star_errors`, so `mag_cal_error` still carries the sub-millimag error the floor's own rationale declares not credible.**

The issue-#694 star with a 0.00068 mag observed error: its fit weight is clamped here as unbelievable, yet `star_errors` (line ~1508, outside this diff) passes 0.00068 unfloored into `_calibrated_with_uncertainty`, so the output table's `mag_cal_error` measurement term — which the AAVSO writer submits as a real uncertainty — is the very number the `_MIN_FIT_SIGMA` comment says "is not reporting a measurement that good."

If that's intentional (floor the *leverage*, report the *claim*), the scope of the floor deserves a sentence in the docstring; if not, the same floor belongs there too.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -783,6 +936,13 @@ def transform_to_catalog(
         NaN wherever ``mag_cal`` is, for whatever reason, so a finite
         calibrated error always has a calibrated magnitude to go with it.
 
+        A star is allowed to *influence the fit*, as opposed to be given
+        values derived from its match, only if its nearest catalog entry is
+        within the tighter limit of 1.0 arcsec. So the stars behind the
```
mwcraig: **Doc/code boundary mismatch: "within the tighter limit of 1.0 arcsec" but the code cut is strictly less-than.**

The fit cut uses `d2d.arcsecond < _FIT_MATCH_ARCSEC` while the calibration cut deliberately uses `<=`. A star matched at exactly 1.000 arcsec is documented here as entering the fit but is excluded, so a user reproducing the fit's star set from the documented rule selects one star too many — the same doc-vs-code imprecision this PR's changelog entry exists to fix (it advertises correcting 1.5 → 1.0 but reintroduces an off-by-boundary).

stellarphot/utils/magnitude_transforms.py
```diff
@@ -47,6 +48,29 @@
...
+# single such star can take over the fit -- 88.5% of one on the data behind
+# issue #694, at a claimed sigma of 0.00068 mag. The floor is not tuned to that
+# dataset: a single ground-based measurement tied to a survey catalog is not
+# credible below about 0.01 mag whatever the catalog says, and 0.01 is the
+# smallest value that bounds the leverage in every band that was checked.
+#
+# Applied to the total sigma the fit divides by, so it bounds the weight of a
+# star however its sigma got small -- a zero catalog error, a tiny observed
+# error, or a band whose catalog has no error column at all.
+_MIN_FIT_SIGMA = 0.01
```
mwcraig, line 64: **`_MIN_FIT_SIGMA = 0.01` is a hard-coded private constant with no parameter or opt-out, though its own comment justifies it only for single ground-based measurements tied to survey catalogs.**

A user calibrating bright stars against refcat2 (which quotes millimag errors) or using stacked photometry has legitimate combined sigmas of 0.003–0.008 mag; the floor silently flattens the weights of exactly their best-measured stars and computes `fit_redchi` against inflated sigmas, with no way to disable it short of monkeypatching a private constant.

`transform_to_catalog` already exposes comparable knobs (`vary`, `expected`, `fit_diff`); a `min_fit_sigma` keyword defaulting to 0.01 is the same fix at the right depth.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -270,6 +294,133 @@ def _underdetermined_reason(fit_result, vary):
...
+        Keyed by the three column names, ready to be written to the rows of
+        this image.
+    """
+    # A catalog with no error column for the band knows nothing about any of
+    # these stars, which is the same statement as a column of zeros rather than
+    # a question that does not apply, so it reads as "all of them missing".
+    if cat_error_for_fit is None:
+        missing_fraction = 1.0
+    else:
+        missing_fraction = float(np.mean(cat_error_for_fit == 0.0))
```
mwcraig: **`_fit_diagnostics` reconstructs caller state instead of receiving it.**

The usability decision already exists as a boolean at the call site (`np.isfinite(cat_error) & (cat_error > 0)`); decoding it here from `cat_error_for_fit == 0.0` couples the two sites by convention only — a future change of the fallback value (say, band-median error instead of 0.0) silently makes `fit_cat_error_missing_frac` read 0.0 everywhere while every weight test still passes.

Likewise the `broadcast_to(asarray(...)**2, ...)` trick re-derives "what weight did the fit use" from sigma when the caller already holds `weights = 1.0 / sigma`. Passing the mask (or count) and the weights in keeps one definition of each.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -972,6 +1180,14 @@ def transform_to_catalog(
     coefficients = {name: np.full(n_rows, np.nan) for name in _COEFF_NAMES}
     coefficient_errors = {name: np.full(n_rows, np.nan) for name in _COEFF_NAMES}
     fit_redchis = np.full(n_rows, np.nan)
+    fit_diagnostics = {
+        name: np.full(n_rows, np.nan)
+        for name in (
+            _FIT_CAT_ERROR_MISSING_COLUMN,
```
mwcraig: **The three diagnostic column names are spelled out in three places** (the constants block, the `_fit_diagnostics` return dict, and this tuple) with nothing tying them together — plus a fourth copy in the test file's `_FIT_COLUMNS`.

Adding or renaming a diagnostic takes three synchronized edits, and a missed one only surfaces as a runtime `KeyError` in the per-image write loop the first time a fit succeeds. A module-level `_FIT_DIAGNOSTIC_COLUMNS` tuple next to the name constants — mirroring how `_COEFF_NAMES` already serves the coefficients dicts one line above — removes the duplication.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1252,6 +1288,204 @@ def test_transform_to_catalog_reports_unweighted_fit_statistic(mocker):
...
+    # the column says exactly zero rather than a small positive number that
+    # would read as a real finding.
+    result, _, _ = _fit_a_catalog(
+        mocker, n_stars=200, sigma=0.02, seed=31415, mag_error=0.05, cat_error=None
+    )
+
+    assert result["fit_excess_scatter"][0] == 0.0
+
+
+def test_transform_to_catalog_diagnostics_for_an_unweighted_fit(mocker):
```
mwcraig: **This test hand-rolls the identical fit setup as the existing `test_transform_to_catalog_reports_unweighted_fit_statistic`** — same `n_stars=100`, `noise_sigma=0.02`, `seed=13579`, `pytest.warns`, `obs_error_column=None`.

The same ~10-line setup now lives in two tests computing the same fit twice, and the copies can silently drift (change one seed and they stop testing the same fit). Worth extracting a shared fixture (cf. the module-scoped `_noisy_fit_result`) and refactoring the older test alongside this one.

CHANGES.rst
```diff
@@ -33,6 +33,14 @@ New Features
   too. The USNO'-to-SDSS DR7 step used by
   ``transform_apass_bands(apply_sdssdr7_transform=True)`` has no published
   residual, so the errors on that path are a floor. [#685]
++ ``transform_to_catalog()`` now reports how each image's fit was weighted,
```
mwcraig, line 36: **This entry runs 8 lines and restates all three column definitions plus the rationale.** The house style for this changelog is what changed, the user-visible effect, and the issue number — pointing to the docstring for the why. The inline definitions and the closing editorial line ("A `fit_redchi` near one is not on its own evidence of a healthy fit.") already live in the `transform_to_catalog` docstring. The two Bug Fixes entries are the right size.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -47,6 +48,29 @@
 # Name of the column reporting how well one image's transform fit its stars.
 _FIT_REDCHI_COLUMN = "fit_redchi"
 
+# Smallest uncertainty, in magnitudes, that a star is allowed to claim when the
+# fit weights it. A star weighted by a sigma far below this is not reporting a
+# measurement that good; it is reporting one whose error bookkeeping broke
+# down, and because a least-squares weight goes as one over the variance, a
+# single such star can take over the fit -- 88.5% of one on the data behind
+# issue #694, at a claimed sigma of 0.00068 mag. The floor is not tuned to that
```
mwcraig: **The #694 war story, with its specific numbers (88.5% weight share, redchi 1.87, 0.00068 mag, 422 stars), is retold in five places:** this comment, the `_fit_diagnostics` docstring, the `transform_to_catalog` Notes, a test-section comment, and the notebook markdown.

Five hand-copied retellings will drift the first time anyone refines the numbers, and a future editor can't tell which copy is authoritative — roughly 60 lines of duplicated prose for content the issue tracker already records. Keeping the full narrative once in the user-facing Notes and reducing the other four to one line ("one star held ~88% of a fit's weight; see issue #694") preserves the pedagogy without the drift risk.

mwcraig (commented):

Critical review of the sigma floor + fit diagnostics. Ten findings, posted inline; the two that matter most are correctness issues that were verified numerically rather than eyeballed:

1. **The floor can invert the under-quoted-errors alarm, not just cap it** (comment on the `np.maximum` line). Quoted errors below the floor make a genuinely bad fit report `fit_redchi` *below* one and `fit_excess_scatter` of exactly 0.0 — the diagnostic affirmatively vouches for the errors it exists to distrust.
2. **`brentq` can raise `ValueError` and abort the whole call** when `fit_result.redchi` lands within a few ULPs above 1.0, because the gate and the bracket endpoint compute redchi two slightly different ways.

The rest: the floor doesn't reach `star_errors`/`mag_cal_error`; a docstring boundary mismatch (`within` vs strict `<`); the floor has no opt-out; `_fit_diagnostics` decodes caller state from the `== 0.0` sentinel; the diagnostic column names are spelled in three places; the new unweighted-fit test duplicates an existing test's setup; the CHANGES entry runs long; and the #694 narrative is retold in five places.

Refuted along the way (not posted): broadcast-shape/masked-array concerns (the code paths guarantee plain floored arrays) and the apparent meta-channel contradiction (deliberate and documented).

stellarphot/utils/magnitude_transforms.py
```diff
@@ -835,6 +995,54 @@ def transform_to_catalog(
     ``fit_redchi`` well above one, and where the catalog's uncertainty is the
     reason, weighting cannot know that but the scatter still shows up there.
 
+    Whichever way it is weighted, the total uncertainty the fit divides each
+    star's residual by is bounded below at 0.01 mag. A least-squares weight
```
mwcraig: This needs to be updated if the uncertainty boundary becomes a parameter

stellarphot/utils/magnitude_transforms.py
```diff
@@ -835,6 +995,54 @@ def transform_to_catalog(
...
+    goes as one over the variance, so a star claiming a sigma far below that
+    does not merely count for a little more -- it can run the fit. On the data
+    behind issue #694 a single star with a catalog error of zero and an
+    observed error of 0.00068 mag held 88.5% of the weight of a fit over 422
+    stars, and the ``fit_redchi`` of 1.87 that fit reported looked healthier
```
mwcraig: Yikes, this is the full history of this. Not needed.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1208,7 +1239,12 @@ def test_transform_to_catalog_reports_reduced_chi_square(
 ):
     # See issue #677.
     n_stars = 100
-    sigma = 0.02
+    # Large enough that ``error_scale * sigma`` stays above `_MIN_FIT_SIGMA`
```
mwcraig: If _MIN_FIT_SIGMA becomes a parameter this test needs to change.

stellarphot/utils/magnitude_transforms.py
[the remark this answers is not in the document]
mwcraig: This behavior is deliberate, documented, and tested — the column's stated meaning does cover it. The `transform_to_catalog` docstring defines `fit_cat_error_missing_frac` as "the fraction of the stars in the fit whose catalog error the fit could not use" and then says explicitly: *"It is 1.0 when the catalog has no error column for the band and when the fit is unweighted, both of which are the same statement -- nothing was known about any of them."* The `_fit_diagnostics` parameter doc likewise names the unweighted case as one of the meanings of `None`, and `test_transform_to_catalog_diagnostics_for_an_unweighted_fit` asserts the 1.0.

The column reports what the *fit* used, not what the catalog contains — it lives in a constants block headed "Names of the columns describing how the fit was weighted." An explicitly unweighted fit used none of the catalog errors, however valid they were, and the warning 1.0 carries is exactly right for it: the column's purpose is "this image's `fit_redchi` is not comparable across bands," and an unweighted `fit_redchi` (mag², not a chi-square) is the least comparable of all. Preserving and reporting catalog errors the fit deliberately ignored would break the column's own charter.

Nor is the unweighted case ambiguous in the output: `fit_excess_scatter` is NaN only for unweighted fits, and `result.meta["transform_weighting"]` records `"unweighted"` per band.

Happy to switch the unweighted case to NaN if @mwcraig prefers, but it would cost the don't-compare-this signal for the fits where it is most warranted.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed — reproduced independently rather than taken on faith: 300 stars with quoted errors of 0.002 mag and true scatter of 0.008 give a pre-floor `fit_redchi` of 13.9; with the floor at 0.01 the same fit reports 0.56, and `_excess_scatter`'s `redchi <= 1.0` branch returns exactly 0.0. The floor as written converts a clear alarm into an affirmative all-clear, for precisely the case the diagnostic exists to catch.

Agreed fix, as suggested: keep the floor on the weights the minimizer uses — that bounds the leverage, which is all #694 needs — and compute the *reported* `fit_redchi` and `fit_excess_scatter` against the unfloored sigmas, recovered after the fit from `fit_result.residual * sigma_floored` and divided by the raw sigma. The raw sigma is always positive on both weighted paths (observed errors are filtered to `> 0`; unusable catalog errors become 0 inside the quadrature sum), so the recomputation cannot divide by zero. On the example above this reports `fit_redchi` ≈ 14 and an excess scatter of ≈ 0.0077 mag — the alarm survives.

Consequences to fold in: the Notes sentence about residual units becomes true again; the docstring gains the below-the-floor case; and `test_transform_to_catalog_floor_caps_the_reported_redchi` inverts — the cap was a side effect this fix removes, so it becomes a test that the alarm is preserved.

Not applied on the branch yet; will push with the other agreed fixes on request.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed — the mechanism reproduces: constructing weighted residuals whose lmfit-style `redchi` lands a few ULPs above 1.0, `f(0) = sum((r*sigma)**2 / sigma**2) / nfree - 1` evaluated ≤ 0 in roughly 6% of trials, which together with `f(upper) ≤ 0` is exactly the `f(a) and f(b) must have different signs` `ValueError` from `brentq`, aborting the whole call for a healthy image.

Agreed fix: gate on the bracket function itself — evaluate `reduced_chi_square_less_one(0.0)` once and return 0.0 when it is ≤ 0 — so the gate and the bracket endpoint are the same computation and cannot disagree in sign. This change becomes necessary anyway under the fix agreed in the sigma-flooring thread, since the reported statistic will no longer be lmfit's `fit_result.redchi`.

stellarphot/utils/magnitude_transforms.py
mwcraig: The scope is intentional: floor the *leverage*, report the *claim*. The floor exists to stop one star outvoting the rest — a statement about relative weight inside a single fit. `mag_cal_error` is a reported measurement, and rewriting it is a different contract: that column is already known to understate the true uncertainty by 3–5× (#674/#691), and clamping its measurement term up to 0.01 would layer a second silent adjustment onto a number those issues say is wrong for a different reason. The principled fix — folding the measured scatter, e.g. `fit_excess_scatter`, into the reported error — belongs to those issues, not to a floor whose rationale is about fit weighting.

Agreed on the ask, though: the docstring should say so. Will add a sentence stating that the floor applies only to the fit's weighting and leaves `mag_cal_error` and every reported per-star error untouched, alongside the other agreed edits.

stellarphot/utils/magnitude_transforms.py
mwcraig: Valid — `d2d.arcsecond < _FIT_MATCH_ARCSEC` is strict while the 1.5 arcsec calibration cut is deliberately `<=`, so "within" overstates the fit's star set by the boundary point. Will reword the docstring to "closer than 1.0 arcsec"; the 1.5 arcsec wording ("more than 1.5 arcsec away" → excluded) already matches its `<=` and stays.

Fixing the word rather than flipping the comparison to `<=`: the strict cut predates this PR, and changing it would silently change which stars enter the fit on existing data for the sake of a boundary case that is measure-zero in practice.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — and the refcat2-bright-stars case is exactly the argument: the 0.01 rationale is a claim about a *single ground-based measurement*, and stacked photometry breaks its premise, so the knob belongs at the same depth as `vary` and `expected`. The maintainer's follow-up comments on the docstring and on `test_transform_to_catalog_reports_reduced_chi_square` already anticipate this.

Plan: a `min_fit_sigma=0.01` keyword on `transform_to_catalog`, validated non-negative, with `0` meaning no floor; `_MIN_FIT_SIGMA` remains as the default's single home. The Notes sentence will name the parameter instead of a bare 0.01, and the reduced-chi-square test will pin `min_fit_sigma` explicitly instead of arranging its errors to dodge the default.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed on both couplings: the `== 0.0` sentinel decodes by convention what the call site already knows as `np.isfinite(cat_error) & (cat_error > 0)`, and the `broadcast_to(asarray(...)**2, ...)` expression re-derives the weights the caller already holds. Will change `_fit_diagnostics` to take the usability mask (restricted to the fitted stars) and the weights, so each fact has exactly one owner — and the weights version handles the unweighted scalar `1.0` for free, which also drops the `1.0 if sigma is None else` dance.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed on the fix, with one correction to the record: a *rename* is already a single edit — all three module sites spell the names via the `_FIT_*_COLUMN` constants, so nothing desynchronizes on rename. The real hazard is the one this comment ends on: *adding* a diagnostic takes a synchronized edit to the return dict and to this tuple, and a miss surfaces only as a runtime `KeyError` in the per-image write loop. A module-level `_FIT_DIAGNOSTIC_COLUMNS` tuple next to the constants, mirroring `_COEFF_NAMES`, closes that cheaply — will add it and use it here and in the output-column block.

The test file's `_FIT_COLUMNS` copy is worth keeping independent, though: a test that imports the production tuple can no longer notice the production names changing out from under users.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — same `n_stars=100`, `noise_sigma=0.02`, `seed=13579`, same `pytest.warns` guard, same `obs_error_column=None`; the two tests compute the identical fit twice and the copies can drift silently. Will extract the shared setup as a module-scoped fixture that computes the unweighted fit once (in the style of `_noisy_fit_result`) and refactor `test_transform_to_catalog_reports_unweighted_fit_statistic` onto it in the same change, per the house rule that the older twin gets refactored alongside the new one.

CHANGES.rst
mwcraig: Agreed — the entry restates definitions that live in the `transform_to_catalog` docstring, and the closing editorial line is the docstring's job too. Will trim it to the house shape: the three column names, the one-line user-visible point, and the issue number, matching the length of the neighboring entries.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed on consolidating — and @mwcraig's comment on the Notes paragraph ("this is the full history of this. Not needed.") settles where the one authoritative copy lives: in issue #694 itself, not anywhere in the code. So the plan goes one step further than this comment suggested: every in-code retelling — this constants comment, the `_fit_diagnostics` docstring, the `transform_to_catalog` Notes, the test-section comment — drops to the one-line form with a pointer to #694, and the notebook keeps a short reader-facing caution without the history. The specific numbers then exist in exactly one place, which is the place that already records them.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — folded into the `min_fit_sigma` parameterization discussed on the `_MIN_FIT_SIGMA` thread: this sentence will name the parameter and its default of 0.01 instead of a bare number, and will say what passing `0` does.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — will cut this down to the one-sentence rationale plus a pointer to issue #694, and give the other in-code retellings (constants comment, `_fit_diagnostics` docstring, test-section comment) the same treatment, per the duplication thread on the constants block. The issue tracker keeps the full history; the code keeps the reason.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — once `min_fit_sigma` is a keyword this test will pin it explicitly (small enough that both error scales measure the reduced chi-square directly, or `0` for no floor), and this comment about arranging the errors to stay above the default goes away with it.

PR #696, opened 2026-08-14
Believe the quoted errors in the transform fit's reported uncertainties

mwcraig:
Fixes #690.

`transform_to_catalog` now calls `lmfit.minimize` with `scale_covar=False` for weighted fits, so the reported `a_error`…`z_error` and `mag_cal_error` come from the covariance of the fit **as it was weighted** — the quoted errors, floored at `min_fit_sigma` — instead of being rescaled by the reduced chi-square to track the observed scatter.

## Why

Condensed from the discussion on #690:

- **#695 already chose this philosophy in the same function.** It floors only the *weights* and measures `fit_redchi`/`fit_excess_scatter` against the errors as quoted: believe the inputs, report the discrepancy, never silently correct it. Leaving `scale_covar` on meant one output table simultaneously reported the discrepancy in `fit_redchi`/`fit_excess_scatter` and silently corrected for it in the `*_error` columns — one observation dressed as two agreeing ones.
- **Post-#695 the rescaling was incoherent anyway**: lmfit's internal redchi is computed against the *floored* weights, which is neither the reported `fit_redchi` nor the observed scatter. The docstring used to apologize for this; now it doesn't have to.
- **Nothing is lost.** The excess scatter the rescaling absorbed implicitly is reported explicitly in `fit_excess_scatter`, as an honest number the caller can choose to propagate; `fit_cat_error_missing_frac` flags the missing-error case.
- **It makes the #680/#692 error plumbing visible.** Under the old default, a uniform rescaling of every weight was a no-op on the reported uncertainties (covar × redchi is invariant), so getting the quoted errors right could never move them. There is now a test asserting that tripling the quoted errors triples every reported uncertainty.
- **#691's color term inherits a consistent convention**: all of `mag_cal_error`'s terms now share one epistemology, and `fit_redchi` becomes an independent diagnostic.

**One deliberate deviation from a blanket `scale_covar=False`**: an unweighted fit (no `obs_error_column`) quotes no errors to believe — its weights are 1, so an unscaled covariance would price every star at 1 mag. Those fits keep the rescaling (`scale_covar=sigma is None`), and a test pins their `*_error` to `(JᵀJ)⁻¹ × fit_redchi`.

## Honest cost

For images with `fit_redchi` > 1 the reported `*_error` and `mag_cal_error` **shrink**, and #694 showed `mag_cal_error` already understates the true scatter 3–5× on real data. But the rescaling only recovered about `excess × sqrt(p/N)` of that (~1/12 on the #694 data); the real fix is propagating `fit_excess_scatter` into `mag_cal_error`, which is a #694 follow-up, not this convention.

A second cost, surfaced by adversarial review: "believe the errors as the fit was weighted" means the `min_fit_sigma` floor now reaches the reported uncertainties. When every star's quoted error genuinely sits under the floor (stacked photometry of bright stars), the reported errors come out at the floor's scale — overstated by about floor/quoted — where the old rescaling cancelled the floor out exactly. `min_fit_sigma=0` is the documented escape; the docstrings now say this in all three places that previously claimed the floor could not touch reported errors, and a test pins the floor-engaged covariance so the trade-off stays deliberate.

## Testing

Written test-first (confirmed red before the change):

- Under-quoted errors (true scatter 5× the quoted): `a_error`/`c_error`/`z_error` match a hand-computed `(JᵀWJ)⁻¹` from the quoted errors to 1e-6 while `fit_redchi` still reports ≈25; `mag_cal_error` matches the same unscaled covariance pushed through the model gradient, correlations included.
- Uniform rescale: tripling the quoted errors triples every reported uncertainty (structurally impossible before).
- Unweighted fits keep the scatter-scaled covariance.
- Three existing tests silently relied on the old rescaling collapsing the transform error term to ~0 on noiseless fits (redchi ≈ 1e-20); they were rewritten to state what they actually test — the measurement half is isolated by a twin fit sharing the same floored weights, and the unusable-row tests now assert surviving rows are bit-identical to a run that never carried the spoiled rows.

Full suite: 1011 passed, 49 skipped (baseline at #695 merge: 1007/49; the four new tests account for the difference).

https://claude.ai/code/session_01UhAvKVUNXqqKYwzwr3La9d


mwcraig (commented):
*Superseded by the inline review below — the same findings, rewritten as comments anchored to the lines they discuss. (Both written by Claude at @mwcraig's direction.)*

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1519,11 +1536,20 @@ def transform_to_catalog(
...
+        # folded silently into the coefficient uncertainties. An unweighted
+        # fit quotes no errors to believe, so the observed scatter is the
+        # only scale its covariance can take and the rescaling stays on.
+        # See issue #690.
         fit_result = lmfit.minimize(
             _transform_residual,
             params,
             method="least_squares",
             args=(fit_mag, fit_color, fit_data, weights),
+            scale_covar=sigma is None,
```
mwcraig, line 1606: **Verified — the core change is sound.** Three refutation attempts against this line all failed:

- `sigma is None` coincides *exactly* with the unweighted path: `sigma` is reassigned every image from `obs_error_column`, which is constant for the whole call, so no weighted-but-`None` (or stale-variable) state exists.
- lmfit honors `scale_covar` for `method="least_squares"` — checked down to the pinned version's source (`minimizer.py`, the flag reaches the covariance construction from the `least_squares` path).
- The reported covariance was confirmed numerically to be `(JᵀWJ)⁻¹` with the floored weights, by hand-computation independent of lmfit (linear model, so the prediction is exact).

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1118,18 +1125,24 @@ def transform_to_catalog(
...
-    where the weighting stops following the quoted errors and the reported
-    ``fit_redchi`` does not. They therefore describe the scatter actually
-    observed about the fit rather than the instrumental errors that were
-    quoted, which is the more robust of the two: it stays right when the
-    quoted errors are systematically wrong. The price is that ``fit_redchi``
-    and the ``*_error`` columns are not independent diagnostics. An image
-    whose stars scatter twice as far from the model as they claim reports both
-    a ``fit_redchi`` near four and coefficient uncertainties twice as large,
-    and that is one observation rather than two agreeing ones.
+    The reported uncertainties believe the errors the fit was weighted by --
```
mwcraig, line 1196: **The one substantive discovery of the review lives in this paragraph.** "Floored at `min_fit_sigma`" is doing real work: under the old rescaling the floor cancelled out of `covar × redchi` exactly, so it provably could not touch any reported error. Now it can. Demonstrated: 60 stars with *truthful* 0.001 mag errors (below the 0.01 floor) report `z_error` 10× too large and `mag_cal_error` 2.7× too large while `fit_redchi` = 0.78 looks perfectly healthy — no diagnostic flags that the floor took over the covariance scale, because the excess-scatter alarm only fires for *under*-quoted errors.

This is inherent to "believe the errors as the fit was weighted", not a bug — but the initial commit inherited three #695-era doc claims saying the floor touches nothing reported, which this change falsified. 7ef42c6 corrected all of them and pinned the floor-engaged covariance with a test; `min_fit_sigma=0` is the documented escape for genuinely sub-floor errors.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -989,9 +990,13 @@ def transform_to_catalog(
         believable below the default, e.g. stacked photometry of bright
         stars against a catalog with trustworthy millimagnitude errors. Must
         not be negative, and has no effect on an unweighted fit. The floor
-        applies only to the fit's weighting: ``fit_redchi``,
-        ``fit_excess_scatter`` and every reported per-star error, including
-        ``mag_cal_error``, are measured against the errors as quoted.
+        applies to the fit's weighting, and through it to the covariance
```
mwcraig: This paragraph (and the module comment on `_MIN_FIT_SIGMA`, and the Notes paragraph below) said the opposite in the initial commit — "the floor applies only to the fit's weighting: … every reported per-star error … measured against the errors as quoted", which `scale_covar=False` falsified. All three sites now tell the same story, including the overstatement trade-off and the `min_fit_sigma=0` escape.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1145,19 +1158,23 @@ def transform_to_catalog(
...
-    those quoted errors. So an image with a ``fit_redchi`` far from one is
-    reporting calibrated errors whose two halves disagree about how much to
-    trust the input, which is the reason to read that column.
+    transform, which lands in ``fit_excess_scatter``; and its systematic tie
+    to the standard system -- about 0.02 mag for APASS DR9 -- which is
+    identical for every star in every image, so a per-star column would
+    mislead, appearing to average down by the square root of the number of
+    stars.
+
+    Both halves of ``mag_cal_error`` therefore believe the errors the fit
```
mwcraig: The sub-floor caveat here matters for anyone using the √`fit_redchi` rule of thumb: `fit_redchi` is measured against the *raw* quoted sigmas while the transform half of `mag_cal_error` believes the *floored* ones, so below the floor the two disagree by ~floor/quoted and the correction overcorrects. The notebook's diagnostics cell carries the same caveat now.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1229,34 +1229,322 @@ def test_transform_to_catalog_uncertainty_falls_as_stars_are_added(mocker):
...
+        Predicted ``mag_cal_error`` of each star.
+    """
+    gradients = _design_matrix(result)
+    transform_var = np.einsum("si,ij,sj->s", gradients, covariance, gradients)
+    measurement_var = ((1 + result["a"][0]) * mag_error) ** 2
+
+    return np.sqrt(measurement_var + transform_var)
+
+
+def test_transform_to_catalog_uncertainties_believe_the_quoted_errors(
```
mwcraig, line 1535: **Non-circularity of the new tests was mutation-verified in both directions.** The hand-built `(JᵀWJ)⁻¹` prediction shares no code with lmfit; forcing `scale_covar=True` at session level kills exactly this test plus the unscaled-covariance and uniform-rescale tests (the unweighted test still passes), and forcing `False` everywhere kills exactly the unweighted test. One caveat for future refactors: a naive *function-scoped* patch does not reach the module-scoped `_underquoted_fit_result` fixture — guards moved onto module fixtures need session-level mutation to exercise.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1229,34 +1229,322 @@ def test_transform_to_catalog_uncertainty_falls_as_stars_are_added(mocker):
...
+    covariance = _predicted_coefficient_covariance(result, sigma_quoted)
+
+    np.testing.assert_allclose(
+        np.asarray(result["mag_cal_error"]),
+        _predicted_mag_cal_error(result, covariance, _UNDERQUOTED_CLAIMED_ERROR),
+        rtol=1e-6,
+    )
+
+
+def test_transform_to_catalog_uncertainties_believe_the_floor_when_it_binds(mocker):
```
mwcraig, line 1582: Added in 7ef42c6/632ffe7 after the adversarial pass showed the floor-engaged regime was otherwise unpinned: every other #690 test quotes errors above the floor, and the twin-fit tests cancel the transform term by construction, so a regression that unfloored the covariance path (or floored the measurement half) would have passed everything. This test pins both the coefficient covariance and `mag_cal_error`'s absolute decomposition below the floor.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1751,25 +2039,70 @@ def test_transform_to_catalog_not_in_place_leaves_input_alone(mocker):
         np.testing.assert_array_equal(observed[name], values)
 
 
-def test_transform_to_catalog_error_reduces_to_instrumental_for_a_perfect_fit(mocker):
+def _measurement_half_of_the_error(mocker, **fit_kwargs):
```
mwcraig, line 2190: **Why the twin-fit construction replaced the old perfect-fit pins:** the old tests asserted `mag_cal_error == (1+a)·mag_error` on noiseless fits, which only held because the rescaling multiplied the transform term by redchi ≈ 1e-20 — i.e. any propagation bug in that term was multiplied by ~0. That property is unpreservable by design now. The subtraction here pins the measurement half exactly (atol 1e-8), and the transform half is pinned harder than before by the unscaled-covariance test (rtol 1e-6, correlations included). Verified: both runs share weights bit-for-bit (both quoted sigmas at/under the floor, `cat_error=None` keeps the catalog from lifting either off it), so the transform term cancels exactly rather than approximately.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2817,6 +3142,30 @@ def _observations_with_unusable_rows(case, ra, dec, instrumental):
     return observed
 
 
+def _assert_rows_blind_to_the_spoiled_ones(mocker, catalog, observed, result):
```
mwcraig, line 3293: Strictly stronger than what it replaced: the old assertion checked surviving rows against a constant at atol 1e-8 (again relying on the transform term vanishing); this compares every output column — per-star and per-image diagnostics alike — bit for bit against a run that never carried the spoiled rows, so contamination cannot cancel and a leak into the diagnostics-only columns is caught too.

CHANGES.rst
```diff
@@ -131,6 +131,13 @@ Other Changes and Additions
   settings have been migrated to the new format. [#666]
 + ``stellarphot.transit_fitting`` now imports its transit-model classes
   lazily. [#666]
++ ``transform_to_catalog()`` no longer rescales the fit covariance by the
```
mwcraig, line 134: **The one consequence deliberately not fixed in this PR:** the AAVSO writer (`io/aavso.py`, used with `mag_error_column="mag_cal_error"` as the user guide instructs) will submit MAGERR values up to ~√`fit_redchi` smaller than before for high-redchi images, with no export-time guard — it formats whatever it is given. The remedy is propagating `fit_excess_scatter` into `mag_cal_error` (the #694 follow-up), not reverting this convention: the old rescaling only recovered ~1/12 of the known understatement anyway. Worth deciding consciously whether that follow-up lands before the next release. Relatedly, the acceptance test proposed on #691 was run against the NGC 6946 data (results posted there): catalog color errors are *not* what `fit_excess_scatter` is made of, so the excess propagation is the right follow-up and cannot double-count against #691.

mwcraig (commented):
## Critical review of this PR — inline edition

Two independent passes were run against the initial commit (15c0d42): a multi-agent code review at high effort, and an adversarial pass instructed to refute the change's safety — find any consumer assuming the old redchi-scaled semantics, and any test weakened rather than corrected. Findings were fixed in 7ef42c6 and 632ffe7; the inline comments below anchor each verified claim and each finding to the code it is about.

Overall verdict: the core change survived every refutation attempt (details on the `scale_covar` line), the rewritten tests are as strong or stronger than what they replaced (details on the test helpers), no downstream consumer encodes the old semantics, and the one substantive discovery — the `min_fit_sigma` floor now reaches the reported uncertainties — is documented, pinned by tests, and disclosed rather than papered over (details on the docstring). The single accepted-not-fixed consequence, for AAVSO submissions, is anchored to the changelog entry.

*This review was run and written by Claude at @mwcraig's direction.*

stellarphot/utils/magnitude_transforms.py
mwcraig: Re-verified independently in this session rather than taken on faith:

- The `sigma is None` ⇔ unweighted coincidence holds by construction: `sigma` is assigned in exactly one place per image (lines 1493–1506), and the branch taken is decided by `obs_error_column is None`, which is constant for the whole call. No path leaves `sigma` stale from a previous image.
- lmfit honoring the flag was confirmed two ways in the pinned version (1.3.4): reading the source — `_calculate_uncertainties_correlations` applies `covar *= redchi` only under `if self.scale_covar`, shared by the `least_squares` path — and empirically, by a session-level patch forcing `scale_covar=True`, which flips all four weighted-covariance tests red in 0.3 s. The flag demonstrably reaches this call site.

Nothing to change; the verification stands.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed, including the arithmetic: with quoted errors at 0.001 and the floor at 0.01, the covariance is the floor's, so the 10× on `z_error` is exactly floor/quoted — the demonstration generalizes, it isn't a quirk of that seed. The regime is now pinned by `test_transform_to_catalog_uncertainties_believe_the_floor_when_it_binds` (quoted 0.002, floor 0.01, predicted 5× checked at rtol 1e-6), and that test dies under a forced `scale_covar=True` mutation, so it guards the actual covariance path.

The "no diagnostic fires" point is worth keeping alive: `fit_redchi` and `fit_excess_scatter` only alarm on *under*-quoted errors, and `fit_max_weight_share` can't see the floor either. A cheap follow-up would be a per-image "fraction of fitted stars whose sigma sat at the floor" column — near 1 is precisely the regime where this overstatement (and the `min_fit_sigma=0` escape) applies. Happy to file that as an issue if wanted; it is deliberately not in this PR.

stellarphot/utils/magnitude_transforms.py
mwcraig: Checked all three sites at head (d3facb2) and they now agree: this `min_fit_sigma` docstring paragraph, the `_MIN_FIT_SIGMA` module comment ("It floors the weighting, and with it the covariance the reported uncertainties come from"), and the Notes paragraph on `mag_cal_error`'s two halves. All three state the same trade-off — sub-floor quoted errors come out overstated by about floor/quoted — and all three name `min_fit_sigma=0` as the escape while keeping `fit_redchi`/`fit_excess_scatter` measured against the errors as quoted. No contradictory claim survives.

stellarphot/utils/magnitude_transforms.py
mwcraig: Verified in the code: `_reported_redchi` divides the residuals by the *raw* quoted sigmas (undoing the floored weighting first), while the covariance behind the transform half comes from `weights = 1 / max(sigma, min_fit_sigma)`. So the disagreement below the floor is real and is ~floor/quoted, exactly as this paragraph says. The notebook's "How well did each image fit?" cell was checked too — it carries the same caveat, with the `min_fit_sigma=0` escape.

One nuance worth stating for readers of this thread: the two scales differing below the floor is a *feature* of the split, not an inconsistency — measuring `fit_redchi` against the floored sigmas instead would call badly under-quoted errors healthy (the comment above the flooring in the code says why). The rule-of-thumb caveat is the price of keeping the alarm honest.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Reproduced independently in this session with a fresh session-level pytest plugin (patching `lmfit.minimize` to force `scale_covar=True` before any test module imports): 4 failed, 1 passed in 0.3 s. The kill set is the three tests this comment names *plus* `test_transform_to_catalog_uncertainties_believe_the_floor_when_it_binds` — added in 632ffe7 after this comment's pass — and the unweighted test survives, as claimed. So the matrix is one test stronger than stated, in the right direction.

The module-fixture caveat is confirmed real: `_underquoted_fit_result` is `scope="module"`, so a function-scoped patch runs after the fixture has already fit. Worth remembering if the guard ever moves.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Confirmed on both counts. The gap it closes is real: every other #690 test quotes errors above the floor, and the twin-fit helper cancels the transform term by construction (its docstring says so explicitly), so before this test nothing pinned the floor-engaged covariance in absolute terms. And it does guard the live path — under a forced `scale_covar=True` mutation this test fails along with the other weighted-covariance tests, so a regression that re-scaled or un-floored the covariance cannot slip past it.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: The exact-cancellation argument was re-checked from the helper's construction rather than trusted: both runs quote `mag_error` at or under `_MIN_FIT_SIGMA` (`_MIN_FIT_SIGMA` and `1e-12`), and `cat_error=None` keeps the catalog from lifting either total sigma off the floor, so `weights = 1 / max(sigma, min_fit_sigma)` is bit-identical between the twins, the fits share their covariance, and the transform term subtracts out exactly — which is what lets the measurement half be pinned at atol 1e-8 rather than approximately. The old perfect-fit pins genuinely were multiplying any transform-term bug by redchi ≈ 0, so this is a strengthening, not a compensating weakening: the transform half now carries its own harder pin (rtol 1e-6, correlations included) in the unscaled-covariance test.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Confirmed strictly stronger: `_assert_rows_blind_to_the_spoiled_ones` iterates every column in `_TRANSFORM_OUTPUT_COLUMNS` — per-star values and per-image diagnostics alike — with `assert_array_equal`, i.e. bit-for-bit, against a reference run whose input never contained the spoiled rows. The old check compared surviving rows to a constant at atol 1e-8 and could only see contamination that moved `mag_cal_error`; this one catches a spoiled row leaking into *any* output, including diagnostics-only columns like `fit_max_weight_share`, and exact equality means nothing can cancel. Both consuming tests (`bad_error` and `no_magnitude` cases) pass in the current subset run.

CHANGES.rst
mwcraig: The writer-side claim was re-verified: `io/aavso.py` has no magnitude-error guard — `_format_magerr` only maps non-finite values to `"na"` and otherwise formats whatever it is given — so smaller `mag_cal_error` values flow straight into MAGERR, as this comment says.

Two things frame the decision this comment leaves open:

1. The direction of the change is toward *smaller* reported errors only for images whose `fit_redchi` was already > 1 — i.e. images where #694 established the old values were understated 3–5× anyway. Neither the old nor new number is right for those images; the new one is at least the honestly-derived half of the story, with `fit_excess_scatter` reporting the other half explicitly.
2. Whether the #694 follow-up (propagating `fit_excess_scatter` into `mag_cal_error`) lands before the next release is @mwcraig's call, not something this PR should decide implicitly. Flagging it here so the release decision is made consciously; happy to start that follow-up on a branch once this merges.

stellarphot/utils/magnitude_transforms.py
mwcraig, line 1063: THis docstring entry has gotten waayyyyy toooooo loooooong. THe content matters, so in addition to a *brief* description here and a section in the notes that discusses the errors in more detail (but still try rto condense)

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1118,18 +1125,24 @@ def transform_to_catalog(
...
-    observed about the fit rather than the instrumental errors that were
-    quoted, which is the more robust of the two: it stays right when the
-    quoted errors are systematically wrong. The price is that ``fit_redchi``
-    and the ``*_error`` columns are not independent diagnostics. An image
-    whose stars scatter twice as far from the model as they claim reports both
-    a ``fit_redchi`` near four and coefficient uncertainties twice as large,
-    and that is one observation rather than two agreeing ones.
+    The reported uncertainties believe the errors the fit was weighted by --
+    the quoted errors, floored at ``min_fit_sigma`` -- rather than the
+    scatter actually observed about the fit: `lmfit`'s ``scale_covar``
```
mwcraig: This seems toolong/historical

stellarphot/utils/magnitude_transforms.py
mwcraig: Is this essentially a repetition of what is above?

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1229,34 +1229,325 @@ def test_transform_to_catalog_uncertainty_falls_as_stars_are_added(mocker):
...
+    color = np.asarray(result["color_cat"])
+
+    return np.column_stack([mag_inst, color, np.ones_like(mag_inst)])
+
+
+def _predicted_coefficient_covariance(result, sigma_quoted):
+    """
+    The covariance a fit of the default terms should report, by hand.
+
+    The transform model is linear in its coefficients, so the covariance of
```
mwcraig: Move to notes

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1229,34 +1229,325 @@ def test_transform_to_catalog_uncertainty_falls_as_stars_are_added(mocker):
...
+    # than becoming an accident. Under the old scale_covar default the floor
+    # cancelled out of the rescaled covariance and could not reach the
+    # reported errors at all. See issue #690.
+    claimed = 0.002
+
+    result, _, _ = _fit_a_catalog(
+        mocker,
+        n_stars=40,
+        sigma=claimed,
+        seed=20260814,
```
mwcraig: Why are there multiple seeds in the tests in here? Is there any reason not to have just a single seed defined as a module-level constant?

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — it grew a clause per issue (#694, then #690) and reads like it. Proposed brief version, with the detail folded into the condensed Notes errors section proposed in the thread at line 1130:

```
min_fit_sigma : float, optional
    Floor, in magnitudes, on the per-star uncertainty the fit weights
    by; pass ``0`` to apply no floor. The default of 0.01 is about the
    smallest error credible for a single ground-based measurement tied
    to a survey catalog; see issue #694. Must not be negative, and has
    no effect on an unweighted fit. See Notes for how the floor reaches
    the reported uncertainties.
```

The overstatement trade-off and the "when to pass 0" guidance move to Notes rather than being cut — they are the part of the content that matters.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed — it explains the old `scale_covar` behavior twice ("which is what the rescaling did…", "where the rescaling reported…"), and the history is already told once in CHANGES.rst, which is where it belongs. Proposed present-tense replacement, which also absorbs the paragraph at line 1167 (see that thread — it is partly a repetition of this one):

```
The reported uncertainties believe the errors the fit was weighted by --
the quoted errors, floored at ``min_fit_sigma`` -- not the scatter
observed about the fit. That keeps ``fit_redchi`` and the ``*_error``
columns independent: stars scattering beyond their quoted errors leave
the reported uncertainties alone and instead raise ``fit_redchi``, with
``fit_excess_scatter`` giving the size of what the quoted errors
missed. When the quoted errors are wrong, the ``*_error`` columns are
wrong with them, and ``fit_redchi`` far from one is the alarm that says
so -- ``mag_cal_error`` is then off by roughly its square root. The one
exception is quoted errors below the floor: there the reported
uncertainties take the floor's scale, overstated by about
floor/quoted, and that rule of thumb overcorrects; pass
``min_fit_sigma=0`` when errors that small are genuine. An unweighted
fit quotes no errors to believe, so its uncertainties are scaled to the
observed scatter, the only scale it has.
```

That is ~18 lines replacing the ~40 across the two paragraphs, with nothing load-bearing dropped: the convention, the alarm, the rule of thumb, the floor caveat, and the escape are all still there.

stellarphot/utils/magnitude_transforms.py
mwcraig: Partly, yes. "Believe the errors the fit was weighted by" is restated from the paragraph above; what is *not* repeated is (a) which half of ``mag_cal_error`` uses the raw quoted error vs. the floored covariance, and (b) the √``fit_redchi`` rule of thumb with its sub-floor caveat. The condensed rewrite proposed in the thread at line 1130 merges the two paragraphs into one, keeping (a) and (b) and dropping the restatement — this paragraph disappears as a separate unit.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Will do — the extended summary becomes a one-line description plus a Notes section holding the ``(J^T W J)^-1`` explanation and the "the fit's own residuals must not rescale it" statement of intent. Same treatment for `_predicted_mag_cal_error` next to it, which has the same shape.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: No reason beyond each test picking one when it was written (several are just dates). Checked empirically rather than argued: a copy of this file with every fixed seed replaced by a single value (13579) passes all 117 tests in 2.9 s, so nothing hidden depends on a particular realization — the loose-tolerance assertions (the `rel=0.3` redchi checks and similar) all hold at the shared seed.

So a single module-level constant works, with two notes for the implementation:

- The Monte Carlo comparison uses ``seed=10000 + trial`` deliberately — that one must stay per-trial, presumably ``SEED + trial``.
- The one theoretical argument for distinct seeds is that the suite samples more noise realizations overall, but the assertions here are either exact (rtol 1e-6 against hand-computed predictions, where the realization cancels) or deliberately loose, so that value is negligible next to having one seed to reason about. If any future seed change is needed, a constant also means one rerun checks everything.

Docstrings that cite specific seed values (`_noisy_fit_result`, `_underquoted_fit_result`) get updated in the same pass.

stellarphot/utils/magnitude_transforms.py
mwcraig: Open that issue

CHANGES.rst
mwcraig: Make anissue to remind us of this or provide a link to an existing issue in this thread.

stellarphot/utils/magnitude_transforms.py
mwcraig: Filed as #697. (This reply was written by Claude at @mwcraig's direction.)

CHANGES.rst
mwcraig: Filed as #698. (This reply was written by Claude at @mwcraig's direction.)

stellarphot/utils/magnitude_transforms.py
```diff
@@ -234,7 +242,9 @@ def _to_float_array(values):
     `numpy.ndarray`
         Float array with NaN wherever the input was masked.
     """
-    return np.ma.filled(np.ma.asarray(values, dtype=float), np.nan)
+    return np.asarray(
+        np.ma.filled(np.ma.asarray(values, dtype=float), np.nan), dtype=float
+    )
```
mwcraig: **Silent unit misread**: `np.ma.asarray(values, dtype=float)` strips a `Quantity`'s unit without converting, so an error column in mmag enters the fit as if it were mag — weights off by 1e6 in variance and `mag_cal_error` inflated 1000x, all silent. Reproduced: `_to_float_array(10 * u.mmag)` returns `10.0`, not `0.01`. Before this PR the same input crashed loudly with `UnitConversionError` in `np.hypot`. Converting when the unit is convertible to mag (or warning when it isn't dimensionless) keeps the masked-value fix without silencing real unit mistakes.

*Comment written by Claude Code at Matt's direction.*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -790,15 +796,15 @@ def _unweighted_fit_result(module_mocker):
     `test_transform_to_catalog_reports_unweighted_fit_statistic` and
     `test_transform_to_catalog_diagnostics_for_an_unweighted_fit` both read
     -- and neither mutates -- the result of the same fit with no error
-    column: 100 stars, noise sigma 0.02, seed 13579. Computing it once here
-    rather than twice keeps the tests from drifting apart on inputs they
-    mean to share.
+    column: 100 stars, noise sigma 0.02, the shared seed. Computing it once
```
mwcraig: **Test-setup duplication**: this fixture re-implements `_fit_a_catalog`'s body inline only because `_fit_a_catalog` can't forward `obs_error_column=None`, and its `n_stars=100` / `sigma=0.02` literals are re-declared in the dependent tests with "must match" comments. A `**fit_kwargs` pass-through on `_fit_a_catalog` (which would also absorb the ad-hoc `min_fit_sigma` plumbing) makes this fixture a one-liner, and module constants — the pattern this PR already established with `_UNDERQUOTED_*` — keep the literals in one place.

*Comment written by Claude Code at Matt's direction.*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1229,34 +1281,329 @@ def test_transform_to_catalog_uncertainty_falls_as_stars_are_added(mocker):
...
+    # errors and that alarm were one observation dressed as two.
+    result = _underquoted_fit_result
+
+    sigma_quoted = np.hypot(_UNDERQUOTED_CLAIMED_ERROR, _FAKE_CATALOG_ERROR)
+    covariance = _predicted_coefficient_covariance(result, sigma_quoted)
+
+    for index, term in enumerate(("a", "c", "z")):
+        assert result[f"{term}_error"][0] == pytest.approx(
+            np.sqrt(covariance[index, index]), rel=1e-6
+        ), term
```
mwcraig: **Pasted three times**: this coefficient-error-vs-covariance loop appears verbatim in three of the new tests (~L1465, ~L1529, ~L1603). Each copy independently encodes the `('a', 'c', 'z')` ordering contract with `_design_matrix` and the `rel=1e-6` tolerance. A small `_assert_coefficient_errors_match(result, covariance)` helper next to `_design_matrix` states that contract once.

*Comment written by Claude Code at Matt's direction.*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1415,7 +1762,7 @@ def test_transform_to_catalog_respects_a_custom_min_fit_sigma(mocker):
         mocker=mocker,
         n_stars=n_stars,
         sigma=0.02,
-        seed=24680,
+        seed=_SEED,
```
mwcraig: **Near-verbatim copy**: `test_transform_to_catalog_respects_a_custom_min_fit_sigma` repeats `test_transform_to_catalog_floors_the_fit_sigma` with different literals (and this copy already dropped the `cat_error` parametrization the original has). Parametrizing the original over the floor value (default vs 0.05), with the error arrays derived from it, does both jobs in one body — today a change to the flooring contract must be edited in two places that can silently diverge.

*Comment written by Claude Code at Matt's direction.*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1528,7 +1875,7 @@ def test_transform_to_catalog_reports_max_single_star_weight_share(mocker):
         mocker,
         n_stars=n_stars,
         sigma=0.02,
-        seed=2468,
+        seed=_SEED,
```
mwcraig, line 1970: **Duplicated expectation**: the expected max-weight-share computation (`statistical = 1/errors**2`; `expected = statistical.max()/statistical.sum()`) is hand-restated both here and in `test_transform_to_catalog_min_fit_sigma_zero_disables_the_floor`. An `_expected_max_weight_share(errors)` helper keeps the two from diverging — e.g. one later "fixed" to `1/sigma` instead of `1/sigma**2`, exactly the mistake the nearby comment warns about.

*Comment written by Claude Code at Matt's direction.*

mwcraig (commented):
**Critical review of this PR** (focus: critical bugs, redundancy, complexity). Five findings are inline on the diff; the five below anchor to lines this diff doesn't touch, so they're collected here. The top three bugs were reproduced empirically, not just eyeballed.

---

### 1. Should fix before merge — `_excess_scatter`'s brentq bracket can crash the whole transform

https://github.com/feder-observatory/stellarphot/blob/8d392a39048fd67e5e091b60901aa27b9d70912c/stellarphot/utils/magnitude_transforms.py#L443-L453

When an image's quoted sigmas are many orders of magnitude below the residual RMS (e.g. placeholder errors of `1e-12` mag — which pass the `errors > 0` gate and are exactly the under-quoted-errors input `min_fit_sigma` exists to survive), `f(upper) <= 0` holds mathematically only by a margin of ~`(sigma/upper)**2`, which falls below the sum's floating-point rounding error. Both bracket endpoints then evaluate positive and `brentq` raises an uncaught `ValueError('f(a) and f(b) must have different signs')`, aborting the entire `transform_to_catalog` call mid-loop instead of degrading gracefully. Reproduced in 66/200 seeded trials with `sigma=1e-12`.

Small fix: evaluate `f(upper)` first and return `upper` when it is non-negative (or widen the bracket with `np.nextafter`).

### 2. `min_fit_sigma` validation lets NaN and inf through

https://github.com/feder-observatory/stellarphot/blob/8d392a39048fd67e5e091b60901aa27b9d70912c/stellarphot/utils/magnitude_transforms.py#L1174

`nan < 0` is `False`, so `min_fit_sigma=float('nan')` (e.g. deserialized from a settings file) skips the `ValueError` and `np.maximum(sigma, nan)` turns every weight NaN — every image's fit fails, surfaced only as per-image "Fit did not succeed" warnings and all-NaN columns, with no pointer to the bad argument. `inf` zeroes all weights so the "fit" sits at the initial parameters. Suggested guard: `if not (min_fit_sigma >= 0) or not np.isfinite(min_fit_sigma):`.

### 3. Redundancy: `_reported_redchi` re-implements `_excess_scatter`'s bracket function

https://github.com/feder-observatory/stellarphot/blob/8d392a39048fd67e5e091b60901aa27b9d70912c/stellarphot/utils/magnitude_transforms.py#L456-L493

`_reported_redchi(fit, sigma, weights)` is exactly `reduced_chi_square_less_one(0.0) + 1` from `_excess_scatter`; both independently unweight `fit_result.residual / weights`, so the same quoted-sigma chi-square is computed up to three times per image in different operand orders. The two formulations can disagree by ULPs around 1.0 (a `fit_redchi` slightly above 1 alongside `fit_excess_scatter == 0.0`), which the code defends against with a long comment instead of computing the sum once. One shared quoted-chi-square helper (unweight once per image, gate the excess on `redchi_quoted - 1`) removes the duplication, the duplicated `None`-branch, and the ULP defense outright.

### 4. Follow-up: the AAVSO workflow consumes `mag_cal_error` under the old contract

`docs/stellarphot/user_guide/calibrating_photometry.rst` feeds `mag_cal_error` to MAGERR. With `scale_covar` off, submitted errors are now smaller than before whenever `fit_redchi > 1`; the semantic shift is documented in the docstring and CHANGES.rst, but not at this one real downstream consumer. Worth a sentence in the guide pointing at `fit_redchi`/`fit_excess_scatter`.

### 5. Follow-up: the transit fit still uses `scale_covar=True`

`TransitModelFit._run_fit` (`stellarphot/transit_fitting/core.py`) weights by `1/err` and leaves lmfit's default rescaling on, so transit depth/duration uncertainties are still silently rescaled by redchi — the behavior #690 calls misleading. The repo's two weighted fits now report uncertainties under contradictory conventions; neither #697 nor #698 covers this. Worth filing so the convention lands in one shared place.

stellarphot/utils/magnitude_transforms.py
mwcraig: Fixed in 253e4a3. `_to_float_array` now reads the unit off the input *before* the `numpy.ma` round-trip: a unit convertible to mag is converted (mmag comes back divided by 1000), dimensionless or no unit passes through unchanged, and anything else raises `UnitConversionError` — the same loud failure `np.hypot` used to produce before this function handled units at all. Mask→NaN behavior is preserved through the conversion.

New tests cover mmag conversion (Quantity and `MaskedColumn`), mag pass-through, the incompatible-unit raise, and unitless input; each was confirmed failing against the unfixed code first. `test_to_float_array_strips_units`, whose premise this intentionally changes, was replaced, and the tolerates-units integration test now uses `u.mag` since its old `u.one / u.adu` unit is exactly what the fix rejects.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Fixed in 253e4a3. `_fit_a_catalog` now takes a single `**fit_kwargs`, split by a `_CATALOG_COEFFICIENT_NAMES` frozenset: coefficient/catalog-shape names go to `_generate_fake_catalog`, everything else (`min_fit_sigma`, `obs_error_column`, …) falls through to `transform_to_catalog` — which also absorbed the ad-hoc `min_fit_sigma` plumbing. The `_unweighted_fit_result` fixture is now a thin `_fit_a_catalog(..., obs_error_column=None)` call, and module constants `_UNWEIGHTED_N_STARS` / `_UNWEIGHTED_SIGMA` replaced the re-declared literals and their "must match" comments.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Fixed in 253e4a3. The loop is now `_assert_coefficient_errors_match(result, covariance)`, placed right after `_design_matrix` with the `('a', 'c', 'z')` ordering contract stated once in its docstring; all three copies were replaced with calls.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Fixed in 253e4a3. `test_transform_to_catalog_floors_the_fit_sigma` is now parametrized over the floor (`None` for the default, `0.05` for a caller-chosen one) on top of the existing `cat_error` parametrization, with the error arrays derived from the floor value; the near-copy was deleted. The merge nets four cases where the two tests had three, so the custom floor now also runs the combined-error path the copy had dropped.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Fixed in 253e4a3. Both tests now call a shared `_expected_max_weight_share(errors)` whose docstring pins that a fit's weights go as `1/sigma**2` — the exact divergence the old comments warned about.

mwcraig:
Responses to the five findings in the review body (they anchor to lines outside the diff, so no thread to reply in). All code changes are in 253e4a3.

**1. brentq bracket crash** — Fixed. `_excess_scatter` evaluates the bracket function at `upper` first and returns `upper` when the value rounds to the positive side; mathematically it is at most zero there, so a positive value only means `upper` already is the root to within rounding. The regression test (`sigma=1e-12`, order-unity residuals, seed 2) reproduced the exact `ValueError` before the fix.

**2. NaN/inf `min_fit_sigma`** — Fixed. The guard is now `not np.isfinite(min_fit_sigma) or min_fit_sigma < 0`, and the validation test is parametrized over `[-0.01, nan, inf]`. Confirmed beforehand that `nan` surfaced only as an unrelated lmfit error and `inf` didn't raise at all.

**3. `_reported_redchi` redundancy** — Fixed, with one wrinkle beyond the sketch in the review. `_reported_redchi` is now `_quoted_redchi`, computed once per image in `transform_to_catalog` and passed through `_fit_diagnostics` into `_excess_scatter`, whose gate is `redchi_quoted - 1.0 <= 0.0`; the duplicated `None` branch and the second sum are gone. The wrinkle: the gate value had to adopt the bracket function's operand order (`residual**2 / sigma**2`, not `(residual / sigma) ** 2`), otherwise a one-ULP disagreement around 1.0 could still pass the gate while handing brentq two negative endpoints — the same crash the old comment defended against, reintroduced. A comment on `_quoted_redchi` pins that constraint.

**4. AAVSO guide** — Done. `calibrating_photometry.rst` now says `mag_cal_error` is computed from the uncertainties as quoted and no longer rescaled by the fit's goodness, and points at `fit_redchi` / `fit_excess_scatter` for judging whether the quoted errors account for the observed scatter.

**5. Transit fit `scale_covar`** — Filed as #699.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -339,6 +368,13 @@ def _fit_diagnostics(fit_result, sigma, weights, cat_error_usable):
         own ``weights``, not something re-derived from ``sigma``. The scalar
         ``1.0`` for an unweighted fit.
 
+    redchi_quoted : float
+        ``fit_redchi`` for this fit, i.e. `_quoted_redchi` already called on
```
mwcraig: COndense and/or rephrase -- seems like documentation for _excess_scatter hiding in the docstring for a parameter

stellarphot/utils/magnitude_transforms.py
```diff
@@ -371,11 +407,64 @@ def _fit_diagnostics(fit_result, sigma, weights, cat_error_usable):
...
+        ),
     }
 
 
-def _excess_scatter(fit_result, sigma, weights):
+def _quoted_redchi(fit_result, sigma, weights):
+    """
+    Reduced chi-square measured against the uncertainties as quoted.
+
+    `lmfit`'s ``redchi`` is measured against the sigmas the fit actually
```
mwcraig: For the love of all that is holy, are you incapable of remembering that long descriptions in a numpy docstring always belong in the nnotes?

stellarphot/utils/magnitude_transforms.py
```diff
@@ -412,6 +501,15 @@ def _excess_scatter(fit_result, sigma, weights):
         Weight the fit gave each residual, floor included. Unused when
         ``sigma`` is `None`.
 
+    redchi_quoted : float, optional
+        ``_quoted_redchi(fit_result, sigma, weights)`` for this fit, if the
```
mwcraig, line 504: Similar comment to the one above documentation for a function in a parameter description

Issue #699, opened 2026-08-20
Transit fit still rescales uncertainties by redchi (scale_covar default) — inconsistent with #696


mwcraig:
#696 turned off lmfit's default `scale_covar` rescaling in `transform_to_catalog`, so the transform fit now reports coefficient and magnitude uncertainties measured against the errors as quoted, with `fit_redchi` / `fit_excess_scatter` carrying the goodness-of-fit information separately (#690).

The transit fit still uses the old convention. `TransitModelFit._run_fit` (`stellarphot/transit_fitting/core.py`) weights the residual by `1 / flux_errors` and calls `lmfit.minimize` without `scale_covar=False`, so lmfit's default rescaling is on: transit depth/duration/center uncertainties are silently multiplied by the square root of the reduced chi-square — the behavior #690 called misleading for the magnitude transform.

The repo's two weighted fits now report uncertainties under contradictory conventions. Whichever convention is right, it should be one shared decision:

- switch the transit fit to `scale_covar=False` and expose the reduced chi-square (and possibly an excess-scatter estimate) the way `transform_to_catalog` now does, or
- document explicitly that the transit fit intentionally keeps rescaled uncertainties and why.

Neither #697 nor #698 covers this.

*This issue was written by Claude Code at Matt's direction.*

PR #701, opened 2026-08-22
Fold fit_excess_scatter into mag_cal_error

### LINKED ISSUE #698: Propagate fit_excess_scatter into mag_cal_error
Issue #694 showed that `mag_cal_error` understates true scatter by 3-5x on real data. PR #696 makes that excess explicit via the new `fit_excess_scatter` diagnostic column, but does not fold that excess into `mag_cal_error` itself — the reported per-star calibrated-magnitude error still does not reflect the full observed scatter.

This predates #696, too: the old `scale_covar` rescaling only ever recovered about `excess * sqrt(p/N)` of the excess scatter (roughly 1/12 of it on the #694 data, where p is the number of fit parameters and N the number of stars), so even before #696 most of the excess was already missing from `mag_cal_error`. #696 just stopped hiding that gap behind a rescaled covariance and started reporting it as a separate number instead of folding it in at all.

### Why this matters for release

`stellarphot/io/aavso.py` formats `mag_cal_error` straight into the AAVSO report's MAGERR field, with no guard against understated errors. Whether to propagate `fit_excess_scatter` into `mag_cal_error` before the next release — as opposed to shipping the diagnostic-only version from #696 — is therefore a conscious release decision, not just a nice-to-have follow-up.

### Not a double-count against #691

The #691 acceptance test result confirms that catalog color errors are **not** what `fit_excess_scatter` is made of, so folding `fit_excess_scatter` into `mag_cal_error` here would not double-count against the catalog color-uncertainty term #691 is tracking.

### References

- #690
- #694
- #696

---
This issue was written by Claude at @mwcraig's direction.

mwcraig:
Closes #698 — the user-visible half of #690/#696.

`mag_cal_error` from `transform_to_catalog()` now includes the image's `fit_excess_scatter` in quadrature, so the number `write_aavso_extended` copies into `MAGERR` reflects the scatter actually observed about the transform rather than only the quoted errors (which understated it several-fold on the #694 data). The previous value is recoverable as `sqrt(mag_cal_error**2 - fit_excess_scatter**2)`.

Design points (see the Notes sections):
- Added **after** propagation through the transform, not to the star's measurement error before it: the excess is measured in calibrated-magnitude residual space, so routing it through `(1 + a)` would apply that sensitivity twice.
- The coefficient covariance stays unscaled — rescaling it would be the `scale_covar` behaviour #690 removed. The transform term remains `~sqrt(fit_redchi)` understated when quoted errors are wrong, but it is the `O(1/N)` part of the total.
- No opt-out keyword; the honest number is the default. A zero excess leaves the column bit-identical to before.
- The `_fit_diagnostics` block now runs before `_calibrated_with_uncertainty` and the same float feeds both the column and the error term.

Tests: error ≈ observed noise on the under-quoted fixture (the #694 closure); unchanged when errors describe the data; after-vs-through-the-transform distinguished at `a = 0.3` (at 0.5 the pre-fit 1-mag cut starts dropping stars and breaks the hand-computed covariance); covariance-shape tests strip the excess first via a shared helper.

This PR was written by Claude Code at Matt's direction.

https://claude.ai/code/session_015RURMHNvNLcGuqnx4u7D4V


stellarphot/utils/magnitude_transforms.py
mwcraig, line 470: The extended description belongs in notes,

mwcraig, line 1127: This is an extraordinary long note, especially for a private function. If it matters land I suspect it does) then it should be in decimation somewhere.

mwcraig (commented):
Critical review focused on complexity and redundancy; no correctness problems found. Roughly ~110 lines of the net +330 can go without losing coverage or information. Biggest wins: merge the duplicate underquoted test, delete the now-subset zero-excess test and inline its fixture, and collapse the six restatements of the excess→`mag_cal_error` fact down to two. Details inline. Kept as-is on purpose: `_error_without_the_excess_scatter` (`_noisy_fit_result`'s excess is genuinely nonzero, ~0.009), `..._adds_the_excess_after_the_transform`, and the `base_error` change. Notebook diff is clean.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -474,9 +474,11 @@ def _excess_scatter(fit_result, sigma, weights, redchi_quoted=None):
     own photometry, neither of which any weighting scheme can fix; see
     issue #694.
 
-    Reported rather than folded into the weights. A fit that absorbs its own
-    excess scatter has a reduced chi-square of one by construction, which
-    destroys the one diagnostic that revealed any of this.
+    Reported rather than folded into the weights, and carried into
```
mwcraig: The "excess feeds `mag_cal_error`" fact is now stated six times in this file. Suggest two homes: the `mag_cal_error` paragraph of `transform_to_catalog`'s docstring (public semantics) and `_calibrated_with_uncertainty`'s Parameters/Notes (the after-not-through rationale). This clause is one of the four drive-bys to cut — a helper doesn't need to list its callers.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1193,37 +1214,47 @@ def transform_to_catalog(
...
+    errors leave those columns alone and instead raise ``fit_redchi``, with
     ``fit_excess_scatter`` giving the size of what the quoted errors
-    missed. When the quoted errors are wrong, the ``*_error`` columns are
-    wrong with them, and ``fit_redchi`` far from one is the alarm that says
-    so -- ``mag_cal_error`` is then off by roughly its square root. The one
-    exception is quoted errors below the floor: there the reported
-    uncertainties take the floor's scale, overstated by about
-    floor/quoted, and that rule of thumb overcorrects; pass
-    ``min_fit_sigma=0`` when errors that small are genuine. An unweighted
+    missed; only ``mag_cal_error``, below, picks that excess up. When the
```
mwcraig: Drive-by restatement; the `mag_cal_error` paragraph two below already says this in full. Suggest dropping "only ``mag_cal_error``, below, picks that excess up".

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1193,37 +1214,47 @@ def transform_to_catalog(
...
-    rather than falling back to the measurement error alone, for an image
-    whose fit left no usable covariance behind.
+    included, and then adds the image's ``fit_excess_scatter`` in
+    quadrature. The transform term is a significant contribution -- the
+    part a plain measurement-error column omits entirely -- that grows as
+    the number of fitted stars shrinks, and it is worked out per star
+    because a fit predicts best at the centroid of the stars it was fit to.
+    The excess term is what makes the column reflect the scatter actually
+    observed about the transform rather than only the quoted errors, which
+    on real data understated it by several times; the value without it is
```
mwcraig: "several times" also appears in the CHANGES entry, and the test comment says "five times". Keep the quantifier in one place — here is fine — and drop it from the changelog.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1193,37 +1214,47 @@ def transform_to_catalog(
...
+    falling back to the measurement error alone, for an image whose fit
+    left no usable covariance behind.
 
     A star's catalog entry decides whether it was matched. When a color term
     (``c`` or ``d``) is fit, though, the model also applies that star's
     catalog color, so the color's uncertainty is a real, knowingly omitted
     contribution -- roughly ``c * sigma_color`` -- tracked as issue #691. What
     the catalog reliably contributes is its field-wide scatter about the
-    transform, which lands in ``fit_excess_scatter``; and its systematic tie
+    transform, which lands in ``fit_excess_scatter`` and, through it, in
```
mwcraig: Same restatement again ("and, through it, in ``mag_cal_error``"). This paragraph is about the catalog's systematic tie; suggest cutting the aside.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1593,7 +1624,8 @@ def transform_to_catalog(
         # The covariance is reported as the fit was weighted rather than
         # rescaled by the reduced chi-square: the quoted errors are believed,
         # and their disagreement with the observed scatter is reported once,
-        # in ``fit_redchi`` and ``fit_excess_scatter``, instead of also being
+        # in ``fit_redchi`` and ``fit_excess_scatter`` -- and carried into
```
mwcraig: This comment is about why the coefficient covariance stays unscaled; splicing the `mag_cal_error` fact into the middle of it is scope creep (and the fourth restatement). Suggest reverting this line to its pre-PR form.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1667,11 +1699,40 @@ def transform_to_catalog(
...
+        # Computed once and shared with `_fit_diagnostics` below, so the
+        # value reported here as ``fit_redchi`` is the exact float that
+        # gates ``fit_excess_scatter`` too, rather than two independent sums
+        # of the same quantity that could round differently.
+        redchi_quoted = _quoted_redchi(fit_result, sigma, weights)
+        fit_redchis[rows] = redchi_quoted
+
+        # How the fit was weighted, rather than where it landed. Written
+        # alongside ``fit_redchi`` because they are the numbers that say
+        # whether that one can be taken at face value. Computed before the
```
mwcraig: The added second half re-spends the "one float, two consumers, could disagree" argument already made three lines above for `redchi_quoted` and in `_excess_scatter`'s Notes — and here there's no real risk to explain, it's one dict value read twice. Suggest trimming to something like:

```
# Before the calibrated errors, which need the excess scatter too.
```

stellarphot/utils/magnitude_transforms.py
```diff
@@ -740,7 +756,12 @@ def _calibrated_with_uncertainty(fit_result, covar, mag_inst, color, errors, fit
         if fit_diff:
             calibrated = calibrated + star_mag
 
-        return unumpy.nominal_values(calibrated), unumpy.std_devs(calibrated)
+        uncertainty = unumpy.std_devs(calibrated)
+
+    if np.isfinite(excess_scatter) and excess_scatter > 0:
```
mwcraig: The `np.isfinite` half is dead: `_excess_scatter` returns NaN only when `sigma is None`, i.e. an unweighted fit, which is exactly when `star_errors is None` and this function already returned at line 734. And `nan > 0` is `False` anyway. Suggest:

```python
if excess_scatter > 0:
```

(or just unconditional `np.hypot`, since `hypot(x, 0) == x`). Either way the "Zero, NaN and negative values add nothing" sentence at line 701 documents cases the call site makes impossible and can go.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1565,20 +1603,115 @@ def test_transform_to_catalog_error_uses_the_unscaled_covariance(
...
+            result,
+            covariance,
+            _UNDERQUOTED_CLAIMED_ERROR,
+            result["fit_excess_scatter"][0],
+        ),
         rtol=1e-6,
     )
 
 
+def test_transform_to_catalog_error_includes_the_excess_scatter(
```
mwcraig: Near-verbatim duplicate of `test_transform_to_catalog_error_uses_the_unscaled_covariance` just above: same fixture, same `sigma_quoted`/covariance recomputation, identical `rtol=1e-6` assertion. The only unique lines are `assert excess > 0` and the closure check against `_UNDERQUOTED_NOISE_SIGMA`. Suggest moving those two lines into the test above and deleting this one (~25 lines). Either way, `_underquoted_fit_result`'s docstring says "both" and names two tests but now has three readers — it needs updating.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2010,16 +2152,12 @@ def test_transform_to_catalog_reports_excess_scatter(mocker):
 
 
 def test_transform_to_catalog_reports_no_excess_scatter_when_errors_describe_the_data(
```
mwcraig: Now a one-line subset of `test_transform_to_catalog_error_unchanged_when_errors_describe_the_data`, which makes this same `== 0.0` assertion as its precondition. Suggest deleting it. That leaves `_generous_errors_fit_result` (plus its two constants and docstring) with a single consumer, so it should be inlined back into that one test as a plain `_fit_a_catalog(...)` call — the file's own convention for single-use fits (~30 lines total). The counter-argument is that this test is the symmetric pair of `..._reports_excess_scatter`; I'd still cut since the assertion is textually identical.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1565,20 +1603,115 @@ def test_transform_to_catalog_error_uses_the_unscaled_covariance(
...
+        result, covariance, _UNDERQUOTED_CLAIMED_ERROR, excess
+    )
+    # The wrong answer: the excess folded into the quoted error and then
+    # scaled by 1 + a along with it.
+    through_the_transform = _predicted_mag_cal_error(
+        result, covariance, np.hypot(_UNDERQUOTED_CLAIMED_ERROR, excess)
+    )
+
+    np.testing.assert_allclose(reported, after_the_transform, rtol=1e-6)
+    assert not np.allclose(reported, through_the_transform, rtol=0.1)
```
mwcraig: Optional: this negative check is implied by the `rtol=1e-6` positive match on the line above — the comment itself says the two candidates differ by ~30%. The test as a whole is worth keeping (it's the only one that distinguishes after-vs-through).

CHANGES.rst
```diff
@@ -138,6 +138,11 @@ Other Changes and Additions
   come out smaller than before, and ``fit_excess_scatter`` reports
   explicitly what the rescaling used to absorb silently. Unweighted fits
   still scale by the observed scatter, the only scale they have. [#690]
++ ``mag_cal_error`` from ``transform_to_catalog()`` now includes the image's
+  ``fit_excess_scatter`` in quadrature, so it reflects the scatter actually
+  observed about the transform rather than only the quoted errors, which
+  understated it by several times on real data. The previous value is
+  ``sqrt(mag_cal_error**2 - fit_excess_scatter**2)``. [#698]
```
mwcraig: Suggest trimming to three lines — the recovery formula is implementation detail already in the docstring, and "several times" is unsourced here:

```rst
+ ``mag_cal_error`` from ``transform_to_catalog()`` now adds the image's
  ``fit_excess_scatter`` in quadrature, so it reflects the scatter actually
  observed about the transform rather than only the quoted errors. [#698]
```

docs/stellarphot/user_guide/calibrating_photometry.rst
```diff
@@ -48,13 +48,14 @@ The ``mag_column`` and ``mag_error_column`` arguments name the calibrated
...
-``mag_cal_error`` is computed from the input uncertainties as quoted and is
-no longer rescaled by the goodness of the transform fit, so it is smaller
-than in earlier releases whenever the fit's scatter exceeds the quoted
-errors. Before submitting, check the ``fit_redchi`` and
-``fit_excess_scatter`` columns produced by
+``mag_cal_error`` combines the input uncertainties as quoted, the
+uncertainty of the transform fit and, in quadrature, the image's
+``fit_excess_scatter`` -- the scatter observed about the transform beyond
+what the quoted errors account for -- so it reflects the scatter actually
+seen in the data. Before submitting, it is still worth checking the
```
mwcraig: "it is still worth checking" is a hedge the old "check" didn't have; suggest reverting to "Before submitting, check the ...". The sentence above also inlines a full definition of `fit_excess_scatter` ("-- the scatter observed about the transform beyond what the quoted errors account for --") inside an already long sentence; the column name plus the docstring link is enough here.

stellarphot/utils/magnitude_transforms.py
mwcraig: Partly agree. The docstring already has a Notes section (it holds the `redchi_quoted` remark), and the three paragraphs before Parameters are pre-existing on `main` — this PR only touches the second one (the "carried into `mag_cal_error`" clause that the thread below asks to cut). Happy to move the second and third paragraphs ("Reported rather than folded into the weights…" and "The best-fit residuals are held fixed…") down into Notes while I'm in here, leaving just the first paragraph as the extended summary — but strictly it's a drive-by on pre-PR text, so say if you'd rather keep this PR's diff to its own changes.

mwcraig: Small correction to the premise: line 1129 is the `Notes` heading of `transform_to_catalog`, which is public, not a private function. The section runs 1129–1262 (~130 lines); this PR's net change to it is about +10 lines (the `mag_cal_error` paragraph rewrite plus the two asides flagged in the threads below). The bulk is pre-existing weighting/error-model narrative from #695/#696.

Agreed on the underlying point, though: the weighting model, what `fit_redchi`/`fit_excess_scatter` mean, when to pass `min_fit_sigma=0`, the `sqrt(fit_redchi)` rule of thumb, and the catalog-tie discussion are user-guide material, and `calibrating_photometry.rst` currently only gestures at them. I'd suggest a separate PR that moves that narrative into the user guide and leaves the Notes with a short summary plus a `:ref:` link, rather than growing this PR's scope. For this PR I'll trim the restatements the other threads identify. The one genuinely private long Notes this PR adds is the 8-line after-vs-through paragraph in `_calibrated_with_uncertainty`; that rationale isn't stated anywhere public, so I'd keep it there (it's the only place that explains why the test `..._adds_the_excess_after_the_transform` exists).

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed, and six is an undercount — counting the `excess_scatter` parameter entry and the two inline code comments it's nine places in this file, all added by this PR. Will cut this clause (a helper needn't list its callers), the asides at 1223 and 1256, and the splice in the covariance comment at 1627, leaving the `mag_cal_error` paragraph in `transform_to_catalog` (public semantics) and `_calibrated_with_uncertainty`'s Parameters/Notes (after-not-through rationale).

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed; will drop "only `mag_cal_error`, below, picks that excess up". One nit on the pointer: the `mag_cal_error` paragraph is the immediately following one, not two below.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed on keeping one quantifier; will drop it from the changelog. Note the tests say "five times" in four places (the `_UNDERQUOTED_*` constants comment, `_underquoted_fit_result`'s docstring, and two test comments — three of those pre-date this PR). That's a different number because it's the synthetic fixture's ratio (`0.1 / 0.02`), not the real-data observation, so I'd leave those alone; they describe the fixture, not the motivation.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed; will revert to "which lands in `fit_excess_scatter`; and its systematic tie".

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed; will revert to the pre-PR wording. The splice also reads oddly as written: "reported once, in `fit_redchi` and `fit_excess_scatter` -- and carried into `mag_cal_error` per star from there --" contradicts itself ("once" followed immediately by a second place).

stellarphot/utils/magnitude_transforms.py
mwcraig: Mostly agreed. Minor point: it isn't quite the same argument re-spent — the three lines above and `_excess_scatter`'s Notes are about `redchi_quoted`, this one is about the excess value. But the conclusion holds: `diagnostics[_FIT_EXCESS_SCATTER_COLUMN]` is a dict entry read twice, so there's no disagreement to prevent. Will trim the second half to your one-liner, keeping the first two sentences ("How the fit was weighted… face value.") since they're the pre-existing description of what the diagnostics are.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed that `np.isfinite` is redundant, but not for quite the stated reason, and I'd pick `if excess_scatter > 0:` rather than unconditional `hypot`.

`_excess_scatter` has a second NaN return: `fit_result.nfree <= 0` (line 524), which is a *weighted* fit, so `star_errors` is not `None` there. Whether that reaches this function with `covar is not None` depends on what lmfit hands back for a zero-dof fit, and I'd rather not rely on it: `np.hypot(x, nan)` is NaN, so the unconditional form would silently blank every calibrated error in that image. `nan > 0` is `False`, so the `> 0` form handles both NaN routes and negatives in one comparison. Will make that change and cut the "Zero, NaN and negative values add nothing" sentence down to "Non-positive values add nothing."

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — diffed the two bodies and they share the fixture, `sigma_quoted`, covariance recomputation, and the `rtol=1e-6` assertion; only `assert excess > 0` and the `rtol=0.2` closure check against `_UNDERQUOTED_NOISE_SIGMA` are unique. Will fold those two into `..._uses_the_unscaled_covariance` and delete this test, and fix `_underquoted_fit_result`'s docstring, which currently names two readers (it has three on this branch, two after the merge).

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed. Its only assertion is verbatim the precondition at the top of `..._error_unchanged_when_errors_describe_the_data`, and that leaves `_generous_errors_fit_result` single-use. Direct `_fit_a_catalog(mocker, ...)` calls are the file's convention for one-off fits (ten or so examples), and the pre-PR version of this very test did exactly that. Will delete the test and inline the fixture plus its two `_GENEROUS_*` constants; roughly 35 lines net.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Half agree. For the current constants the negative check is implied: the comment says the candidates differ by "nearly a third of the excess", and working it through (`(1+a)^2 - 1 ≈ 0.69` on `e ≈ 0.098`) puts them ~28% apart, so a 1e-6 match to one rules out the other.

What it adds is a power guard: if the fixture were ever retuned so the excess became small relative to the quoted error, the two candidates would converge, the positive match would pass vacuously, and this line would be the only thing that failed. `assert excess > 0` doesn't cover that (a tiny positive excess passes). So I'd keep it, with a one-line comment saying that's its job — or drop it if you'd rather not carry the extra line; either way it's no longer a silent redundancy.

CHANGES.rst
mwcraig: Agreed; will use your three-line version. The recovery formula and "several times" are both already in the `mag_cal_error` paragraph of `transform_to_catalog`'s Notes, so nothing is lost.

docs/stellarphot/user_guide/calibrating_photometry.rst
mwcraig: Agreed on both. Will restore "Before submitting, check the …" and drop the inline definition — the `:func:` link was already there pre-PR, so the column name suffices. Also going to drop the closing "how much of that error the quoted uncertainties failed to explain", which is a third paraphrase of the same definition in that paragraph.

stellarphot/utils/magnitude_transforms.py
mwcraig: More it

mwcraig: Just do it here.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Drop

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1108,126 +1127,48 @@ def transform_to_catalog(
...
-    transform, which lands in ``fit_excess_scatter``; and its systematic tie
-    to the standard system -- about 0.02 mag for APASS DR9 -- which is
-    identical for every star in every image, so a per-star column would
-    mislead, appearing to average down by the square root of the number of
-    stars.
+    included, and then adds the image's ``fit_excess_scatter`` in
+    quadrature. The excess term is what makes the column reflect the scatter
+    actually observed about the transform rather than only the quoted
+    errors, which on real data understated it by several times; the value
+    without it is ``sqrt(mag_cal_error**2 - fit_excess_scatter**2)``.
```
mwcraig: Substantive: `fit_excess_scatter` is measured on the *comparison* stars and, per the user guide, includes "the catalog's own photometry". A target star is not a catalog star, so comparison-star catalog noise reaches it only through the coefficients (already in the transform term, at ~`s/sqrt(N)`), yet this adds the full `s`. Flat-field-type excess *does* hit the target at O(1) — the code can't tell the two apart. Simulation (N=30, quoted obs error 0.01, hidden catalog scatter 0.03, z-only fit): true target error 0.0115, pre-PR 0.0102, post-PR 0.0313 (~2.7x high). So the column is a conservative upper bound when the excess is catalog scatter. No code change needed if erring high is the intent for MAGERR, but one sentence here saying so would make the claim accurate — and note the #674 CHANGES entry ("Catalog magnitude uncertainty is still excluded") is now implicitly reversed.

docs/stellarphot/user_guide/calibrating_photometry.rst
```diff
@@ -48,17 +48,159 @@ The ``mag_column`` and ``mag_error_column`` arguments name the calibrated
...
+
+Which bands have an error depends on the catalog.
+:class:`~stellarphot.CatalogData` builds ``mag_error_<band>`` for the bands a
+catalog measured itself -- B, V, SG, SR and SI for APASS DR9, the Sloan bands
+for refcat2 -- while the Johnson-Cousins R and I are added afterwards by
+:func:`~stellarphot.utils.magnitude_system_transforms.transform_apass_bands`
+and
+:func:`~stellarphot.utils.magnitude_system_transforms.transform_refcat2_bands`,
+which return magnitudes and no errors. So the fallback is the normal case for
+R and I, the bands most often calibrated, until issue #685 teaches those
```
mwcraig: Stale: #685 is merged on main — `transform_apass_bands` now sets `mag_error_R` / `mag_error_I` (`magnitude_system_transforms.py:515-516, 597`). Pre-existing wording, but this PR is the one moving it into the user guide. Suggest: R/I errors now come from the band transforms, and the fallback applies only where a catalog lacks an error for a star or band.

docs/stellarphot/user_guide/calibrating_photometry.rst
```diff
@@ -48,17 +48,159 @@ The ``mag_column`` and ``mag_error_column`` arguments name the calibrated
...
+
+What the catalog contributes
+----------------------------
+
+A star's catalog entry decides whether it was matched. When a color term
+(``c`` or ``d``) is fit, though, the model also applies that star's catalog
+color, so the color's uncertainty is a real, knowingly omitted contribution
+-- roughly ``c * sigma_color`` -- tracked as issue #691. What the catalog
+reliably contributes is its field-wide scatter about the transform, which
+lands in ``fit_excess_scatter``; and its systematic tie to the standard
```
mwcraig, line 206: This is the sentence the O(1)-vs-O(1/sqrt N) point above hinges on: the catalog's field-wide scatter lands in `fit_excess_scatter`, which is now added per star to the target too. Worth saying here that for a non-catalog target that part is an over-count.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -714,6 +720,14 @@ def _calibrated_with_uncertainty(fit_result, covar, mag_inst, color, errors, fit
     This slightly over-counts the uncertainty of a star that was itself in
     the fit, since its own noise also helped move the coefficients; erring
     high there is deliberate.
+
+    The excess scatter is added after the propagation, not to the star's
+    measurement error before it. `_excess_scatter` measures it in
+    calibrated-magnitude residual space -- against ``hypot(errors,
+    cat_error)`` with no ``1 + a`` factor -- so adding it before the
+    propagation would scale it by the ``fit_diff`` sensitivity a second
```
mwcraig: Nit: "a second time" — `s` is never scaled by `(1 + a)` anywhere, so adding it before the propagation would scale it *once*. The conclusion (add after) is right; maybe "would scale it by the `fit_diff` sensitivity, which it has already absorbed".

stellarphot/utils/magnitude_transforms.py
```diff
@@ -692,6 +694,10 @@ def _calibrated_with_uncertainty(fit_result, covar, mag_inst, color, errors, fit
         instrumental magnitudes, in which case the instrumental magnitude is
         added back here.
 
+    excess_scatter : float, optional
+        The image's ``fit_excess_scatter``, added in quadrature to every
+        star's uncertainty. Non-positive values add nothing.
```
mwcraig: Nit: NaN — the unweighted-fit value — also adds nothing (`nan > 0` is False). "Non-positive or NaN values add nothing."

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1108,126 +1127,48 @@ def transform_to_catalog(
...
+    in magnitudes, that would have to be added in quadrature to every star's
+    uncertainty to bring ``fit_redchi`` to one -- zero when the residuals
+    are already no larger than the errors claim, NaN for an unweighted fit.
+
+    The coefficient uncertainties believe the errors the fit was weighted by
+    -- the quoted errors, floored at ``min_fit_sigma`` -- rather than the
+    scatter observed about the fit, so stars scattering beyond their quoted
+    errors raise ``fit_redchi`` and ``fit_excess_scatter`` instead of
+    growing the ``*_error`` columns. The floor reaches the weighting only:
+    ``fit_redchi``, ``fit_excess_scatter`` and the measurement half of
```
mwcraig: Nit: `mag_cal_error` now has three terms, so "measurement half" → "measurement term" (same at `calibrating_photometry.rst:111`).

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1631,7 +1710,11 @@ def test_transform_to_catalog_uncertainties_scale_with_the_quoted_errors(mocker)
     # error plumbing like issues #680 and #692 invisible in every reported
     # error. Believing the quoted errors means tripling them must triple
     # each one. See issue #690.
-    base_error = 0.02
+    # The base error is quoted generously, above the noise, so that neither
```
mwcraig, line 1713: Once excess > 0, `mag_cal_error` is no longer homogeneous in the quoted errors — the #690 property this test pins (triple the errors, triple every reported error) holds only in the zero-excess regime. Retuning `base_error` to stay there is fine, but the new behaviour deserves a sentence in the Notes or user guide rather than only being dodged here.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1565,16 +1571,87 @@ def test_transform_to_catalog_error_uses_the_unscaled_covariance(
...
+        rtol=1e-6,
+    )
+    np.testing.assert_allclose(reported, _UNDERQUOTED_NOISE_SIGMA, rtol=0.2)
+
+
+def test_transform_to_catalog_error_unchanged_when_errors_describe_the_data(mocker):
+    # The other direction of issue #698: errors quoted generously, above the
+    # true scatter, so the excess is exactly zero rather than a small
+    # positive number that would read as a real finding, and adding zero in
+    # quadrature must leave mag_cal_error bit-identical to what the quoted
```
mwcraig: Nit: the comment says "bit-identical" but the assertion is `rtol=1e-6` against a hand formula, which can't be bit-identical to the `uncertainties` propagation anyway. Also a coverage gap, low risk: no multi-image test asserts each image's `mag_cal_error` uses its *own* excess (the existing multi-image tests are exact fits, excess = 0).

stellarphot/notebooks/transform-to-appas-dr9.ipynb
```diff
@@ -85,7 +85,7 @@
...
...see issue #694). The other three columns are what catch that.\n\n- `fit_cat_error_missing_frac` -- the fraction of the fitted stars whose *catalog* error could not be used. APASS DR9 reports an error of exactly zero for most of its B stars but almost none of its V stars, so this can be large in one band and tiny in another for the same field. Where it is large, that band's `fit_redchi` cannot be compared with another band's.\n- `fit_max_weight_share` -- the largest share of the fit's weight held by any one star. A fit spread evenly over N stars reads `1/N`; anything approaching 1 is one star running the fit.\n- `fit_excess_scatter` -- the scatter, in magnitudes, that would have to be added to every star's uncertainty to bring `fit_redchi` to 1. Flat-field gradients across the field and the catalog's own photometry both land here, and no weighting scheme can repair either. It is also added in quadrature to every star's `mag_cal_error`, which is why that column tracks the observed scatter even when the quoted errors do not.\n\nAn image with a `fit_redchi` far from the others, a `z_error` much larger than the others, or a `fit_max_weight_share` far above `1/N` is the one to look at.",
```
mwcraig: Overclaim: "`mag_cal_error` does not have that problem" — its transform term is still understated by ~`sqrt(fit_redchi)`; only the O(1/N) part is affected, which the user guide says correctly but this doesn't.

CHANGES.rst
```diff
@@ -138,6 +138,9 @@ Other Changes and Additions
   come out smaller than before, and ``fit_excess_scatter`` reports
   explicitly what the rescaling used to absorb silently. Unweighted fits
   still scale by the observed scatter, the only scale they have. [#690]
++ ``mag_cal_error`` from ``transform_to_catalog()`` now adds the image's
+  ``fit_excess_scatter`` in quadrature, so it reflects the scatter actually
+  observed about the transform rather than only the quoted errors. [#698]
```
mwcraig, line 143: Optional: one more line with the recovery formula `sqrt(mag_cal_error**2 - fit_excess_scatter**2)` keeps it within the 3–6 line convention and tells readers of the adjacent #690 entry ("come out smaller than before") that this partly reverses it for `mag_cal_error` only.

mwcraig (commented):

Critical review, four independent passes (statistics, code+tests, docs, design/prior-thread audit) diffed against `upstream/main`. Tests on the branch: 127 passed, 2 skipped. Every round-1 thread is addressed in 30dc328; the loop reorder is safe (no `continue` between `_fit_diagnostics` and the column writes).

**Verified clean:** the excess is in calibrated-magnitude space for both `fit_diff` modes; no double-count of the measurement error; the recovery formula holds in every branch (NaN / 0 / `covar is None`); Quantities and masks never reach `hypot`; no callers of `mag_cal_error` in `gui/`, `differential_photometry/` or `transit_fitting/` (#699 unaffected); notebook diff is one markdown line with no output churn; RST cross-refs resolve and the `-W` docs build passes. The after-vs-through test genuinely discriminates (1e-10 vs 0.2 relative difference against the wrong model).

**Two things worth a decision before merge**, both inline: the excess is added at O(1) to a target that the catalog-scatter part of it only reaches at O(1/sqrt N) — so the column is an upper bound, not quite "the honest number", and it implicitly reverses the #674 changelog's "catalog magnitude uncertainty is still excluded"; and the moved user-guide text still says #685 is open. The rest are wording nits.

One process note: the round-1 reply on the `through_the_transform` negative assertion said "I'd keep it, with a one-line comment" — the commit dropped it instead. No loss of power (the positive assertion discriminates on its own), just a reply/commit mismatch.

stellarphot/utils/magnitude_transforms.py
mwcraig: *Follow-up written by Claude Code at Matt's direction, after testing the claim above on data.*

**Retracting most of this.** Matt asked whether the two sources of excess could be separated, so I ran a variance decomposition on the EY UMa set from the #696 comparison (178 SR images + 8–9 each B/V/SI, same frames calibrated against both APASS and refcat2; script `scratchpad/decompose_excess_scatter_701.py`, reproduces the `fit_excess_scatter` column to 4 decimals). Per-star residuals across images split into a **per-image** part (within-star scatter) and a **static** part (scatter of star means); the static part is then split by cross-catalog correlation — an offset shared between the APASS and refcat2 runs is the star as measured in these frames, the uncorrelated remainder is catalog.

SR, mmag beyond the quoted errors:

| component | refcat2 | APASS | reaches a target at |
|---|---|---|---|
| per-image | 28 | 28 | O(1) |
| static, shared across catalogs | ~39–48 | ~39–48 | O(1) |
| static, catalog-specific | ~20 | ~22 | O(1/√N) |
| `fit_excess_scatter` | 52–57 | 40–41 | |

The per-image part agreeing between catalogs to 0.4 mmag validates the method. The shared static part correlates 0.76 between catalogs overall and 0.65 for stars fainter than 13.5, where refcat2 is Pan-STARRS and independent of APASS — so it isn't shared provenance. It is not a gradient (a plane removes <10% of it) and has no color or magnitude dependence: per-star systematics the photon-noise `mag_error` cannot know about (neighbours in the forced aperture, local background), which a target suffers the same way. Same picture in V, B and SI.

For a typical SR star (obs err 0.0195), taking obs ⊕ per-image ⊕ shared-static as the honest target error: pre-#701 0.020, **#701 0.060 vs honest 0.051 (×1.17) with refcat2; #701 0.045 vs 0.052 (×0.87) with APASS.** The APASS case understates because its generous quoted errors (46 mmag rms vs ~22 actual) absorb real observational static scatter into "quoted", where the excess never sees it.

So the assumption behind my comment — that star-fixed excess must be catalog noise — is wrong on real data; only ~1/3 of the excess variance is catalog, and routing the excess through the covariance instead would have given ~0.020 against an honest 0.051. #701's per-star treatment is the right order and within ~20% in either direction depending on the catalog. The only thing I'd still suggest is a sentence in the Notes saying the excess is added per star because, measured across images and catalogs, most of it is per-star observational scatter, with catalog noise a minority over-count that the catalog over-quoting pays back. Separating the two in production would need a second catalog and buys ~10–20%.

docs/stellarphot/user_guide/calibrating_photometry.rst
mwcraig: *Claude, at Matt's direction:* withdrawn — see the follow-up on the `magnitude_transforms.py:1164` thread. Measured across images and two catalogs, most of the field-wide scatter here is per-star observational, so adding it to a non-catalog target is right to within ~20%.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: add to the notes

CHANGES.rst
mwcraig: Nobody reads a changelog like that...

docs/stellarphot/user_guide/calibrating_photometry.rst
mwcraig: Done in 3309093: R and I now get their errors from the band transforms (a floor, since the USNO'-to-SDSS DR7 step has no published residual), and the fallback is described as the exception for a star or band the catalog has no error for. — *Claude, at Matt's direction.*

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 3309093: "would scale it by a `1 + a` sensitivity it already carries". — *Claude, at Matt's direction.*

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 3309093. — *Claude, at Matt's direction.*

stellarphot/utils/magnitude_transforms.py
mwcraig: Done in 3309093, both places. — *Claude, at Matt's direction.*

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Added to the `mag_cal_error` paragraph of `transform_to_catalog`'s Notes in 3309093: once the excess is above zero, `mag_cal_error` no longer scales with the quoted errors the way `a_error` through `z_error` do. — *Claude, at Matt's direction.*

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Done in 3309093: the comment now says "equal, to the 1e-6 the hand formula is checked at". Coverage gap filled with `test_transform_to_catalog_error_uses_each_images_own_excess_scatter`: two images, same under-quoted error, different noise, each checked against the hand formula with its own excess. It fails under a mutation that feeds every image one constant excess (50/50 rows mismatched). The two-image setup is factored into `_fit_two_images`, which the two existing multi-image tests now use too. — *Claude, at Matt's direction.*

stellarphot/notebooks/transform-to-appas-dr9.ipynb
mwcraig: Done in 3309093: "`mag_cal_error` is mostly immune to it: ... only its transform term, the small part, still carries that understatement." — *Claude, at Matt's direction.*

stellarphot/utils/magnitude_transforms.py
mwcraig: Per the 👍, 3309093 adds the Notes sentence: the excess is added per star because, measured across many images against two catalogs, most of it is observational scatter a target suffers too; the catalog's own noise is the minority, over-counted by of order 20% and under-counted by about as much when a catalog over-quotes. The user guide's catalog paragraph gets a one-line pointer to it. — *Claude, at Matt's direction.*

PR #702, opened 2026-08-22
Believe the quoted weights in the transit fit's uncertainties

### LINKED ISSUE #699: Transit fit still rescales uncertainties by redchi (scale_covar default) — inconsistent with #696

mwcraig:
Closes #699 — brings `TransitModelFit` onto the uncertainty convention #696 adopted for `transform_to_catalog`.

- `_run_fit` passes `scale_covar=self.weights is None`: a weighted fit's `stderr` values now believe the quoted flux errors; an unweighted fit keeps the rescaling since the residuals are its only noise estimate. Before this, errors quoted 10× too small gave the *same* `stderr` (the failing test showed 0.01133 vs 0.01135).
- New attributes `fit_redchi` and `fit_excess_scatter` (None before `fit()`, NaN excess when unweighted), computed before any state is assigned so a failing fit still leaves the instance untouched.
- `_quoted_redchi`/`_excess_scatter` move from `magnitude_transforms.py` to a new `stellarphot.utils.fit_diagnostics` module as public `quoted_redchi`/`excess_scatter` (numpy + `scipy.optimize.brentq` only; automodapi entry added; docs build with `-W` passes). No behaviour change on the transform side.
- `compare_detrend_options` unchanged (BIC is independent of `scale_covar`). No weight floor in this PR.
- Tests: stderr ratio 0.1 between under-quoted and truthful weights with `fit_redchi ≈ 100`; excess ≈ `sqrt(noise² − quoted²)`; unweighted stderr = truthful stderr × `sqrt(fit_redchi)` (the exact relation `scale_covar` implements); attributes None before fit.
- One markdown cell added after `mod.params` in both TESS model-fit notebooks; CHANGES entry.

This PR was written by Claude Code at Matt's direction.

https://claude.ai/code/session_015RURMHNvNLcGuqnx4u7D4V

### DIFF
--- a/stellarphot/transit_fitting/core.py
+++ b/stellarphot/transit_fitting/core.py
@@ -5,6 +5,8 @@ from astropy import units as u
 from astropy.table import Table
 from pydantic import BaseModel
 
+from ..utils.fit_diagnostics import excess_scatter
+
 try:
     from pytransit import RoadRunnerModel
 except ImportError:  # pragma: no cover
@@ -98,6 +100,19 @@ class TransitModelFit:
         Result of the most recent call to ``fit``, including fit statistics
         like ``bic`` and ``nvarys``. ``None`` until ``fit`` has been run.
 
+    fit_redchi : float or None
+        Reduced chi-square of the most recent fit: the summed squared
+        residuals per degree of freedom, in units of the quoted errors when
+        ``weights`` are set and in flux units squared otherwise. ``None``
+        until ``fit`` has been run.
+
+    fit_excess_scatter : float or None
+        Scatter, in the units of ``data``, that would have to be added in
+        quadrature to every point's quoted error to bring ``fit_redchi`` to
+        one; zero when the residuals are already no larger than the errors
+        claim, and NaN for an unweighted fit. ``None`` until ``fit`` has
+        been run.
+
     times, airmass, width, spp, data, weights : array-like or None
         Independent variables and data for the fit; see the property
         docstrings.
@@ -111,6 +126,17 @@ class TransitModelFit:
     set. Enable them via `TransitModelOptions`, directly (e.g.
     ``mod.params["airmass_trend"].vary = True``) or with
     ``compare_detrend_options(apply_best=True)``.
+
+    The parameter uncertainties of a weighted fit are measured against the
+    flux errors as quoted (``weights = 1 / error``): lmfit's default of
+    rescaling the covariance so that the reduced chi-square comes out one
+    is turned off, as it is in
+    `~stellarphot.utils.magnitude_transforms.transform_to_catalog` (issues
+    #690 and #699). Errors quoted ten times too small therefore give
+    ``stderr`` values ten times too small -- and a ``fit_redchi`` near 100
+    and a ``fit_excess_scatter`` near the real noise to say so. An
+    unweighted fit keeps the rescaling, since it has no quoted errors and
+    the residuals are then its only estimate of the noise.
 
     Examples
     --------
@@ -128,6 +154,11 @@ class TransitModelFit:
 
         print(mod.params["rp"].value, mod.params["rp"].stderr)
         print(result.bic)
+
+        # Do the flux errors describe the data? A reduced chi-square well
+        # above one, and an excess scatter comparable to the errors, say
+        # they are too small -- and so are the stderr values above.
+        print(mod.fit_redchi, mod.fit_excess_scatter)
 
         # Which detrending parameters does the BIC favor?
         bic_table = mod.compare_detrend_options()
@@ -160,6 +191,8 @@ class TransitModelFit:
         self.weights = None
         self.params = _default_params()
         self.fit_result = None
+        self.fit_redchi = None
+        self.fit_excess_scatter = None
         self._detrend_parameters = set()
         self._all_detrend_params = ["airmass", "width", "spp"]
 
@@ -331,10 +364,22 @@ class TransitModelFit:
         # handles a starting value that sits exactly on a bound (the default
         # inclination of 90 degrees is on its upper bound); MINPACK leastsq
         # does not.
+        #
+        # The stderr of a weighted fit is taken at face value rather than
+        # rescaled so that its reduced chi-square is one: lmfit's default
+        # ``scale_covar=True`` would make that rescaling silently, and so
+        # erase the one statistic that shows whether the quoted flux errors
+        # describe the data. A mismatch is reported in ``fit_redchi`` and
+        # ``fit_excess_scatter`` instead, the same convention as
+        # ``transform_to_catalog`` (issues #690 and #699). An unweighted fit
+        # has no quoted errors to believe, so the rescaling stays on there,
+        # with redchi -- in flux units squared -- as the one estimate of the
+        # noise available.
         return lmfit.minimize(
             self._residual,
             params,
             method="least_squares",
+            scale_covar=self.weights is None,
             args=(
                 covariates["airmass"],
                 covariates["width"],
@@ -491,7 +536,16 @@ class TransitModelFit:
         ------
         ValueError
             If the times or the data have not been set. If the fit itself
-            raises, ``params`` and ``fit_result`` are left untouched.
+            raises, ``params``, ``fit_result``, ``fit_redchi`` and
+            ``fit_excess_scatter`` are left untouched.
+
+        Notes
+        -----
+        When ``weights`` are set, the uncertainties written to ``params`` are
+        measured against the errors as quoted, not rescaled to force the
+        reduced chi-square to one (lmfit's ``scale_covar`` is off). Whether
+        the quoted errors describe the data is reported instead, in
+        ``fit_redchi`` and ``fit_excess_scatter``; see the class docstring.
         """
         if self.times is None:
             raise ValueError("The times must be set before trying to fit.")
@@ -500,7 +554,18 @@ class TransitModelFit:
 
         result = self._run_fit(self.params)
 
+        # Every diagnostic is computed before any state is assigned, so a
+        # fit that raises leaves the instance exactly as it was.
+        if self.weights is None:
+            sigma = None
+        else:
+            sigma = 1 / np.asarray(self.weights, dtype=float)
+        redchi = result.redchi
+        excess = excess_scatter(result, sigma, self.weights, redchi_quoted=redchi)
+
         self.fit_result = result
+        self.fit_redchi = redchi
+        self.fit_excess_scatter = excess
 
         # Copy the best-fit values and uncertainties back into the user's
         # parameters without touching vary/min/max, so user choices (e.g. a
--- a/stellarphot/transit_fitting/core.py
+++ b/stellarphot/transit_fitting/core.py
@@ -5,7 +5,7 @@ from astropy import units as u
 from astropy.table import Table
 from pydantic import BaseModel
 
-from ..utils.fit_diagnostics import excess_scatter
+from ..utils.fit_diagnostics import excess_scatter, quoted_redchi
 
 try:
     from pytransit import RoadRunnerModel
@@ -112,6 +112,12 @@ class TransitModelFit:
         one; zero when the residuals are already no larger than the errors
         claim, and NaN for an unweighted fit. ``None`` until ``fit`` has
         been run.
+
+        Both attributes describe the most recent call to ``fit``.
+        ``compare_detrend_options()`` alone does not update them, since it
+        fits each candidate combination directly; called with
+        ``apply_best=True`` it does, because it finishes by calling ``fit``
+        on the winning combination.
 
     times, airmass, width, spp, data, weights : array-like or None
         Independent variables and data for the fit; see the property
@@ -549,8 +555,12 @@ class TransitModelFit:
         if self.weights is None:
             sigma = None
         else:
-            sigma = 1 / np.asarray(self.weights, dtype=float)
-        redchi = result.redchi
+            # A weight of zero -- lmfit's idiom for excluding a point -- maps
+            # to an infinite sigma here; quoted_redchi and excess_scatter
+            # both drop such points rather than dividing by that weight.
+            with np.errstate(divide="ignore"):
+                sigma = 1 / np.asarray(self.weights, dtype=float)
+        redchi = quoted_redchi(result, sigma, self.weights)
         excess = excess_scatter(result, sigma, self.weights, redchi_quoted=redchi)
 
         self.fit_result = result
--- a/stellarphot/transit_fitting/core.py
+++ b/stellarphot/transit_fitting/core.py
@@ -365,16 +365,8 @@ class TransitModelFit:
         # inclination of 90 degrees is on its upper bound); MINPACK leastsq
         # does not.
         #
-        # The stderr of a weighted fit is taken at face value rather than
-        # rescaled so that its reduced chi-square is one: lmfit's default
-        # ``scale_covar=True`` would make that rescaling silently, and so
-        # erase the one statistic that shows whether the quoted flux errors
-        # describe the data. A mismatch is reported in ``fit_redchi`` and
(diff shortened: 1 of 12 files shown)


stellarphot/transit_fitting/core.py
```diff
@@ -500,7 +544,18 @@ def fit(self):
 
         result = self._run_fit(self.params)
 
+        # Every diagnostic is computed before any state is assigned, so a
+        # fit that raises leaves the instance exactly as it was.
+        if self.weights is None:
+            sigma = None
+        else:
+            sigma = 1 / np.asarray(self.weights, dtype=float)
```
mwcraig: **Regression: a zero weight now makes `fit()` raise after lmfit succeeded.** `weights` is an unvalidated attribute and lmfit tolerates `weight == 0` (the point is just ignored), which is the natural way to mask a point. On `main` that fit succeeds. Here `sigma` becomes `inf`, `residual / weights` in `excess_scatter` is `0/0 = NaN`, and when the quoted reduced chi-square is above 1 the bracket function evaluates NaN at zero. Reproduced on this branch with `weights[3] = 0` and under-quoted errors:

```
ValueError: The function value at x=0.0 is NaN; solver cannot continue.
```

plus a `divide by zero` RuntimeWarning here (an error under the repo's `filterwarnings = error`). When redchi ≤ 1 it silently returns `excess = 0`, so the failure is data-dependent. Suggest masking `weights == 0` points out of the diagnostics (they contribute nothing to chi-square anyway), or validating weights and documenting the precondition. Copilot's `0/0` comment is this same issue; it's a crash, not a NaN result.

*Review written by Claude Code at Matt's direction.*

stellarphot/transit_fitting/core.py
```diff
@@ -500,7 +544,18 @@ def fit(self):
...
+        redchi = result.redchi
```
mwcraig: Use `quoted_redchi(result, sigma, self.weights)` here rather than `result.redchi`. lmfit sums `residual**2` while the bracket in `excess_scatter` sums `(residual/w)**2 / sigma**2`, and the two round differently: in a 200k-trial simulation of fits whose reduced chi-square sits within a few ULP of 1, ~3% had lmfit's value `> 1` while the bracket value at zero was `≤ 1`, and `brentq` then raises on two negative endpoints (scipy only tolerates `f(a) == 0` exactly). Practically this needs redchi within ~1e-15 of 1, so the risk is tiny — but the whole reason `redchi_quoted` is passed in (per the comments in `fit_diagnostics.py`) is that the gate value *is* the reported value, and this caller breaks that contract for free. `quoted_redchi` returns `result.redchi` unchanged when `sigma is None`, so the unweighted test still holds.

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,176 @@
...
+    if sigma is None:
+        return fit_result.redchi
+
+    residual = np.asarray(fit_result.residual) / weights
+    # The same operand order as `excess_scatter`'s bracket function at zero
+    # excess -- ``residual**2 / sigma**2``, not ``(residual / sigma)**2``.
+    # The gate there compares this value to one, and a value that disagreed
+    # with the bracket by one ULP around it could hand
+    # `~scipy.optimize.brentq` two negative endpoints -- a `ValueError`.
+    return float(np.sum(residual**2 / sigma**2) / fit_result.nfree)
```
mwcraig: Now that this is public: lmfit computes `redchi = chisqr / max(1, nfree)`, while this divides by `nfree` unguarded, so `nfree == 0` gives `inf`/`ZeroDivisionError` where lmfit gives `chisqr`. `excess_scatter` guards `nfree <= 0` but this doesn't. Pre-existing from #696, but worth either mirroring lmfit or stating the precondition — and if the transit caller switches to `quoted_redchi` (comment above) an over-parameterized transit fit becomes exposed to it.

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,176 @@
+"""
+Diagnostics of a weighted least-squares fit: how the residuals compare to the
+uncertainties the points were quoted with.
+
+Shared by `~stellarphot.utils.magnitude_transforms.transform_to_catalog` and
+`~stellarphot.transit_fitting.TransitModelFit`, both of which fit with
+``scale_covar=False`` (issues #690 and #699) and report these numbers instead
```
mwcraig: Not literally true: both callers use `scale_covar=(weights/sigma) is None` (`magnitude_transforms.py:1400`, `core.py:374`), so unweighted fits keep the rescaling. "both of which turn off the rescaling for a weighted fit" would match. (Agree with Copilot here.)

mwcraig (commented):
Critical review of the branch at 16daf58. Ran the transit and magnitude-transform test files locally (160 passed, 2 remote-data skipped); the `_excess_scatter` → `excess_scatter` rename is complete, there is no circular import, the new automodapi entry is needed (the functions aren't in `magnitude_transforms.__all__`), the unweighted stderr × √redchi relation the test asserts is exactly what lmfit implements, and the diagnostics really are computed before any state is assigned. Inline comments cover the substantive points; the one that should block is the zero-weight regression on `core.py` line 552.

Three smaller things not worth an inline thread:

- `compare_detrend_options()` calls `_run_fit` per combination and never `fit()`, so after `compare_detrend_options()` (default `apply_best=False`) the new `fit_redchi` / `fit_excess_scatter` still describe whatever fit came *before* the comparison. That's consistent with how `fit_result` already behaves, but the new attributes invite "check the diagnostics after comparing", so a sentence in their attribute docs would head that off.

- `test_fit_stderr_believes_quoted_weights` takes ~7.8 s (the under-quoted fit converges slowly); the other three new tests take 0.03 s combined. Not blocking, but it's now the slowest test in the file.
- CHANGES: "follows the same convention" implicitly means the #690 entry three bullets up; "the convention #690 adopted for `transform_to_catalog`" would make that explicit.

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,176 @@
...
+
+    redchi_quoted : float, optional
+        ``quoted_redchi(fit_result, sigma, weights)`` for this fit, if the
+        caller already has it -- computed here otherwise. Ignored when
+        ``sigma`` is `None`.
+
+    Returns
+    -------
+    float
+        The excess scatter in magnitudes; zero when the residuals are already
```
mwcraig: "in magnitudes" is wrong for the transit caller, where this is in flux units — "in the units of `sigma`" covers both. The flat-field / catalog-photometry sentence in the summary paragraph is also transform-specific; fine to keep if labelled as that case. (Agree with Copilot here.)

stellarphot/transit_fitting/core.py
```diff
@@ -500,7 +544,18 @@ def fit(self):
...
+        excess = excess_scatter(result, sigma, self.weights, redchi_quoted=redchi)
```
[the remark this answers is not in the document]
mwcraig: Agreed — and this is the same point as the review comment on line 553. The helper's whole reason for taking `redchi_quoted` is that the gate is the reported float, and `_run_fit`'s caller dropped that by handing over `result.redchi`. Will compute `redchi = quoted_redchi(result, sigma, self.weights)` and pass that through; `quoted_redchi` returns `result.redchi` unchanged when `sigma is None`, so the unweighted path and its test are unaffected.

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,176 @@
...
+
+    This is also the value `excess_scatter` gates on: whether any scatter
+    needs inventing at all is exactly the question of whether this reduced
+    chi-square already sits at or below one, so the two share this one
+    computation rather than each summing the same quantity independently.
+    """
+    if sigma is None:
+        return fit_result.redchi
+
+    residual = np.asarray(fit_result.residual) / weights
```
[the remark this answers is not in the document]
mwcraig: Confirmed — reproduced with a 30-point linear fit, `weights[3] = 0`: lmfit fits fine, then both `quoted_redchi` and `excess_scatter` hit `0 / 0` here (a `RuntimeWarning`, which the repo's `filterwarnings = error` turns into an exception under pytest). Will fix in both places by dropping zero-weight points from the sums rather than documenting a contract: they contribute exactly zero to chi-square, so the result for every other input is unchanged. Concretely, `good = np.broadcast_to(weights, residual.shape) != 0` and then index `residual`, `weights` and `sigma` by `good`; `nfree` stays lmfit's, which counts those points in `ndata` — matching the unweighted branch that returns `fit_result.redchi` as-is.

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,176 @@
...
+        ``quoted_redchi(fit_result, sigma, weights)`` for this fit, if the
+        caller already has it -- computed here otherwise. Ignored when
+        ``sigma`` is `None`.
+
+    Returns
+    -------
+    float
+        The excess scatter in magnitudes; zero when the residuals are already
+        no larger than the errors claim, and NaN for an unweighted fit, whose
+        residuals have no errors to be excessive with respect to.
```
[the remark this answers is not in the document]
mwcraig: Agreed — "in magnitudes" was carried over from when this lived in `magnitude_transforms.py`. Will change to "in the units of `sigma`" and label the flat-field / catalog sentence in the summary as the transform case.

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,176 @@
+"""
+Diagnostics of a weighted least-squares fit: how the residuals compare to the
+uncertainties the points were quoted with.
+
+Shared by `~stellarphot.utils.magnitude_transforms.transform_to_catalog` and
+`~stellarphot.transit_fitting.TransitModelFit`, both of which fit with
+``scale_covar=False`` (issues #690 and #699) and report these numbers instead
+of letting `lmfit` rescale the covariance to make the reduced chi-square one.
```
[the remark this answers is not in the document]
mwcraig: Agreed — both callers use `scale_covar=(sigma is None)` (`magnitude_transforms.py:1400`, `core.py:374`), so only the weighted fit turns the rescaling off. Will reword to "both of which turn off the rescaling for a weighted fit".

stellarphot/transit_fitting/core.py
mwcraig: Confirmed, and it's worse than data-dependent under the test suite: the `divide by zero` `RuntimeWarning` fires in `residual / weights` before the redchi gate, so with `filterwarnings = error` any zero weight raises regardless of redchi. Outside pytest it's as described — silent `excess = 0` when redchi ≤ 1, `ValueError` from `brentq` otherwise. Copilot's suppressed comment on `fit_diagnostics.py:153` and its inline one on line 64 are this same bug.

Going with masking rather than validation, since zero weights are lmfit's own masking idiom and `main` accepts them: in both `quoted_redchi` and `excess_scatter`, select `weights != 0` (broadcast, so a scalar weight still works) before dividing. Those points contribute zero to chi-square so nothing else changes; `nfree` stays lmfit's value. Will add a test with a zero weight under both redchi > 1 and ≤ 1, plus a `quoted_redchi` test.

stellarphot/transit_fitting/core.py
mwcraig: Agreed; same fix as Copilot's comment on line 554 — `redchi = quoted_redchi(result, sigma, self.weights)` and pass it as `redchi_quoted`. This also makes the transit caller consistent with `transform_to_catalog`, which already does exactly that.

stellarphot/utils/fit_diagnostics.py
mwcraig: Verified against lmfit 1.3.4: `MinimizerResult._calculate_statistics` does `self.redchi = self.chisqr / max(1, self.nfree)`. Will mirror that here (`/ max(1, fit_result.nfree)`) so `quoted_redchi` agrees with `fit_result.redchi` in the degenerate case too, which matters now that the transit caller will route through it. `excess_scatter`'s `nfree <= 0` → NaN stays as is; that one is a separate question (there is no scatter to find with no degrees of freedom).

stellarphot/utils/fit_diagnostics.py
mwcraig: Agreed — will use that wording. Same as Copilot's comment on line 8.

stellarphot/utils/fit_diagnostics.py
mwcraig: Agreed — "in the units of `sigma`", and the flat-field / catalog sentence gets labelled as the magnitude-transform case. Same as Copilot's comment on line 110.

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,200 @@
...
+        The excess scatter, in the units of ``sigma``; zero when the
+        residuals are already no larger than the errors claim, and NaN for
+        an unweighted fit, whose residuals have no errors to be excessive
+        with respect to.
+
+    Notes
+    -----
+    Reported rather than folded into the weights. A fit that absorbs its own
+    excess scatter has a reduced chi-square of one by construction, which
+    destroys the one diagnostic that revealed any of this.
```
mwcraig, line 122: It is completely unclear what "this" refers to, whichmeans it probably isn't needed

stellarphot/utils/fit_diagnostics.py
```diff
@@ -0,0 +1,200 @@
...
+    that is the same for every point barely changes where a fit lands -- it
+    rescales the weights nearly uniformly -- and holding them fixed keeps this
+    a description of the fit that was actually reported.
+
+    Callers such as
+    `~stellarphot.utils.magnitude_transforms.transform_to_catalog` pass
+    ``redchi_quoted`` in so that the value gating this function is the
+    exact float they report as ``fit_redchi``,
+    rather than a second sum of the same quantity that could round
+    differently by a few ULPs around 1.0.
```
mwcraig: Do we really need to worry about rounding at machine precision???

stellarphot/utils/fit_diagnostics.py
mwcraig: Gone in 7b9fada. The paragraph existed to justify the `redchi_quoted` parameter, and that parameter is gone too: `excess_scatter` now gates on its own bracket function evaluated at zero, so there is only one sum and nothing for "this" to refer to.

stellarphot/utils/fit_diagnostics.py
mwcraig: No — and in 7b9fada we no longer do. Both ULP guards were defending against computing the same number twice. The gate now reuses the bracket function (`f(0) <= 0` → return 0), and the upper endpoint is `2 * rms` instead of `rms`, which makes `f(upper) <= -3/4` by construction rather than "at most zero, barring rounding". The rounds-to-positive guard and its comment are deleted; the existing regression tests for both cases still pass.

PR #703, opened 2026-08-23
Derive the catalog search radius from the observed field

mwcraig:
Closes #686

`transform_to_catalog()` fetched its calibration catalog from a cone fixed at one degree around whichever star happened to be first in the table, and `CatalogData.from_vizier` asks for every row in that cone — so a small field pulled a degree-wide catalog no matter how small it really was.

## What changed

The cone is now **centered on the centroid of the observed positions** and **sized to enclose every one of them plus a one arcminute margin**. The centroid is the mean of the positions' unit cartesian vectors, so a field straddling RA 0 does not average to the antipode the way a plain mean of RA would.

A new `search_radius` keyword takes an explicit angle instead. It is validated up front — before the query — so a bad value costs a raise rather than a Vizier round trip.

A *derived* cone wider than one degree warns. That is a cone no call could previously produce, and a table covering that much sky is usually more than one field's worth of `ra`/`dec`. An explicit `search_radius` is a deliberate choice and passes without comment.

## Effect

Measured at the north galactic pole — the sparsest patch of sky there is, so an ordinary field improves by more:

| | rows returned | `remote_data` test |
|---|---|---|
| `main` (1 degree) | 5778 | 9.6 s |
| this branch (11 arcmin) | 222 | 5.6 s |

## Notes

- `row_limit=-1` in `from_vizier` is deliberately left alone; the cone now bounds the row count.
- `_to_float_array` could not be reused for `ra`/`dec` — it raises `UnitConversionError` for any unit that is neither dimensionless nor mag-equivalent, and coordinate columns carry `deg`. A sibling `_to_degree_array` handles angles.
- The `remote_data` test's field comment explained the old degree-wide workaround and has been rewritten. The galactic-pole field itself is kept: it still buys a quick answer to the *separate* ten-arcminute query that supplies the observed stars.

## Testing

12 new tests for the derived and explicit radius, the argument validation, and `_observed_field` itself (single star, the RA wrap, masked/NaN positions, no finite position); 3 more for the wide-field warning. Full suite: 1018 passed. The `remote_data` test passes against both catalogs.

https://claude.ai/code/session_018zVojGsHye1BX8afuo9Yig

### DIFF
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -5,7 +5,7 @@ from contextlib import contextmanager
 import lmfit
 import numpy as np
 from astropy import units as u
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import SkyCoord, UnitSphericalRepresentation
 from astropy.utils.exceptions import AstropyUserWarning
 from numpy.linalg import LinAlgError
 from uncertainties import correlated_values, unumpy
@@ -96,6 +96,24 @@ _FIT_DIAGNOSTIC_COLUMNS = (
 # to calibrate everything else against.
 _FIT_MATCH_ARCSEC = 1.0
 _CAL_MATCH_ARCSEC = 1.5
+
+# Added to the radius of the cone the calibration catalog is fetched from, so
+# that the cone is a little larger than the field the observations cover. A
+# star at the very edge of the field still has to find its catalog counterpart
+# inside the cone, and a counterpart is at most `_CAL_MATCH_ARCSEC` away, so an
+# arcminute is generous for that alone. It is also what gives a field whose
+# positions span nothing at all -- a single star, or one image of one target --
+# a cone of a sensible size rather than of no size.
+_CATALOG_RADIUS_MARGIN = 1 * u.arcmin
+
+# Derived cone size above which the caller is told what is about to be asked
+# of Vizier. One degree because that is exactly what the radius used to be
+# fixed at: a cone wider than this is one no call could previously produce,
+# and a field that large is usually a mosaic, or ra/dec columns holding more
+# than one field's worth of stars, rather than a single image. Only a cone
+# that came out this wide on its own is worth a warning -- an explicit
+# `search_radius` is a deliberate choice and passes without comment.
+_WIDE_FIELD_RADIUS = 1 * u.degree
 
 # The color conventionally used with each canonical passband, for a caller
 # who names a band but no color. Which color goes with a band is a convention
@@ -277,6 +295,93 @@ def _to_float_array(values):
         )
 
     return (filled * unit).to_value(u.mag)
+
+
+def _to_degree_array(values):
+    """
+    Convert a coordinate column to plain degrees, masked entries becoming NaN.
+
+    The sibling of `_to_float_array` for angles rather than magnitudes: that
+    one raises for a unit like ``deg``, since a magnitude column carrying one
+    is a mistake, and coordinate columns carry one routinely.
+
+    Parameters
+    ----------
+
+    values : array-like
+        Values to convert. May be a `~astropy.table.MaskedColumn` or a
+        `~astropy.units.Quantity`. Values with no unit are taken to be in
+        degrees, which is the convention the ``ra`` and ``dec`` columns of a
+        photometry table follow.
+
+    Returns
+    -------
+    `numpy.ndarray`
+        Float array of degrees, NaN wherever the input was masked.
+    """
+    unit = getattr(values, "unit", None)
+    filled = np.asarray(
+        np.ma.filled(np.ma.asarray(values, dtype=float), np.nan), dtype=float
+    )
+
+    return filled if unit is None else (filled * unit).to_value(u.degree)
+
+
+def _observed_field(ra, dec):
+    """
+    Find the patch of sky the observed positions cover.
+
+    Parameters
+    ----------
+
+    ra, dec : array-like
+        Right ascension and declination of the observed positions, in the
+        form `_to_degree_array` accepts.
+
+    Returns
+    -------
+    center : `astropy.coordinates.SkyCoord`
+        Centroid of the positions.
+
+    radius : `astropy.units.Quantity`
+        Angular radius about ``center`` enclosing every position, plus
+        `_CATALOG_RADIUS_MARGIN`.
+
+    Raises
+    ------
+    `ValueError`
+        If no position has both a finite right ascension and a finite
+        declination.
+
+    Notes
+    -----
+
+    The center is the mean of the positions' unit cartesian vectors rather
+    than the mean of their right ascensions and declinations. The latter is
+    wrong across the RA = 0 wrap, and wrong by half the sky: a field
+    straddling it averages to the antipode, and a cone drawn there holds none
+    of the stars it was meant to hold.
+
+    A row is dropped unless both of its coordinates are usable, since half a
+    position places nothing, and a masked or NaN entry left in would poison
+    the mean rather than merely be ignored by it.
+    """
+    ra_degree = _to_degree_array(ra)
+    dec_degree = _to_degree_array(dec)
+    usable = np.isfinite(ra_degree) & np.isfinite(dec_degree)
+    if not usable.any():
+        raise ValueError(
+            "No observed position has both a finite ra and a finite dec, so "
+            "there is no field to search a catalog around."
+        )
+
+    coords = SkyCoord(ra_degree[usable], dec_degree[usable], unit="degree")
+    direction = coords.cartesian.mean().represent_as(UnitSphericalRepresentation)
+    center = SkyCoord(
+        direction.lon, direction.lat, frame=coords.frame.replicate_without_data()
+    )
+
+    return center, coords.separation(center).max() + _CATALOG_RADIUS_MARGIN
 
 
 def _underdetermined_reason(fit_result, vary):
@@ -842,6 +947,7 @@ def transform_to_catalog(
     in_place=True,
     fit_diff=True,
     min_fit_sigma=_MIN_FIT_SIGMA,
+    search_radius=None,
 ):
     """
     Transform a set of instrumental magnitudes to a standard system using either
@@ -949,6 +1055,12 @@ def transform_to_catalog(
         :ref:`calibration-fit-quality` for how the floor reaches the
         reported uncertainties.
 
+    search_radius : `astropy.units.Quantity`, optional
+        Angular radius of the cone the calibration catalog is fetched from,
+        which must be a positive, finite angle. Defaults to the radius that
+        encloses every observed position, plus a one arcminute margin, about
+        the centroid of those positions. See Notes.
+
     Returns
     -------
 
@@ -1039,6 +1151,18 @@ def transform_to_catalog(
     ``mag_cal_error`` is NaN, rather than falling back to the measurement
     error alone, for an image whose fit left no usable covariance behind.
 
+    The catalog is fetched once per call, from a cone centered on the
+    centroid of the observed positions and just large enough to hold them
+    all. It used to be a fixed degree around whichever star happened to be
+    first in the table, so a call asked Vizier for every row of a degree-wide
+    cone however small the field really was; see issue #686.
+    ``search_radius`` is therefore the way to make the cone deliberately
+    *larger* than the field -- to calibrate against stars outside it, say --
+    rather than the way to keep a query small, which the default already
+    does. A derived cone wider than a degree warns, since a table covering
+    that much sky is usually more than one field; passing ``search_radius``
+    says the size was intended and silences it.
+
     How the fit is weighted, what the floor does and does not reach, how to
     read the diagnostics and the coefficient uncertainties, and what the
     catalog contributes that none of these columns carry are described in
@@ -1065,6 +1189,23 @@ def transform_to_catalog(
             f"{min_fit_sigma}. Pass 0 to apply no floor to the uncertainties "
             "the fit weights by."
         )
+
+    # Checked here rather than where the cone is actually drawn, so that an
+    # argument the caller can fix without leaving their keyboard does not cost
+    # them a Vizier round trip first.
+    if search_radius is not None:
+        search_radius = u.Quantity(search_radius)
+        if not search_radius.unit.is_equivalent(u.degree):
+            raise ValueError(
+                "search_radius must be an angle, such as 5 * u.arcmin, got "
+                f"{search_radius!r}."
+            )
+        if not np.isfinite(search_radius.value) or search_radius.value <= 0:
+            raise ValueError(
+                "search_radius must be a positive, finite angle, got "
+                f"{search_radius!r}. Leave it out to search the field the "
+                "observations cover."
+            )
 
     # Preserve the order the caller gave, minus any duplicates.
     vary = tuple(dict.fromkeys(vary))
@@ -1158,17 +1299,37 @@ def transform_to_catalog(
     cat_mags = np.full(n_rows, np.nan)
     cat_colors = np.full(n_rows, np.nan)
 
-    one_coord = SkyCoord(
-        observed_mags_grouped["ra"][0], observed_mags_grouped["dec"][0], unit="degree"
+    # Every row of the table, rather than only the rows in this passband:
+    # the images in it are of one field, and one cone drawn around all of
+    # them serves the calls for the other passbands too.
+    field_center, field_radius = _observed_field(
+        observed_mags_grouped["ra"], observed_mags_grouped["dec"]
     )
+    if search_radius is None:
+        search_radius = field_radius
+        if search_radius > _WIDE_FIELD_RADIUS:
+            warnings.warn(
+                f"The observed positions span {field_radius.to(u.degree):.2f}, "
+                "so the catalog will be fetched from a cone that wide and the "
+                "query may be slow. That is usually a table holding more than "
+                "one field; if it is not, pass search_radius to set the cone "
+                "size yourself.",
+                AstropyUserWarning,
+                stacklevel=2,
+            )
+
     if cat_name == "apass_dr9":
-        cat = apass_dr9(one_coord, radius=1 * u.degree, clip_by_frame=False, padding=0)
+        cat = apass_dr9(
+            field_center, radius=search_radius, clip_by_frame=False, padding=0
+        )
         cat = cat.passband_columns(
             passbands=["B", "V", "R", "I", "SR", "SG", "SI"],
             transformer=transform_apass_bands,
         )
     elif cat_name == "refcat2":
-        cat = refcat2(one_coord, radius=1 * u.degree, clip_by_frame=False, padding=0)
+        cat = refcat2(
+            field_center, radius=search_radius, clip_by_frame=False, padding=0
+        )
         cat = cat.passband_columns(
             # Catalog native passbands will be automatically
             # made to.
--- a/stellarphot/utils/magnitude_transforms.py
+++ b/stellarphot/utils/magnitude_transforms.py
@@ -115,6 +115,12 @@ _CATALOG_RADIUS_MARGIN = 1 * u.arcmin
 # `search_radius` is a deliberate choice and passes without comment.
 _WIDE_FIELD_RADIUS = 1 * u.degree
 
+# The catalogs `transform_to_catalog` knows how to fetch. Named here so that
+# the check of the caller's `cat_name`, which happens up front alongside the
+# other argument checks, and the fetch itself, which happens much further
+# down, cannot drift apart.
+_CATALOG_NAMES = ("apass_dr9", "refcat2")
+
 # The color conventionally used with each canonical passband, for a caller
 # who names a band but no color. Which color goes with a band is a convention
 # rather than something derivable, so a band that is not here has to be asked
@@ -1057,9 +1063,9 @@ def transform_to_catalog(
 
     search_radius : `astropy.units.Quantity`, optional
         Angular radius of the cone the calibration catalog is fetched from,
-        which must be a positive, finite angle. Defaults to the radius that
-        encloses every observed position, plus a one arcminute margin, about
-        the centroid of those positions. See Notes.
+        which must be a single positive, finite angle. Defaults to the radius
+        that encloses every position observed in ``obs_filter``, plus a one
+        arcminute margin, about the centroid of those positions. See Notes.
 
     Returns
     -------
@@ -1152,10 +1158,14 @@ def transform_to_catalog(
     error alone, for an image whose fit left no usable covariance behind.
 
     The catalog is fetched once per call, from a cone centered on the
-    centroid of the observed positions and just large enough to hold them
-    all. It used to be a fixed degree around whichever star happened to be
-    first in the table, so a call asked Vizier for every row of a degree-wide
-    cone however small the field really was; see issue #686.
+    centroid of the positions observed in ``obs_filter`` and just large
+    enough to hold them all. Rows in other passbands are left out of it:
+    they are never matched against the catalog, so sizing the cone from them
+    only costs sky -- and the field of view can shift a little between bands
+    even when the images are of one target. It used to be a fixed degree
+    around whichever star happened to be first in the table, so a call asked
+    Vizier for every row of a degree-wide cone however small the field really
+    was; see issue #686.
     ``search_radius`` is therefore the way to make the cone deliberately
     *larger* than the field -- to calibrate against stars outside it, say --
     rather than the way to keep a query small, which the default already
@@ -1200,12 +1210,22 @@ def transform_to_catalog(
                 "search_radius must be an angle, such as 5 * u.arcmin, got "
                 f"{search_radius!r}."
             )
+        if not search_radius.isscalar:
+            raise ValueError(
+                f"search_radius must be a single angle, got {search_radius!r}."
+            )
         if not np.isfinite(search_radius.value) or search_radius.value <= 0:
             raise ValueError(
                 "search_radius must be a positive, finite angle, got "
                 f"{search_radius!r}. Leave it out to search the field the "
                 "observations cover."
             )
+
+    if cat_name not in _CATALOG_NAMES:
+        raise ValueError(
+            f"Unknown catalog name {cat_name}. Must be one of "
+            f"{' or '.join(repr(name) for name in _CATALOG_NAMES)}."
+        )
 
     # Preserve the order the caller gave, minus any duplicates.
     vary = tuple(dict.fromkeys(vary))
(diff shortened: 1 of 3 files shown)


stellarphot/utils/magnitude_transforms.py
```diff
@@ -1158,17 +1299,37 @@ def transform_to_catalog(
...
     cat_colors = np.full(n_rows, np.nan)
 
-    one_coord = SkyCoord(
-        observed_mags_grouped["ra"][0], observed_mags_grouped["dec"][0], unit="degree"
+    # Every row of the table, rather than only the rows in this passband:
+    # the images in it are of one field, and one cone drawn around all of
+    # them serves the calls for the other passbands too.
+    field_center, field_radius = _observed_field(
+        observed_mags_grouped["ra"], observed_mags_grouped["dec"]
     )
```
mwcraig: **The cone is sized from rows that will never be calibrated.**

`in_passband` is computed at line 1239, and every downstream step is restricted to it — the per-image fit (line 1434) and the output write (line 1728). Rows in other passbands are never matched against the catalog, so sizing the cone from them buys nothing and costs sky.

The comment's justification also doesn't hold: *"one cone drawn around all of them serves the calls for the other passbands too."* Nothing is shared between calls — each call re-derives its own cone from scratch. The canonical usage in `transform-to-appas-dr9.ipynb` groups by passband first and loops, so each iteration recomputes the same cone independently rather than reusing one.

The failure this enables: a table holding a night of observations of several targets in several bands. Only one target's rows get fit, but the cone is drawn around *all* of them — a multi-degree query where an arcminute one would do, plus a spurious `_WIDE_FIELD_RADIUS` warning. That is the defect #686 was filed about, reintroduced through a different door.

```suggestion
    # Only the rows in this passband: they are the only ones ever matched
    # against the catalog, so they are the only ones the cone has to cover.
    field_center, field_radius = _observed_field(
        observed_mags_grouped["ra"][in_passband],
        observed_mags_grouped["dec"][in_passband],
    )
```

Worth a test alongside it: a two-passband table whose passbands sit in different parts of the sky, asserting the cone is drawn around the fitted passband only.

Minor, while you are in here: `_observed_field` runs before the `cat_name` validity check below, so a call with a bad `cat_name` *and* an unusable position column reports "no finite ra/dec" instead of "Unknown catalog name". Moving the call below the `if`/`elif`/`else` costs nothing.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1066,6 +1190,23 @@ def transform_to_catalog(
...
+    # argument the caller can fix without leaving their keyboard does not cost
+    # them a Vizier round trip first.
+    if search_radius is not None:
+        search_radius = u.Quantity(search_radius)
+        if not search_radius.unit.is_equivalent(u.degree):
+            raise ValueError(
+                "search_radius must be an angle, such as 5 * u.arcmin, got "
+                f"{search_radius!r}."
+            )
+        if not np.isfinite(search_radius.value) or search_radius.value <= 0:
```
mwcraig, line 1217: **A single-element array `search_radius` passes validation unreduced.**

`u.Quantity([5], u.arcmin)` slips through numpy's single-element truthiness shortcut and reaches `apass_dr9(radius=...)` as a shape-`(1,)` Quantity, and is compared against `_WIDE_FIELD_RADIUS` the same way — neither expects a non-scalar.

A multi-element Quantity has the opposite problem: `not np.isfinite(array)` raises numpy's *"truth value of an array with more than one element is ambiguous"*. That is technically a `ValueError`, so the new parametrized test would pass, but the message is not the one this block promises.

Both are fixed by requiring a scalar:

```suggestion
        if not search_radius.isscalar:
            raise ValueError(
                "search_radius must be a single angle, got "
                f"{search_radius!r}."
            )
        if not np.isfinite(search_radius.value) or search_radius.value <= 0:
```

Separately: `u.Quantity(search_radius)` on the line above raises `TypeError`, not `ValueError`, for a value it cannot coerce to numeric at all (a `SkyCoord`, a dict). Narrower than the array case — it only bites unusual misuse, not the common "forgot the unit", which `is_equivalent` already handles — but it does escape the `ValueError` contract the docstring states.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1158,17 +1299,37 @@ def transform_to_catalog(
...
+    # the images in it are of one field, and one cone drawn around all of
+    # them serves the calls for the other passbands too.
+    field_center, field_radius = _observed_field(
+        observed_mags_grouped["ra"], observed_mags_grouped["dec"]
     )
+    if search_radius is None:
+        search_radius = field_radius
+        if search_radius > _WIDE_FIELD_RADIUS:
+            warnings.warn(
+                f"The observed positions span {field_radius.to(u.degree):.2f}, "
```
mwcraig: **The reported span includes the margin.**

`field_radius` comes back from `_observed_field` with `_CATALOG_RADIUS_MARGIN` already added, but the message presents it as what "the observed positions span" — so the number is systematically ~1 arcmin larger than the actual extent of the data. At `:.2f` degrees that is visible (a field spanning 2.00 deg reads as "2.02"). "Span" also reads as a diameter where the value is a radius.

Either subtract the margin back off, or reword to describe the cone rather than the data:

```suggestion
                f"The observed positions reach {(field_radius - _CATALOG_RADIUS_MARGIN).to(u.degree):.2f} "
                "from their center, so the catalog will be fetched from a "
                "cone about that wide and the query may be slow. That is "
                "usually a table holding more than one field; if it is not, "
                "pass search_radius to set the cone size yourself.",
```

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -511,6 +515,32 @@ def _run_transform_to_catalog(
...
+    center : `astropy.coordinates.SkyCoord`
+        The field center the fetch was called with, i.e. its first
+        positional argument.
+
+    radius : `astropy.units.Quantity`
+        The cone search radius the fetch was called with, i.e. its
+        ``radius`` keyword argument.
+    """
+    call_args = getattr(magnitude_transforms, cat_name).call_args
+    return call_args.args[0], call_args.kwargs["radius"]
```
mwcraig: A 21-line numpydoc block for a two-line `getattr`. The file already has precedent for a one-liner on a small helper — `_expected_max_weight_share` at line 2085 is `"""Max share of a least-squares fit's weight, which goes as 1/sigma**2."""`.

```suggestion
def _fetch_call_args(cat_name="apass_dr9"):
    """Return the (center, radius) the patched catalog fetch was called with."""
    call_args = getattr(magnitude_transforms, cat_name).call_args
    return call_args.args[0], call_args.kwargs["radius"]
```

Saves 20 lines.

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1141,6 +1171,77 @@ def test_to_float_array_leaves_unitless_input_alone():
...
+        dec = MaskedColumn(
+            np.concatenate([dec_good.value, [45.5, 45.5]]),
+            mask=[False, False, False, False, True],
+            unit=u.degree,
+        )
+
+    center, radius = _observed_field(ra, dec)
+
+    assert center.separation(center_good) < 1 * u.mas
+    assert u.allclose(radius, radius_good)
```
mwcraig, line 1200: The two branches build the same shape of bad data twice. Building one value pair and branching only on *how* it is spoiled says the same thing in 9 fewer lines, and makes the symmetry between the NaN and masked cases visible rather than something the reader has to check index by index.

```suggestion
@pytest.mark.parametrize("case", ["nan", "masked"])
def test__observed_field_ignores_positions_that_are_not_finite(case):
    # A NaN or masked entry in either column must drop the whole row, not
    # just poison the mean -- checked here by comparing against the field
    # computed from the good rows alone. Two extra rows, one with a bad ra
    # and one with a bad dec, confirm a row is dropped even when it has one
    # finite coordinate.
    ra_good = u.Quantity([180.0, 180.01, 179.99], u.degree)
    dec_good = u.Quantity([45.0, 45.01, 44.99], u.degree)
    center_good, radius_good = _observed_field(ra_good, dec_good)

    ra_values = np.concatenate([ra_good.value, [180.5, 180.5]])
    dec_values = np.concatenate([dec_good.value, [45.5, 45.5]])
    if case == "nan":
        ra_values[3], dec_values[4] = np.nan, np.nan
        ra = u.Quantity(ra_values, u.degree)
        dec = u.Quantity(dec_values, u.degree)
    else:
        ra = MaskedColumn(ra_values, mask=[False] * 3 + [True, False], unit=u.degree)
        dec = MaskedColumn(dec_values, mask=[False] * 4 + [True], unit=u.degree)

    center, radius = _observed_field(ra, dec)

    assert center.separation(center_good) < 1 * u.mas
    assert u.allclose(radius, radius_good)
```

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2798,6 +2899,131 @@ def test_transform_to_catalog_unknown_catalog_raises():
...
+    expected_center = SkyCoord(
+        np.mean(observed["ra"]), np.mean(observed["dec"]), unit="degree"
+    )
+    assert center.separation(expected_center) < 1 * u.arcsec
+
+    separations = SkyCoord(observed["ra"], observed["dec"], unit="degree").separation(
+        center
+    )
+    assert u.allclose(radius, separations.max() + _CATALOG_RADIUS_MARGIN)
+    assert (separations <= radius).all()
```
mwcraig: This one cannot fail independently of the line above it: `_CATALOG_RADIUS_MARGIN` is strictly positive, so once `radius == separations.max() + margin` holds, every separation is below `radius` by construction.

(The `assert radius < 1 * u.degree` below it is a different matter — that one is the actual point of #686 and should stay.)

```suggestion
```

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2798,6 +2899,131 @@ def test_transform_to_catalog_unknown_catalog_raises():
...
+    observed : `astropy.table.Table`
+        Observations of those positions, grouped by file.
+    """
+    # 20 stars on a grid 40 arcmin apart span about two degrees, so the cone
+    # drawn around them comes out well over the one degree that used to be
+    # hardcoded -- which is exactly the case the warning is about.
+    coordinates = _generate_star_coordinates(20, separation=40 * u.arcmin)
+    catalog, ra, dec, instrumental = _generate_fake_catalog(20, coordinates=coordinates)
+
+    return catalog, _generate_observed_table(ra, dec, instrumental)
```
mwcraig: The `Returns` block restates what `_generate_fake_catalog` and `_generate_observed_table` already document. The rationale comment below it is doing the real work and should stay.

```suggestion
def _wide_field_observations():
    """Observations spread over a field wider than `_WIDE_FIELD_RADIUS`."""
    # 20 stars on a grid 40 arcmin apart span about two degrees, so the cone
    # drawn around them comes out well over the one degree that used to be
    # hardcoded -- which is exactly the case the warning is about.
    coordinates = _generate_star_coordinates(20, separation=40 * u.arcmin)
    catalog, ra, dec, instrumental = _generate_fake_catalog(20, coordinates=coordinates)

    return catalog, _generate_observed_table(ra, dec, instrumental)
```

Saves 12 lines.

mwcraig (commented):
Critical review focused on correctness regressions and on test size. One finding I'd call blocking; the rest are small.

**Bugs**

1. **The cone is sized from rows that will never be calibrated** (`magnitude_transforms.py:1302`). `_observed_field` is handed the whole table, but only `in_passband` rows are ever fit or written. For a table holding several targets across several bands, the cone is drawn around all of them while only one target's rows get calibrated — a multi-degree query where an arcminute one would do, plus a spurious wide-field warning. This is #686's own defect arriving through a different door, and it's the one I'd want fixed before merge.
2. **A single-element array `search_radius` passes validation unreduced** (`:1203`) and reaches the Vizier call as a shape-`(1,)` Quantity. A multi-element one raises numpy's "ambiguous truth value" rather than the intended message.
3. **The wide-field warning overstates the field's span** (`:1312`) by the 1 arcmin margin, which is visible at the `:.2f` degree precision used.
4. Minor: `_observed_field` runs before the `cat_name` check, so a bad `cat_name` on a table with unusable positions reports the wrong error.

**Checked and ruled out** — `_to_degree_array`'s mask handling (verified equivalent to `_to_float_array` across `MaskedColumn`, `Column`, `Quantity`, `Masked(Quantity)`, lists and int dtypes); `stacklevel=2`; the 1 arcmin margin against `_CAL_MATCH_ARCSEC`; any minimum or maximum radius in `from_vizier`; and regression risk from the new `ValueError` — any table that worked before had a finite row 0, which is still `usable`. The near-antipodal centroid degeneracy is real but harmless, since the radius is computed *from* the chosen center and so encloses every point regardless.

**Test size**

About 57 of the 239 added test lines come off with no coverage lost, as inline suggestions below: trimming two helper docstrings (32), collapsing the three warning tests into one parametrized test on `recwarn` (15, including the now-unneeded `import warnings`), compacting the NaN/masked branches (9), and dropping one assertion that its predecessor already implies (1).

I would **not** drop any of the four `_observed_field` unit tests. The integration test uses a plain 10-arcsec grid with no wrap, no masked rows, and more than one star, so it exercises none of their edge cases. Merging the default-radius and explicit-radius tests would save ~13 more lines but needs an early-return branch in the body — not worth it. And no comment in the new tests merely restates its assertion, so none should go.

---
*This review was written by Claude at Matt's direction.*

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -2798,6 +2899,131 @@ def test_transform_to_catalog_unknown_catalog_raises():
...
+    # only a cone that came out wide on its own is.
+    catalog, observed = _wide_field_observations()
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        _run_transform_to_catalog(mocker, catalog, observed, search_radius=3 * u.degree)
+
+    assert not [
+        warning for warning in caught if "search_radius" in str(warning.message)
+    ]
```
mwcraig: These three are one property — *the warning fires for a derived wide cone and nothing else* — split across three tests, two of which repeat the same 5-line list-comprehension assertion.

Collapsing them onto `recwarn` saves 14 lines here and one more at the top of the file: the `import warnings` added in this PR exists only for these two `catch_warnings` blocks. `warnings.` appears nowhere else in the file, and the established idiom here is `pytest.warns` (5 other uses).

```suggestion
@pytest.mark.parametrize(
    "wide_field, search_radius, expect_warn",
    [(False, None, False), (True, None, True), (True, 3 * u.degree, False)],
    ids=["ordinary_field", "derived_wide_cone", "explicit_wide_radius"],
)
def test_transform_to_catalog_warns_only_for_a_derived_wide_cone(
    mocker, recwarn, wide_field, search_radius, expect_warn
):
    # A field this wide is almost always a mosaic, or ra/dec columns that
    # hold more than one field's worth of stars. Before the cone was derived
    # such a table quietly got a one degree catalog and calibrated against
    # whatever fell inside it; now it gets a cone that really covers the
    # table, which is correct but is also a query worth hearing about --
    # unless the caller asked for that width on purpose.
    if wide_field:
        catalog, observed = _wide_field_observations()
    else:
        catalog, ra, dec, instrumental = _generate_fake_catalog(20)
        observed = _generate_observed_table(ra, dec, instrumental)

    kwargs = {} if search_radius is None else {"search_radius": search_radius}
    _run_transform_to_catalog(mocker, catalog, observed, **kwargs)

    caught = [w for w in recwarn.list if "search_radius" in str(w.message)]
    assert bool(caught) is expect_warn
    if expect_warn:
        _, radius = _fetch_call_args()
        assert radius > _WIDE_FIELD_RADIUS
```

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -511,6 +515,32 @@ def _run_transform_to_catalog(
     return transform_to_catalog(observed, obs_filter, cat_name=cat_name, **call_kwargs)
 
 
+def _fetch_call_args(cat_name="apass_dr9"):
```
mwcraig: Can't pytest.mock handle this more easily?

stellarphot/utils/tests/test_magnitude_transforms.py
```diff
@@ -1141,6 +1171,77 @@ def test_to_float_array_leaves_unitless_input_alone():
...
+            unit=u.degree,
+        )
+
+    center, radius = _observed_field(ra, dec)
+
+    assert center.separation(center_good) < 1 * u.mas
+    assert u.allclose(radius, radius_good)
+
+
+def test__observed_field_raises_when_no_position_is_finite():
```
mwcraig: Why a double-underscore in this test name and some of the others?

stellarphot/utils/magnitude_transforms.py
```diff
@@ -279,6 +297,93 @@ def _to_float_array(values):
     return (filled * unit).to_value(u.mag)
 
 
+def _to_degree_array(values):
```
mwcraig, line 306: Why on earth is this needed when one can do coord.ra.degree?

stellarphot/utils/magnitude_transforms.py
```diff
@@ -279,6 +297,93 @@ def _to_float_array(values):
...
+    of the stars it was meant to hold.
+
+    A row is dropped unless both of its coordinates are usable, since half a
+    position places nothing, and a masked or NaN entry left in would poison
+    the mean rather than merely be ignored by it.
+    """
+    ra_degree = _to_degree_array(ra)
+    dec_degree = _to_degree_array(dec)
+    usable = np.isfinite(ra_degree) & np.isfinite(dec_degree)
+    if not usable.any():
```
mwcraig, line 378: I don't think this can ever happen on real data, since any potential photometry sites whose positions turn out to be garbage are dropped during the photometry process. So while an edge case in principle, it is not in real life.

stellarphot/utils/magnitude_transforms.py
```diff
@@ -1039,6 +1151,18 @@ def transform_to_catalog(
     ``mag_cal_error`` is NaN, rather than falling back to the measurement
     error alone, for an image whose fit left no usable covariance behind.
 
+    The catalog is fetched once per call, from a cone centered on the
```
mwcraig, line 1160: That may be too broad a scope, since hte field of view might vary between passbands. It should not vary a lot, but iut might vary a bit/

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed, and agreed — except for the minor point at the end, which does not work as written.

Verified: only `in_passband` rows are ever matched (`one_image = one_image_all_bands[in_this_group]` feeding `match_to_catalog_sky`, lines 1441–1444) and only they are written (line 1728). And nothing is shared between calls — `apass_dr9`/`refcat2` are invoked fresh each time, so "one cone ... serves the calls for the other passbands too" is not true of any code path. In `transform-to-appas-dr9.ipynb` the table is grouped by passband and each group is passed in separately, so there the cone already comes from one band's rows and this change is a no-op; the mixed-band table is where it bites, and that is exactly the case `in_passband` exists for.

@mwcraig makes the same point from the other direction on the Notes paragraph at line 1154 — the field of view varies a little between bands, so all-rows is the wrong scope even for a single field, never mind several.

Will make the change and add the two-passband test.

**The minor ordering point is wrong as stated**, though: `_observed_field` cannot move below the `if`/`elif`/`else`, because `field_center` is what those branches hand to `apass_dr9`/`refcat2` (lines 1323 and 1331) — moving the call down is a `NameError`. To get "Unknown catalog name" to win, `cat_name` has to be checked *early* instead, next to the `search_radius` validation at line 1196:

```python
if cat_name not in ("apass_dr9", "refcat2"):
    raise ValueError(
        f"Unknown catalog name {cat_name}. Must be one of 'apass_dr9' or 'refcat2'."
    )
```

which belongs there on its own merits — it is an argument the caller can fix without leaving their keyboard, the same reason `search_radius` is checked up there. The `else` branch below then stays as a guard but stops being the thing that reports the error.

stellarphot/utils/magnitude_transforms.py
mwcraig: Confirmed, both halves, in a live session:

```
>>> q = u.Quantity([5], u.arcmin)
>>> not np.isfinite(q.value), bool(q.value <= 0), q.isscalar
(False, False, False)
>>> q > 1 * u.degree
<Quantity [False]>
>>> not np.isfinite(u.Quantity([5, 6], u.arcmin).value)
ValueError: The truth value of an array with more than one element is ambiguous.
```

So a shape-`(1,)` Quantity passes validation, passes the `_WIDE_FIELD_RADIUS` comparison as a one-element array, and reaches `apass_dr9(radius=...)` — and from there `CatalogData.from_vizier` (`catalogs.py:223`) — unreduced. Agreed on the `isscalar` check.

Also worth noting the current parametrized list is `[no_unit, wrong_unit, negative, zero, nan]` — neither array case is in it, so nothing today catches this. I will add `u.Quantity([5], u.arcmin)` and `u.Quantity([5, 6], u.arcmin)` alongside the `isscalar` check.

**One correction on the `TypeError` sub-point**: the docstring has no `Raises` section, so there is no stated `ValueError` contract to escape — line 1058 only says the value "must be a positive, finite angle". And what astropy raises for a `SkyCoord` or a dict in that slot is `TypeError: The value must be a valid Python or Numpy numeric type.`, which names the problem accurately. I would leave that one alone unless you want it wrapped for uniformity.

stellarphot/utils/magnitude_transforms.py
mwcraig: Correct — `_observed_field` returns `coords.separation(center).max() + _CATALOG_RADIUS_MARGIN` (line 384), so the number in the message is the cone, not the data, and at `:.2f` degrees the 1 arcmin margin shows up as +0.02.

Taking the second option, rewording rather than subtracting the margin back off: the cone is what actually gets queried and what the caller would pass `search_radius` to change, so reporting a number that is *not* the one used seems worse than reporting the one that is. "Span" also goes — the value is a radius.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed, and @mwcraig asks the same thing about this helper on the line above — see that thread. Going further than the suggestion here: the helper comes out entirely. All three call sites (lines 2911, 2939, 2979) use the default `cat_name`, so the parameter it exists to serve is never exercised, and the file already reaches the mock inline the obvious way at line 3024 (`magnitude_transforms.apass_dr9.call_count == 0`).

So those three become e.g.

```python
_, radius = ..., magnitude_transforms.apass_dr9.call_args.kwargs["radius"]
```

21 lines of docstring plus 4 of helper, gone.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — same coverage, and the NaN/masked symmetry becomes something you can see rather than something you check index by index. Taking the suggestion.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Right — `_CATALOG_RADIUS_MARGIN` is `1 * u.arcmin`, strictly positive, so `radius == separations.max() + margin` on the line above already forces every separation below `radius`. It cannot fail on its own. Removing it; `assert radius < 1 * u.degree` stays.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed — the `Returns` block says nothing the two calls beneath it do not. Taking the suggestion; the rationale comment stays.

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: Agreed, with one amendment.

Checked the two supporting claims: `import warnings` is on line 2 and `warnings.` appears only at lines 2987 and 3001, both added by this PR, so the import goes with these tests; and `pytest.warns` is used at lines 860, 2976, 3522, 3562, 4072 and 4103, so the idiom is well established here.

Also verified `recwarn` behaves under this repo's `filterwarnings = ["error", ...]` (pyproject.toml:236) — wrote the parametrized shape as a scratch test and ran it under `-c pyproject.toml`, both cases pass. That was the one thing I would not have assumed.

**Amendment**: the suggestion drops the warning *category*, which `pytest.warns(AstropyUserWarning, ...)` asserts today. Keeping it costs nothing:

```python
caught = [
    w
    for w in recwarn.list
    if issubclass(w.category, AstropyUserWarning) and "search_radius" in str(w.message)
]
```

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: It can, and more easily than this — the helper should just go.

After `mocker.patch.object(magnitude_transforms, cat_name, return_value=catalog)`, the mock *is* `magnitude_transforms.apass_dr9`, and this file already reaches it that way inline at line 3024: `assert magnitude_transforms.apass_dr9.call_count == 0`. The only thing `_fetch_call_args` adds is the `getattr(..., cat_name)` lookup for a non-default catalog — and all three call sites (2911, 2939, 2979) use the default, so that parameter is never exercised.

So those become

```python
radius = magnitude_transforms.apass_dr9.call_args.kwargs["radius"]
center = magnitude_transforms.apass_dr9.call_args.args[0]
```

and the helper plus its 21-line docstring comes out. (The review above asked for the docstring to be trimmed to a one-liner; deleting the whole thing is strictly better and I will do that instead.)

stellarphot/utils/tests/test_magnitude_transforms.py
mwcraig: No good reason — it is just `test_` prefixed onto the private name `_observed_field`, and I should have collapsed it. It is inconsistent with this PR's own tests for the equally private `_to_float_array`, which are `test_to_float_array_*` (lines 1114–1171), and these four are the only `def test__` names in the whole repo:

```
$ grep -rn "^def test__" --include=*.py stellarphot | wc -l
4
```

Renaming all four to `test_observed_field_*`.

stellarphot/utils/magnitude_transforms.py
mwcraig: Because there is no `SkyCoord` yet at that point, and building one first loses information: **`SkyCoord` silently drops a column's mask.**

```
>>> ra = MaskedColumn([180., 181., 182.], mask=[False, True, False], unit=u.degree)
>>> dec = MaskedColumn([45., 46., 47.], mask=[False, True, False], unit=u.degree)
>>> SkyCoord(ra, dec).ra.degree
array([180., 181., 182.])
```

The masked entry comes back as 181.0, not NaN — so `coord.ra.degree` would fold a position that is explicitly *absent* into the centroid as if it had been measured. The whole point of `_to_degree_array` is to turn masked into NaN **before** the `SkyCoord` exists, so `usable` can drop those rows; once the coordinate object is built, the mask is gone and there is nothing left to filter on.

The NaN case alone would indeed be fine via `coord.ra.degree`. It is the masked case that needs this, and a `ra`/`dec` column read back from ECSV or FITS with missing values is a `MaskedColumn`. That is the same reason `_to_float_array` exists next door for the magnitude columns — this is its three-line sibling for angles, minus the unit check, which is the one thing that differs.

stellarphot/utils/magnitude_transforms.py
mwcraig: It is reachable, though — not from garbage positions, from *absent* ones. `single_image_photometry` fills `ra`/`dec` with NaN for every source when the image has no WCS:

```python
# photometry/photometry.py:563-569
try:
    skypos = ccd_image.wcs.pixel_to_world(xs, ys)
    ...
except AttributeError:
    ra = [np.nan] * len(xs)
    dec = [np.nan] * len(ys)
```

and `source_detection` does the same (`source_detection.py:335-338`). So a table built from WCS-less images has *every* position NaN — nothing was dropped, because nothing was ever measured. Handing that to `transform_to_catalog` is a user error, but it is one the code can produce.

What it costs to leave the check out is the reason I would keep it. Without the guard the failure is:

```
RuntimeWarning: Mean of empty slice.          (x3)
ValueError: zero-size array to reduction operation maximum which has no identity
```

which says nothing about positions, catalogs or what the caller did. Four lines buys "No observed position has both a finite ra and a finite dec". On `main` the same table produced a Vizier query centered on `(nan, nan)`, which was worse still.

The `usable` mask is load-bearing regardless, for the ordinary partial case; the raise is only its boundary condition, so it is not really a separate mechanism being carried for an edge case. Happy to drop the dedicated test if the four lines are what bother you, but I would keep the raise.

stellarphot/utils/magnitude_transforms.py
mwcraig: Agreed, and it is the same defect the first review finding names from the other side — I have replied there in more detail.

Verified that only the `in_passband` rows are ever matched against the catalog (lines 1441–1444) or written (line 1728), and that nothing is cached between calls, so sizing the cone from the whole table buys nothing even when the field really is identical across bands. Your point makes it worse than merely useless: a slightly different pointing or rotation per band means the all-rows cone is systematically larger than any single band needs.

Fixing it to `observed_mags_grouped["ra"][in_passband]`, with the Notes paragraph reworded to match, plus a test with two passbands in different parts of the sky.

